Issue #80, opened 2024-03-01
Docs: Adding files to S3


dvenprasad:
write docs with instructions to add files to S3

PR #330, opened 2024-04-01
Create a script to sync results

### LINKED ISSUE #270: Create a script to sync results
When contributors have results to share, they will need to sync them to their S3 bucket, and we would like to have that syncing maintain the module directory name etc. 

We should include a small python script that creates the proper aws cli command, which will look something like this:

```
aws s3 sync analyses/{module}/results s3://{bucket}/module/results --delete
```

`module` and `bucket` would be set by command line options in the python script

Note the `--delete` option will delete files on S3 that are not in the results directory, which I think is what we want to maintain only current results, we could perhaps disable that with a command line flag on the script. We also may want to set `--exact-timestamps` within the `aws s3 sync` to make the sync a bit more stringent. 

We will also want to include a `--profile` option,  though we might recommend that people set `AWS_PROFILE` environment variable. 

Similarly, we may want to suggest setting an environment variable like `OPENSCPCA_RESULTS_BUCKET` and reading from that rather than using the `--bucket` option on every call.

jashapiro:

This Draft PR adds  a script to sync results and plots folders to a user's analysis bucket.

Closes #270

The idea is to make it as easy as possible for users to sync their results to an S3 bucket for sharing!

In this usual case, this should just mean invoking this script with something like `.sync-results.py --bucket {my_bucket} --module {my_module}` perhaps with an added `--profile` option.

The simple case just syncs the `results` and `plots` subdirectories from the analysis directory into their bucket. In the usual case, the directories are synced exactly, except for skipping hidden files. 

Should be able to get it all done here!

Currently, files that are present in the S3 bucket but _not_ in the results folder will be deleted, unless the `--no-delete` flag is included. I went back and forth about which should be the default, so please let me know if you think I should change the option to `--delete` and keep all files be default!

Similarly, if you think I should add an option to include hidden files, that could be done pretty simply.

Finally, I wasn't sure what location to use for this script. I stashed it in a `scripts` directory for now, but we could move it out to the root, if that seems worthwhile.

One other thought I had was whether to look in the environment for a variable like `OPENSCPCA_BUCKET` that could be set once and save some typing. Thoughts on that idea? (Could be added later too)


sjspielman:
> Finally, I wasn't sure what location to use for this script. I stashed it in a scripts directory for now, but we could move it out to the root, if that seems worthwhile.

Just a quick note that we'll probably want to link to the download script here, or at least tell contributors where in the repo it is. https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/getting-started/accessing-resources/getting-access-to-data.md#using-the-download-data-script

I do think it would be reasonable sense to keep it at the root level along with `create-analysis-module.py`.

jaclyn-taroni (approved):
👍🏻 

I lightly tested this (I didn't try every single option)

I will include instructions for this in #80

PR #340, opened 2024-04-03
Use date with datetime

sjspielman:
This PR fixes a bug I encountered running the data download script - it's not `datetime.today()`, but `datetime.date.today()`.


jashapiro (approved):
Thanks!

PR #341, opened 2024-04-03
Update mkdocs.yml nav to include security policy

jaclyn-taroni:
I noticed the security policy was not listed on the left hand side because I forgot to add this in #320


Jen-OMalley (approved):
🚀

PR #345, opened 2024-04-03
Docs for syncing to S3

sjspielman:
### Which issue does this address?

Closes #82 
Stacked on #342 

### Briefly describe the scope of the added docs file, including whether you link out to any external docs.

This PR adds a section to the S3 docs for syncing your results using the sync script. While I was adding some relative links in from AWS CLI setup, I noticed a few small items there to tweak as well.

Let me know if more details would be helpful here! For example, I should highlight any of the other script flags.

### Will you need additional visual aids for these docs?

No

### Any other comments for reviewers?

No

### Author checklist

- [x] The docs page has been added to the `nav` section in `mkdocs.yml`
- [x] The docs adhere to the style guide [**structure directions**](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/general-style-guide.md#structure)
- [x] The docs adhere to the style guide [**language directions**](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/general-style-guide.md#structure)
- [x] The docs adhere to the style guide [**tone directions**](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/general-style-guide.md#tone)
- [x] The docs follow additional contributing instructions in [the docs contributing file](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/contributing.md)

### Reviewer checklist

Please refer to the [docs style guide](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/general-style-guide.md) and the [docs contributing guide](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/contributing.md) while reviewing this pull request.

- [ ] Do the docs adhere to the style guide [**structure directions**](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/general-style-guide.md#structure)?
- [ ] Do the docs adhere to the style guide [**language directions**](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/general-style-guide.md#structure)?
- [ ] Do the docs adhere to the style guide [**tone directions**](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/general-style-guide.md#tone)?
- [ ] Do the docs follow additional contributing instructions in [the docs contributing file](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/contributing.md)?

PR #347, opened 2024-04-03
Change default delete behavior

jashapiro:
Here I changed the sync script to reverse the default behavior. Now all files are kept unless you use `--destructive-sync`. Scary!

Stacked on #345

### DIFF
--- a/scripts/sync-results.py
+++ b/scripts/sync-results.py
@@ -53,15 +53,18 @@ def main() -> None:
         help="Skip syncing the results directory.",
     )
     parser.add_argument(
-        "--no-delete",
-        action="store_false",
+        "--with-delete",
+        action="store_true",
         dest="delete_extra",
-        help="Do not delete extra files in the S3 bucket.",
+        help=(
+            "Delete any files in the S3 bucket that are not found in the local directories being synced."
+            " Default behavior is to leave any additional files in the S3 bucket."
+        ),
     )
     parser.add_argument(
         "--dryrun",
         action="store_true",
-        help="Perform a dry run: show what would be done but do not sync anything.",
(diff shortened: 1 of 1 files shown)


scripts/sync-results.py
```diff
@@ -53,15 +53,18 @@ def main() -> None:
...
     )
     parser.add_argument(
-        "--no-delete",
-        action="store_false",
+        "--destructive-sync",
+        action="store_true",
         dest="delete_extra",
-        help="Do not delete extra files in the S3 bucket.",
+        help=(
+            "Delete any files in the S3 bucket that are not found in the local directories being synced."
```
sjspielman, line 60:
```suggestion
            "Delete any files present in the S3 bucket that are not present in the local directories being synced."
```

sjspielman (approved):
👍 , take or leave the suggestion.

I wonder if I should emphasize `--dryrun` in the docs, too, as a good place to start to get your bearings? Thoughts?

scripts/sync-results.py
```diff
@@ -136,6 +136,7 @@ def main() -> None:
         sys.exit(1)
 
     print("Sync complete.")
+    print("Files are now available on S3 at: s3://{bucket}/{args.module}/")
```
sjspielman: 👍

jashapiro:
> I wonder if I should emphasize `--dryrun` in the docs, too, as a good place to start to get your bearings? Thoughts?

Yes, I think I said that in my review?

scripts/sync-results.py
jashapiro: in between: 
```suggestion
            "Delete any files in the S3 bucket that are not present in the local directories being synced."
```

PR #350, opened 2024-04-04
Bug fixes and comment updates for sim-sce scritps

jashapiro:
While working on https://github.com/AlexsLemonade/OpenScPCA-nf/issues/9, I found a few little bugs/edge cases in the simulation scripts. This PR fixes them:

- Always read the metadata as characters to prevent reading participant ids that look like numbers as floats (converting floats to factors with `forcats::fct_anon()` fails)
- Make sure there are at least `ncells` cells to select from before selecting. This one probably could be "fixed" by allowing resampling, but having a simulated dataset with fewer than 100 cells seems fine.


allyhawkins (approved):
LGTM

PR #352, opened 2024-04-04
Add code styling workflow

### LINKED ISSUE #85: Add code styling workflow
This issue tracks implementing a code styling workflow which will run periodically over the repo. 

Technical docs: https://github.com/AlexsLemonade/OpenScPCA-admin/blob/main/technical-docs/repository-structure-and-specifications.md#code-styling

sjspielman:
Closes #85 

This PR adds a code styling workflow that will style all `py` and `R`/`Rmd` files in `analyses`. 

I'm opening this as a Draft since I still need to figure out the right tokens to be using in this repo (CC @jashapiro, am I close here?), and also this needs to be tested with temporary PR trigger which has not yet been added.


.github/workflows/code-styling.yml
```diff
@@ -0,0 +1,75 @@
...
+        uses: actions/checkout@v4
+
+      - name: Load 1Password secrets
+        uses: 1password/load-secrets-action@v1
+        with:
+          export-env: true
+        env:
+          # TODO: Which tokens do we actually need here?
+          OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_DOCS_BOT_SERVICE_TOKEN }}
+          DOCS_BOT_GITHUB_TOKEN: ${{ secrets.OP_DOCS_BOT_GITHUB_TOKEN }}
```
jashapiro, line 30: Right now there are no secrets set in this repository, so this won't work. I don't know if we plan a separate set of tokens for this repo, but these are the right idea.

But action v2 is out, so we should probably use that.

.github/workflows/code-styling.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      ## First, style the R code ##
+      - name: Install R
+        uses: r-lib/actions/setup-r@v2
+
+      - name: Install R dependencies
+        run: install.packages("styler", repos = "https://packagemanager.posit.co/cran/__linux__/jammy/latest")
```
jashapiro, line 37: New url who dis
```suggestion
        run: install.packages("styler", repos = "https://p3m.dev/cran/__linux__/jammy/latest")
```

.github/workflows/code-styling.yml
```diff
@@ -0,0 +1,75 @@
...
+        shell: Rscript --no-init-file {0}
+
+      - name: Style analysis modules R code
+        run: styler::style_dir("analyses", filetype = c("R", "r", "Rmd", "rmd")) # exclude_dirs and exclude_files do not seem to like globs...
```
jashapiro, line 41: You could handle this by using purrr (which is already required by styler):
```suggestion
        run: |
          list_dirs("analysis") |>
            purrr::walk(styler::style_dir, filetype = c("R", "r", "Rmd", "rmd"))
```

Since styler already skips the `renv` directory by default, this should handle leaving `activate.R` alond

sjspielman: 😸

.github/workflows/code-styling.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Install python dependencies
+        run: pip install ruff
+
+      - name: Style analysis modules python code
+        run: ruff format analyses/
+
+      - name: Create PR with styled code
+        uses: peter-evans/create-pull-request@v6
+        with:
+          token: ${{ env.DOCS_BOT_GITHUB_TOKEN }}
```
jaclyn-taroni, line 61: Maybe https://docs.github.com/en/actions/security-guides/automatic-token-authentication ?

.github/workflows/code-styling.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      ## Second, style the python code ##
+      - name: Install python
+        uses: actions/setup-python@v2
+
+      - name: Install python dependencies
+        run: pip install ruff
+
+      - name: Style analysis modules python code
+        run: ruff format analyses/
```
jashapiro, line 56: The [euff github page](https://github.com/astral-sh/ruff?tab=readme-ov-file) suggests just using this action:
```suggestion
      - name: Install python
        uses: chartboost/ruff-action@v1
        with:
          args: format analyses/
```

.github/workflows/code-styling.yml
jaclyn-taroni: I checked our repo settings – this should work for this purpose based on my read.

jashapiro: I think the issue with that is that a PR filed by this action would not then trigger required checks.

.github/workflows/code-styling.yml
sjspielman: 👍 I had come across this but did not realize it was _recommended_

sjspielman:
https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/354

.github/workflows/code-styling.yml
```diff
@@ -0,0 +1,76 @@
+name: Code styling
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 1 * *" # first of every month at midnight
```
jashapiro, line 5: I think I would not schedule this yet, and if we do, we probably want it to _not_ be the same as spelling, as then we will always have competing and potentially conflicting PRs.

.github/workflows/code-styling.yml
```diff
@@ -0,0 +1,76 @@
...
+
+
+      ## Second, style the python code ##
+      - name: Format python files
+        uses: chartboost/ruff-action@v1
+        with:
+          args: format analyses/
+
+      - name: Create PR with styled code
+        uses: peter-evans/create-pull-request@v6 # defaults to ${GITHUB_TOKEN} token
```
jashapiro, line 47: Unfortunately, I do think we may have to use secrets for this, or this PR will not trigger the code checks that we require. https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs

Of course, we could just leave it and take the first option of manually closing and reopening each PR.

If we want to be more cautious and still use automatically trigger the test actions, we could potentially create repo-specific tokens, which are a thing now (in beta).

.github/workflows/code-styling.yml
```diff
@@ -0,0 +1,63 @@
...
+          branch: style-analysis-modules
+          base: main
+          delete-branch: true
+          title: "GHA: Automated analysis module code styling"
+          body: |
+            ### Description:
+            This PR was auto-generated by GitHub Actions to style all R (scripts and notebooks) and Python (scripts only) code in `analyses`.
+
+            ### Instruction for reviewers:
+              Make sure that all code changes are acceptable.
```
jashapiro, line 60: Alignment, unless this was intentional.
```suggestion
            ### Instruction for reviewers:
            Make sure that all code changes are acceptable.
```

jashapiro (commented):
Overall this looks good, but I think we probably don't want this to be run too often, as it is likely to introduce all kinds of conflicts in actively developed modules, which makes me think maybe not on a schedule at all.

sjspielman:
> which makes me think maybe not on a schedule at all.

What about have a cron job file an issue every 3 (4?) months to prompt us to run it?

.github/workflows/code-styling.yml
sjspielman: It's definitely fine with me to leave this PR open until we have secrets ready to go

sjspielman:
#360 ✔️

.github/workflows/code-styling.yml
```diff
@@ -0,0 +1,72 @@
+name: Code styling
+on:
+  #pull_request:
+  #  branches:
+  #    - main
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 15 1/3 *" # 15th of every 3rd month, for now
```
jashapiro, line 8: Running this on a schedule seems like overkill for the moment, let's comment this out, then we can at least try it as a manual trigger (for #623)
```suggestion
 #  schedule:
 #    - cron: "0 0 15 1/3 *" # 15th of every 3rd month, for now
```

jashapiro (approved):
Let's merge this in for now to help with #623 and see how things go.

PR #355, opened 2024-04-04
Add a docker image to simulate-sce

jashapiro:
This PR adds a docker image for the `simulate-sce` package. 

I made some minor updates to the `renv.lock` file, then used that to create the docker image, using the Bioconductor:3.18 base image to simplify installation, as all prereqs are already installed. It still isn't the lightest docker image, but I think that is okay. It _is_ lighter than the current `scpcaTools` image.


allyhawkins (approved):
LGTM. I also confirmed I was able to pull the image.

PR #357, opened 2024-04-05
GHA to check links

### LINKED ISSUE #353: Consider github action to detect bad links
Passing thought - The output of `mkdocs serve` provides both `INFO` and `WARNING` messages regarding different kinds of broken links. We might want to capture this information on PRs to main so we can know about bad links before deployment goes through.

sjspielman:
Closes #353 

Opening a draft PR here with a first stab at writing an action to check links. I had hoped to use `linkchecker-markdown` (https://github.com/scivision/linkchecker-markdown), but when testing it locally it did not seem to work as advertised. So I hacked my own way! Let's see if it's useful.


sjspielman:
Failed as expected: https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/8570765973/job/23489428816

Maybe we can also have this run only on files that were modified in the PR.. eh? https://github.com/marketplace/actions/paths-changes-filter

sjspielman:
Noting this is now passing only because `docs` files are no longer modified, but see the failed action linked above for the one bad link that remains.

.github/workflows/check-links.yml
```diff
@@ -0,0 +1,49 @@
...
+        run: |
+          # Build the site and save log output
+          mkdocs build &> build-log.txt
+
+          # Capture all the bad links
+          grep "relative link" build-log.txt > bad_links.txt
+
+          # Save the number of bad links to github output
+          n_bad_links=`wc -l bad_links.txt | cut -f 1 -d " "`
+          echo "n_bad_links=$n_bad_links" >> $GITHUB_OUTPUT
```
jashapiro, line 35: You can do this in 1: 
```
          # Save the number of bad links to github output
          echo "n_bad_links=$(grep "relative link" build-log.txt | wc -l)" >> $GITHUB_OUTPUT
```

The trick here is that `wc -l` doesn't print a file name if it is working on stdin. Also `$()` for command substitution.

But I see you want the file, so the real suggestion is:
```suggestion
          grep "relative link" build-log.txt > bad_links.txt

          # Save the number of bad links to github output
          echo "n_bad_links=$(wc -l < bad_links.txt)" >> $GITHUB_OUTPUT
```

.github/workflows/check-links.yml
```diff
@@ -0,0 +1,49 @@
...
+  check-docs-links:
+    runs-on: ubuntu-latest
+    name: Check links
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install python
+        uses: actions/setup-python@v2
```
jashapiro, line 18: This probably isn't needed. Python is already installed: this action is really only if you need a specific version (and it is at v5)

sjspielman: Oh interesting. I wonder if I can remove this install for the code styling workflow too

sjspielman:
No idea why this is now failing. The code in the `Find all the broken links` step works perfectly fine for me locally in bash and zsh.

.github/workflows/check-links.yml
sjspielman: Ah nevermind, it's been removed already since we replaced it with the recommended ruff action.

jashapiro (approved):
LGTM

Issue #301, opened 2024-03-27
Check links before launch


sjspielman:
This issue tracks a quick check of all links before launch.

sjspielman:
These PRs have addressed a lot of the links: #315, #316, #322, #332, and #335 as of the time of this comment (2024-04-02 1:30 EDT).

sjspielman:
Links all look good to me 🚀

PR #363, opened 2024-04-05
Make headers lowercase after first word

sjspielman:
I did a whole sweep of the website and all links seemed good! (#301 ✅ )

But I did find a couple headers that don't follow the overall style of only capitalizing the first word, so I fixed those here.


allyhawkins (approved):
🚀

PR #365, opened 2024-04-05
Fix a few more simulation bugs

jashapiro:
Just a couple more little ports from https://github.com/AlexsLemonade/OpenScPCA-nf/pull/10

Fixes a type mismatch bug and only calculates UMAPs for libraries that had UMAPS from scpca-nf.


allyhawkins (approved):
👍

Issue #373, opened 2024-04-18
Initiate module for cell typing Ewing sarcoma cells and compile a list of marker genes used to identify tumor cells


allyhawkins:
### Please link to the GitHub Discussion for this proposed analysis.

https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/292

### Describe the goals of this analysis module.

This analysis module will be used to add cell type annotations to all Ewing sarcoma samples. Specifically, the code in the analysis module will be able to provide: 
- Annotations of tumor vs normal cells 
- Cell types for all normal cells
- Annotations of tumor cell states (e.g., Mesenchymal stem cell like) 

Additionally, this module will store a marker gene list used to identify tumor cells and tumor cell states. 

### What software will you require?

This will be done in a mix of R and Python. Most analysis will be done in R using `SingleCellExperiment` objects, but I may use `CellAssign`, which requires Python. If using `CellAssign`, I will try approaches that keep overall computational cost and run time low (e.g., use small lists of marker genes). 

I also plan on using [`InferCNV`](https://github.com/broadinstitute/infercnv) as an approach to identifying tumor or normal cells. 
We may also consider other approaches for cell typing like [`scATOMIC`](https://github.com/abelson-lab/scATOMIC) and [`scGPT`](https://github.com/bowang-lab/scGPT). 

### What will your first pull request contain?

My first pull request will contain just the creation of the module by running the [module creation script](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/create-analysis-module.py). 

The next PR will contain a set of marker genes used to distinguish Ewing sarcoma tumor cells. These will be curated by going through the literature and looking for commonly used genes to distinguish tumor cells, like NKX2.2 and CD99/MIC2. 

### What computational resources will you require?

For this first issue/ PR, I will only be searching through the literature and creating a tsv file with marker genes. So no computational resources are needed.

PR #375, opened 2024-04-19
Initiate analysis module for cell typing Ewing sarcoma samples

allyhawkins:

Here I'm just initiating the skeleton for the module for cell typing Ewing's samples from SCPCP000015. 
I have not added any code yet. 

Towards #373 

To commit the initial files needed for the cell typing module and fill out some initial information in the README file. 

#### Briefly describe the general approach you took to achieve this goal

I ran the `create-analysis-module.py` script using the `--use-renv` flag. I intend to do most analysis in R and will add any python specific files if I end up using python. 

I also briefly updated the README, but did not include a ton of information yet. I plan on filling out more details as I add steps to the module. 

Yes, I plan on working through the plan outlined in #292. 

The only file that I modified is the README file. All other files were generated automatically using the creation script, so review should focus on the README. 

### DIFF
--- a/analyses/ewings-cell-type/notebook-template.Rmd
+++ b/analyses/ewings-cell-type/notebook-template.Rmd
@@ -1,0 +1,88 @@
+---
+title: "ewings-cell-type Analysis"
+author:
+  - Your name here
+date: "`r Sys.Date()`"
+output: html_notebook
+---
+
+
+## Introduction
+
+This a template for an analysis notebook using R Markdown.
+
+It is a good idea to start with a brief introduction to the analysis, including the purpose of the analysis, the data used, and the methods applied.
+Replace this text with your own introduction, and be sure to update the `title:` and `author:` fields in the YAML header at the top of the document.
+
+Don't forget to rename this file as well!
+
+
+## Setup
+
+### Packages
+
+Load required packages in the following chunk, if needed.
+Do not install packages here; only load them with the `library()` function.
+
+```{r packages}
+# load required packages
+# e.g. library(tidyverse)
+```
+
+You may also want to use the `package::function()` syntax within your code for better tracking of which functions come from which packages, and to avoid conflicts between functions with the same name in different packages.
+
+### Paths
+
+Setting paths to directories and files for input and output at the start of a notebook can be helpful for organization and reproducibility.
+You may not know all of the paths you need when you start a notebook, but once you do, keeping them all in one place is helpful, both for yourself in the future and for other users of the notebook.
+
+
+#### Base directories
+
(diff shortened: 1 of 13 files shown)


analyses/ewings-cell-type/notebook-template.Rmd
jashapiro, line 1: I don't know that you need to commit this file now: as a template, it probably doesn't have much value until you add something to it.

analyses/ewings-cell-type/renv.lock
```diff
@@ -0,0 +1,26 @@
...
+      }
+    ]
+  },
+  "Bioconductor": {
+    "Version": "3.18"
+  },
+  "Packages": {
+    "renv": {
+      "Package": "renv",
+      "Version": "1.0.3",
```
jashapiro, line 17: Just noting we should probably update this in the base image.

jashapiro (approved):
LGTM, with the only real comment that I probably wouldn't include the template notebook in the initial commit.

analyses/ewings-cell-type/notebook-template.Rmd
allyhawkins: So this is a good point, but it's created when we create the analysis module? I just committed all the files that are produced by that script, should we not have users do that?

jashapiro: Yeah, I don't know. All the other files seem like they are worth having as placeholders, but just this one seemed maybe not required. Probably worth a discussion of expectations with everyone.

jaclyn-taroni: If they don't commit it and are using GitKraken, depending on the context, I expect one of two things will happen: it will sit as an unstaged file or be stashed automatically.

I would expect it's more straightforward to commit _everything_. I'm not worried about diffs in the future–someone would presumably rename the notebook and wildly alter the content, so I expect it would show up as a deletion of the template + addition of the "new" notebook.

jaclyn-taroni: That being said, I think it's fine to instruct people to delete the template notebook if they're not going to use it.

sjspielman: >  I expect one of two things will happen: it will sit as an unstaged file or be stashed automatically.

jaclyn-taroni: Wouldn't it be easier to instruct people to copy the file from `templates/rmarkdown` (or equivalent)?

sjspielman: > Wouldn't it be easier to instruct people to copy the file from templates/rmarkdown (or equivalent)?

Yes, this is the only answer... 🙃

Issue #374, opened 2024-04-18
Explore manually classifying tumor/normal cells in a single Ewing sarcoma sample


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/292

### Describe the goals of the changes to the analysis module.

After completing #373, I will have a list of marker genes that can be used to classify cells as either tumor or normal. Here, I want to create an exploratory analysis notebook that uses that marker gene list to annotate cells in a single Ewing's sample. 

### What will your pull request contain?

An exploratory analysis notebook that uses a few methods to annotate tumor/ normal in a single Ewing's sample. 
The first section should just include some exploration of the sample, specifically plots showing the expression of each of the marker genes to check that we see variation in these genes. One caveat of all of this is that we may not have any normal cells and they may all be tumor cells. 

Then we can test using those marker genes to annotate cells as either tumor or normal. 
Depending on how long CellAssign takes with a small list of marker genes, I would prefer to use that approach. Alternatively, we could try [`scType`](https://github.com/IanevskiAleksandr/sc-type/tree/master), which we have used in the past. 

Another option is to look for cells that do not express any of the tumor marker genes or have less than a certain threshold and label those as "normal" cells. Then we could use those cells as a reference with InferCNV to see if we can identify tumor cells. 

I then would want to compare the annotations from a marker gene based approach to InferCNV. 

### Will you require additional software beyond what is already in the analysis module?

This will probably involve adding new packages to the conda environment, such as `anndata` and `scvi`. 
We will also need to install `infercnv` to be able to use in R. 

### Will you require different computational resources beyond what the analysis module already uses?

I would like to keep the computational cost here low since it's just an exploratory analysis. The goal is to run this on a laptop. 

### If known, when do you expect to file the pull request?

I do not plan to start on this until #373 has been completed and approved.

allyhawkins:
I am updating this issue to reflect that this will cover only manual annotation of Ewing's cells as tumor/normal cells. I have filed separate follow up issues to look at other methods for annotation, including `CellAssign` and copy number inference.

allyhawkins:
I'm going to close this out for now since we have "explored" manual assignment with marker genes. The `CellAssign` and CNV assignment has been captured in other issues. Any more changes or updates to the manual assignment will require new issues.

PR #376, opened 2024-04-22
Add initial list of marker genes for Ewing sarcoma tumor cells

### LINKED ISSUE #373: Initiate module for cell typing Ewing sarcoma cells and compile a list of marker genes used to identify tumor cells

allyhawkins:

Closes #373 

This PR adds in a TSV file with a list of marker genes that have been used to classify Ewing sarcoma tumor cells as tumor cells in the literature. The goal is to be able to use this list of marker genes to assist in identification of tumor cells in the Ewing's samples. 

To create this TSV file I looked for marker genes that were used to identify tumor cells in published single-cell RNA-seq studies of Ewing sarcoma tumors. I went through the list of publications that I had compiled in https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/292, which all had single-cell RNA-seq of either Ewing's cell lines or patient tumors. I ended up focusing on the two studies that do single-cell RNA-seq on patient tumors rather than cell lines. Those papers had distinct lists of genes that were used to identify tumor cells, many of those genes fit with my expectations of what the markers would be based on my biological knowledge about the disease. The other papers I think will be more helpful when creating marker gene lists for cell states within the tumor cells.  

For https://doi.org/10.1158/2767-9764.CRC-23-0027, I used the genes listed in [Supplemental Methods under `Cell annotation using external data`](https://aacr.silverchair-cdn.com/aacr/content_public/journal/cancerrescommun/3/10/10.1158_2767-9764.crc-23-0027/1/crc-23-0027-s01.pdf?Expires=1716815671&Signature=JgkH5QWs1J0dVeZFTCnZzfi2eDDQHhGsYazaF2UGI5vhPIatv7egTlW-3974FeZ~hM5Ac0gqiq0oBqeenDNWgN0PbU30N9-8n0LrsUmLHA2kCzhBxiHv8D-eKPXL0dSA9H3NKb~PkqTvOeV2QSde~MpCQL~cFh2Yn2Uf9XtAwrXrGhL3nOLIdb~qWBWtQUf1u0w~zynaDsz1hMh32gV5dEQVvZ0XZ4ozc9NO8neUbNvRJk5SSgAIPmQk1WXHKZpIheUbGU2VI2HeggkprVHO0lUELcgb1MmtziupgAkjE-xRrRUtbPVLe8QTWny6QtgJ6IcoH120LbFtRIfL9nOpuA__&Key-Pair-Id=APKAIE5G5CRDK6RD3PGA). They mention other genes in the text and figures that were highly expressed in tumor cells, but I thought it would be better to start with the specific list they mention in the methods. 

For https://doi.org/10.1101/2024.01.18.576251, they mention a core set of genes in the text and then point to Supplemental Figure 2C that contains the genes that were unique to Ewing sarcoma cells. They mention different genes that are associated with different cell states throughout the text, but I chose to only use the ones that were expressed across all Ewing's cells over other normal cells in the samples. 

One thing to keep in mind is that both papers used multiple methods to validate tumor cell identification, which included marker genes, CNV inference, and expression of EWS/FLI1 target genes. There are a few EWS signatures in MsigDB that we might want to use to help validate tumor cells. However, there's also a lot of literature that notes differing expression of EWS targets in tumor cells, including some tumor cells that have a repressed EWS signature, so we should proceed with caution there.

The next PR should be to start the exploration notebook and address #374. 


analyses/ewings-cell-type/references/tumor_marker_genes.tsv
```diff
@@ -0,0 +1,12 @@
+cell_type	gene_symbol	ensembl_gene_id	source
+tumor	CD99	ENSG00000002586	10.1158/2767-9764.CRC-23-0027
```
jashapiro, line 2: In looking at the references, `CD99` was also found in 10.1101/2024.01.18.576251? I don't know if you want to have multiple sources on each line, or repeated lines, but I think you should probably capture when a gene is supported by more than one reference somehow. (I'd probably do the tidier repeated lines with 1 ref per line, and then deduplicate in code)

Update: I see now where you got this specific list, which does not include CD99, though there is a slightly later sentence:
> We found that compared to other cell types in the primary tumor, Ewing cells across all samples express moderate to high levels of PAPPA but low levels of most HLA genes. For immunotherapeutic targets, Ewing sarcoma cells uniquely express higher levels of ADGRG2, IL1RAP, and LINGO1. CD99 and STEAP1 is also highly expressed in Ewing sarcoma cells, although there is significant expression in CAFs as well.

I don't know if you want to include any of those extra genes too? There is also a longer list in Supp table 2, but I don't understand how that list was reduced to the lists in text

jashapiro: From 10.1101/2024.01.18.576251: "Among the Ewing sarcoma cells, a proliferating population was identified by the expression of MKI67 and PCNA (Fig. 1C-D)." Would you want to include such subtyping?

jashapiro (commented):
This looks like a good start. I think the main thing I would suggest is that you add some of the information that you put into the PR text about how you decided which sets of genes to include somewhere in the repository itself. That might be a methods or readme document in the `references` directory or something like that? Somebody coming back to this (you, writing it up) is going to have a lot of questions about how you parsed each reference, and having to search github to find your logic is not going to be very user-friendly!

analyses/ewings-cell-type/references/tumor_marker_genes.tsv
allyhawkins: > Update: I see now where you got this specific list, which does not include CD99, though there is a slightly later sentence:

So CD99 has been the historic marker of Ewing's cells, but it is also expressed at a slightly lower level in normal cells. Fig 1F in 10.1101/2024.01.18.576251 shows that in this particular set of patients its expressed at varying levels across all cells. I got the list I used for this paper specifically from Fig 1C because there are a lot of sentences in the text that mention various genes that were used. However, looking at 1F a lot of those genes are either only in a subset of patients. 

> I don't know if you want to include any of those extra genes too? There is also a longer list in Supp table 2, but I don't understand how that list was reduced to the lists in text

I'm trying to keep this initial list relatively short and only stick to genes that are consistently expressed across multiple patients. I know this is hard because the dataset we have is probably going to have differences, but I want to use CellAssign and the longer the list the more resources that takes. Looking at Figure 1F, maybe we add in ADGRG2, but LINGO1 and IL1RAP are not consistently expressed. 

> From 10.1101/2024.01.18.576251: "Among the Ewing sarcoma cells, a proliferating population was identified by the expression of MKI67 and PCNA (Fig. 1C-D)." Would you want to include such subtyping?

Yes definitely! In this initial list I just wanted to do things that marked tumor cells, but then we should create a list that marks tumor cell states. There are a lot more cell states that have been identified besides this, so that will be a more comprehensive list and I think should be part of a separate review.

jashapiro: This all makes sense! I wrote all that as I was reviewing, and I agree that a lot of it is for potential future directions. I agree especially with keeping the initial list compact. 

As I said in the overall comments, the main thing is to have some documentation of the decisions in the repo itself!

allyhawkins:
Thanks for the review @jashapiro. 

- I changed to using the full URLs for the source of the marker genes. 
- I did end up adding another line for CD99 since it is in both papers. 
- I also added in ADGRG2 based on Fig. 1F. 
- And then I added a README file for the references directory, which includes a description of these marker genes. 

This should be ready for another look!

analyses/ewings-cell-type/references/README.md
```diff
@@ -0,0 +1,16 @@
...
+This directory contains reference files used for cell typing Ewing sarcoma samples from SCPCP000015.
+
+## Tumor marker genes
+
+`tumor_marker_genes.tsv` contains a list of marker genes used to identify tumor cells in previously published studies.
+Currently, this list only contains genes that should mark all tumor cells and we are not yet considering subsets of tumor cells.
+
+The marker gene list was created using gene lists from two publications:
+
+1. https://doi.org/10.1158/2767-9764.CRC-23-0027 - Marker genes were obtained from the list in [Supplemental Methods under `Cell annotation using external data`](https://aacr.silverchair-cdn.com/aacr/content_public/journal/cancerrescommun/3/10/10.1158_2767-9764.crc-23-0027/1/crc-23-0027-s01.pdf?Expires=1716815671&Signature=JgkH5QWs1J0dVeZFTCnZzfi2eDDQHhGsYazaF2UGI5vhPIatv7egTlW-3974FeZ~hM5Ac0gqiq0oBqeenDNWgN0PbU30N9-8n0LrsUmLHA2kCzhBxiHv8D-eKPXL0dSA9H3NKb~PkqTvOeV2QSde~MpCQL~cFh2Yn2Uf9XtAwrXrGhL3nOLIdb~qWBWtQUf1u0w~zynaDsz1hMh32gV5dEQVvZ0XZ4ozc9NO8neUbNvRJk5SSgAIPmQk1WXHKZpIheUbGU2VI2HeggkprVHO0lUELcgb1MmtziupgAkjE-xRrRUtbPVLe8QTWny6QtgJ6IcoH120LbFtRIfL9nOpuA__&Key-Pair-Id=APKAIE5G5CRDK6RD3PGA).
```
jashapiro, line 12: Maybe put a short reference here as well? I also checked, you don't need the full url you pasted in. Most of that seems to be for tracking/login. A good tip for urls like this is to see if you can delete everything from the `?` onward. Usually you can.
```suggestion
1. Visser _et al._ (2023) https://doi.org/10.1158/2767-9764.CRC-23-0027 - Marker genes were obtained from the list in [Supplemental Methods under `Cell annotation using external data`](https://aacr.silverchair-cdn.com/aacr/content_public/journal/cancerrescommun/3/10/10.1158_2767-9764.crc-23-0027/1/crc-23-0027-s01.pdf).
```

analyses/ewings-cell-type/references/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+2. https://doi.org/10.1101/2024.01.18.576251 - Marker genes were determined to be those that were unique to Ewing sarcoma cells in Figure 1C and 1F.
```
jashapiro, line 14:
```suggestion
2. Goodspeed _et al._ (2024) https://doi.org/10.1101/2024.01.18.576251 - Marker genes were determined to be those that were unique to Ewing sarcoma cells in Figure 1C and 1F.
```

jashapiro (approved):
LGTM, with just a little suggestion to add an easier reference for readers and one to remove tokens from a url.

PR #380, opened 2024-04-23
New module: Docs testing

sjspielman:
This PR is being filed strictly for screenshot purposes to support writing docs about responding to code review. 

Please leave the following comments:

1) Inline suggestions in the script
    - suggest to load dplyr in the right place in the script
    - suggest to delete the current dplyr load line  in the script
    - you can also suggest that I actually load readr for real. basically I need at least 2 of these to do a batch, but 3 can't hurt.
2) Inline comment in the script
    - tell me to use `file.path()` in the script. you can leave this comment once or twice for the two spots where I need it, up to you
3) File-level comment for the script
    - tell me that my whole script needs comments
4) An overall "regular" comment
    - Thanks for your contribution, this is a good start, please have a look at the other comments, etc. generic comment.
    - Also please take the time to fill in my module's template README according to instructions in that README, and delete those instructions as I fill it out.

### DIFF
--- a/analyses/cell-type-analysis/scripts/01_read-data.R
+++ b/analyses/cell-type-analysis/scripts/01_read-data.R
@@ -1,0 +1,14 @@
+library(SingleCellExperiment)
+
+sce <- readRDS("../../data/current/SCPCL999990_processed.rds")
+
+library(dplyr)
+
+coldata_df <- colData(sce) |>
+  as.data.frame() |>
+  dplyr::select(barcodes,
+                cluster,
+                singler_celltype_annotation,
+                cellassign_celltype_annotation)
+
+readr::write_tsv("results/sce_annotations.tsv")
--- a/analyses/cell-type-analysis/.Rprofile
+++ b/analyses/cell-type-analysis/.Rprofile
@@ -1,0 +1,1 @@
+source("renv/activate.R")
--- a/analyses/cell-type-analysis/.dockerignore
+++ b/analyses/cell-type-analysis/.dockerignore
@@ -1,0 +1,7 @@
+# Ignore everything by default
+*
+
+# Include specific files in the docker environment
+!/renv.lock
+!/requirements.txt
+!/environment.yml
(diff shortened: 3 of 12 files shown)


analyses/cell-type-analysis/scripts/01_read-data.R
```diff
@@ -0,0 +1,14 @@
+library(SingleCellExperiment)
+
+sce <- readRDS("../../data/current/SCPCL999990_processed.rds")
+
+library(dplyr)
```
allyhawkins, line 5: If we're using `readr`, we won't need to load this. 
```suggestion
```

analyses/cell-type-analysis/scripts/01_read-data.R
```diff
@@ -0,0 +1,14 @@
+library(SingleCellExperiment)
+
```
allyhawkins, line 2: Be sure to load in the `readr` package first! 
```suggestion
library(readr)
```

analyses/cell-type-analysis/scripts/01_read-data.R
```diff
@@ -0,0 +1,14 @@
...
+sce <- readRDS("../../data/current/SCPCL999990_processed.rds")
```
allyhawkins, line 3: I would recommend using `file.path()` here to define the path before reading in the file.

analyses/cell-type-analysis/scripts/01_read-data.R
allyhawkins, line 1: Don't forget to add in comments and document the steps in this script!

allyhawkins (commented):
Thanks for getting this started. I left a few comments that will need to be addressed. 

Additionally, be sure to update the module's `README.md` file with information specific to this module. You can delete any text from the template and replace it with text relevant to your module.

analyses/cell-type-analysis/scripts/01_read-data.R
sjspielman: Thanks for this reminder! I will add a separate section in the script that uses `file.path()` to define input and output files.

sjspielman:
All screenshots have been grabbed. This PR is no longer needed.

PR #366, opened 2024-04-11
Add a CODEOWNERS file

jashapiro:
Closes #286

Here I am adding a codeowners file to allow automatic assignment of reviewers for new submissions. Mostly this is pretty straightforward, but I had to make some decisions that might be worthy of more discussion, so I am filing as a draft.

My general thought was that we would include all data science team members as owners for most of the repo, which would mean that we will all get assigned as reviewers for any new module. Which should be fine, but could get annoying. 
In general, having multiple people as owners will mean that everyone gets more review requests, but it is simple enough to unassign oneself, so it should be okay?

- Administrative and policy files will go to @jaclyn-taroni

- Docs reviews will go to @dvenprasad and @sjspielman (do we want both to be requested?)

- As other modules are developed, an individual owner would be chosen and added to each module: I added myself for the current modules as I have those.

One remaining question is if we want to use github teams to remove some of the overhead of adding and removing individual users when we assign many people. 4 is not too much to manage, but it could be useful for other github features as well, so I think we might want to consider it.

PR #382, opened 2024-04-24
Update CODEOWNERS

sjspielman:
This is a quick PR to update the CODEOWNERS file (I did it all in GitHub directly, hence not filed from fork). 

I was in the process of filing a docs PR, and Deepa was automatically set up as the reviewer. I realized we did not update CODEOWNERS based on this conversation - 

https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/366#issuecomment-2056809657
https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/366#issuecomment-2056841163

So, it is thusly updated. I think we might also want to update the next lines (mkdocs, readme, contributing) to default to me too? What do you think?

Something I can't tell though is whether Deepa was the reviewer because she's first, or because I filed it and therefore GitHub was smart enough to not request that I self-review. My guess is the latter.

Noting also for posterity that as I file this PR, indeed Jackie is already noted as the reviewer as expected!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,7 +5,7 @@
 /.github/CODEOWNERS @jaclyn-taroni
 
 # Primary reviewers for docs and licenses changes
-/docs/ @dvenprasad @sjspielman
+/docs/ @sjspielman
 /CONTRIBUTING.md @dvenprasad @sjspielman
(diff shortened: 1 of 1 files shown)


jaclyn-taroni (approved):
I also think it's because you can't self-review, based on what we've seen with Ally's PRs.

PR #387, opened 2024-04-25
Typos for spellcheck testing

### LINKED ISSUE #385: Check the spellcheck action
In https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/381, we're also modifying the spellcheck action so that when it runs on docs PRs, it only checks docs files.

The syntax is fine and it indeed checks docs files in that PR, but we do want to make sure it behaves as expected in other situations. 

This issue tracks testing it out after #381 goes in.

sjspielman:
Closes #385 

Here I'm testing the spellcheck action with some intentional typos. If changes are needed, I will make them and mark this for review.


sjspielman:
On the first commit, I added typos to both `docs` and top-level. Only the `doc` typo was flagged, as expected. Now I'll fix the `docs` typo, and if all goes well then the top-level typo will get flagged.

sjspielman:
As expected, spellcheck is not run when there is no docs typo.

sjspielman:
Typos were caught as expected in a manual workflow run, so we're all set and no changes are needed.

PR #391, opened 2024-04-25
Annotate tumor cells in Ewing's sample manually and with CellAssign

allyhawkins:

Related to #374 

Here I'm adding a new analysis notebook that attempts to classify tumor and normal cells in a single Ewing sarcoma sample, SCPCS000490. I use the marker genes list that was added in #376 to classify tumor cells manually and automatically with `CellAssign`. 

The main goal of this PR is to see if using `CellAssign` with a list of marker genes for tumor cells is able to accurately identify tumor cells vs normal cells. 

The first section I did some initial exploration of the sample that I was working with. 

- I looked at the expression of all the marker genes across all cells and noted that there is both varying expression within the sample of each gene but also varying expression of each of the marker genes. Some show very striking differences between sets of cells, while other are lowly expressed through most cells. 
- I then took the mean expression of all marker genes and plotted that on a UMAP. I see that there is a distinct group of cells with higher expression of marker genes than all other cells. I anticipate that these are the tumor cells. 
- I would expect that tumor cells and normal cells will cluster separately from each other. So I looked at the cluster assignments present in this sample. 
- Interestingly, there seemed to be two distinct types of clusters, those with high marker gene expression and those with low marker gene expression. I used the distributions of marker gene expression in the cluster to define a cutoff for classifying tumor cells. Any clusters whose median of the mean gene expression distribution was 0.75 was classified as tumor and anything less than 0.75 was normal. I know this is not going to be reproducible across samples since the cutoffs will probably vary, but for this sample I think it was a good starting point to anticipate which cells I would expect to be classified as tumor. 

The second section looks at the results from running `CellAssign` with the marker gene list as the reference. 

- I first ran `CellAssign` using the entire list of marker genes and although it was very fast (less than a minute 🎉 ) it just classified everything as "other" or normal cells. This seems very improbable since this is a tumor sample. 
- I was just curious if I could prune the list of marker genes and get more specific markers that might help guide `CellAssign` so I re-ran it with a filtered gene list. There were a few genes that had very low expression in this sample so I removed those and ran it again. This time a few hundred cells were classified as "tumor" but these were the opposite of the cells that we classified as tumor manually. Also looking at the gene expression for marker genes in tumor vs other, there seems to be higher expression in the "other" cells. 

This didn't entirely make sense to me and I wonder if this just isn't the appropriate use for `CellAssign`. We have never tried to do just one cell type or "other" with such a small gene list. Perhaps it would be better used if we had subtypes of tumor cells we were looking for? 

This PR is just the start of looking at assigning tumor and normal cells. Based on the results here, I would conclude that using `CellAssign` this way is not the most appropriate use. 

In subsequent PRs I would like to use copy number inference methods to identify tumor cells including [`copyKat`](https://github.com/navinlabcode/copykat), [`inferCNV`](https://github.com/broadinstitute/inferCNV/wiki), and [`SCEVAN`](https://github.com/AntonioDeFalco/SCEVAN). Some of these require at least a group of cells to be annotated as "normal". 
We already have annotations from `CellAssign` and `SingleR` generated from `scpca-nf`, so I think we can look for any cells that are identified to be the same "normal" group of cells and use those as the baseline. I would also expect that these cells would line up with the cells I manually am annotating as normal. 

`s3://researcher-211125375652-us-east-2/ewings-cell-type/results`

There are two predictions files that are produced from running `scripts/run-cellassign.py`. These are needed to render the notebook and the code to reproduce those files are in the notebook. 

To generate the predictions files, you will need to first create and activate the `openscpca-ewings-cell-type` environment. Then you can run `scripts/run-cellassign.py` locally. 

As part of this PR: 

- I updated both the main README and the `references` README. 
- I had to make some changes to the `environment.yml` to install `scvi-tools` with `pip` instead of `conda`. 
- I added an Rproj file and updated the `renv.lock` file. 

I've also attached a rendered copy of the notebook for easy review: 

[01-marker-gene-classification-cellassign.html.zip](https://github.com/AlexsLemonade/OpenScPCA-analysis/files/15122705/01-marker-gene-classification-cellassign.html.zip)


jashapiro (commented):
This looks like a good start to the exploration; I have a number of suggestions for the next steps, some of which would be good here and some of which would be genuine next steps.

The first thing is I would separate the CellAssign and manual analysis into separate notebooks. I think there is more you may want to do in both, and keeping them separate will potentially make things a bit easier to follow.

On the manual front, I suggested transforming the marker genes with something like a z transform. Since cell numbers among groups won't be consistent, this might not be great long term either, but having no transformation at all seems like it would be potentially misleading. 

Similarly, I would like to see separate plots for all of the marker genes. It seems to me that part of what you might want to do here is evaluate the utility of the marker genes from other publications, and I don't think we can do that from the means alone.

Also for the "manual" section I think I would like to see the results of FindMarkers or something similar for the clusters that we have: Do those line up with the marker list in a reasonable way? Relatedly, can you use that info to determine what the other clusters might be?

On that front, it would similarly be useful to see what the automated annotation tools did here as a baseline. Do they have assignments that align with the clusters in any meaningful way? I know that some of that be in the QC report, but reproducing it here would be very useful for reference, and to see how the marker genes work with those classifications, not just the automated clustering (I might also redo some clustering here:  relying on our base parameters seems like it might not be the best place to start. I might give [DBscan](https://rdrr.io/github/LTLA/bluster/man/DbscanParam-class.html) a chance, partly because I have an idea that this kind of clustering might be more useful for tumors.)

The last points might allow you to create a better reference for the CellAssign testing. I would expect you would still want immune cells and/or blood cells and/or bone (or other tissue depending on the sample: one is lung?) I think that exploration could be left for a separate PR, but I would definitely do it: I'm curious if your intuition about CellAssign failing because of only 2 cell types is correct.

allyhawkins:
I made quite a few changes to this PR based on your comments. Specifically, I am re-focusing this on just marker gene exploration including classifying cells based on marker gene expression or classifying clusters on marker gene expression. I included plots that will hopefully help us address the usefulness of each of the marker genes, but I am hesitant to remove any from the list without looking across multiple samples. I do expect the expression of these genes to vary by sample. 

The other thing I added was a comparison to SingleR and CellAssign annotations. For the most part they line up with what we are manually assigning as tumor vs. normal cells. 

I did also assign tumor/normal by getting the z-score for each gene and classifying any cell that had expression of at least one marker gene > 0 as a tumor cell. I switched from using the mean to using the sum of the z-scores. I tested both and there wasn't a noticeable difference and sum meant that any cell that expressed at least one of the marker genes was classified as tumor. I don't think we need to stick to this, but this at least estimates which cells I would think would be tumor cells. 

I also did not test other cutoffs (e.g., z-score > 1), which we definitely could do if we want to be stricter, but I think for now the 0 cutoff is good. 

I am planning on filing issues to address clustering and running CellAssign separately.

Here's a copy of the rendered report: 
[01-marker-gene-tumor-classification.html.zip](https://github.com/AlexsLemonade/OpenScPCA-analysis/files/15170391/01-marker-gene-tumor-classification.html.zip)

A copy of the supplemental cell type report for this sample for reference: 
[SCPCL000822_celltype-report.html.zip](https://github.com/AlexsLemonade/OpenScPCA-analysis/files/15170397/SCPCL000822_celltype-report.html.zip)

And then I replied to your statements above below.  

> The first thing is I would separate the CellAssign and manual analysis into separate notebooks. I think there is more you may want to do in both, and keeping them separate will potentially make things a bit easier to follow.

I took out the `CellAssign` stuff for now and will file a separate PR with that notebook. I think this should make it easier for review. When I first started this, I was planning to mostly use `CellAssign` to classify just tumor cells, but that approach clearly didn't work so we should revisit that in another PR. 

> On the manual front, I suggested transforming the marker genes with something like a z transform. Since cell numbers among groups won't be consistent, this might not be great long term either, but having no transformation at all seems like it would be potentially misleading.

This is a good point! I went ahead and scaled each of the marker gene expression vectors and use that scaled value for the majority of the notebook. Here you can see pretty clear groups of cells with either > 0 or < 0 expression. This makes it easier to use an actual cutoff. 

> Similarly, I would like to see separate plots for all of the marker genes. It seems to me that part of what you might want to do here is evaluate the utility of the marker genes from other publications, and I don't think we can do that from the means alone.

I did have UMAPs with each of the marker genes in there as the first plot, but I've gone ahead and expanded this even more to make more plots show individual gene expression of marker genes. This includes: 

- Density plots looking at each marker gene, both the log counts and the transformed score. 
- Density plots looking at marker gene expression in cells that are manually annotated as tumor vs. normal. This plot I think is helpful in seeing which genes might not be great marker genes for this sample. 
- Stacked bar chart to show how many cells in each cluster express a given marker gene (z-score > 0). This is separated by tumor/ normal cells to help show specificity of each marker gene. 
- Density plot looking at individual marker gene expression in cells classified as Normal (endothelial) vs. tumor (muscle) using SingleR and CellAssign. 


> Also for the "manual" section I think I would like to see the results of FindMarkers or something similar for the clusters that we have: Do those line up with the marker list in a reasonable way? Relatedly, can you use that info to determine what the other clusters might be?

So I didn't look at marker genes for each cluster, but I did look at marker genes for tumor vs. normal cells after manual annotation. This should validate which marker genes are important in distinguishing those two groups of cells. I actually wouldn't expect very many tumor marker genes to be markers for individual clusters if comparing that cluster to all other cells. I don't think that we should expect all tumor cells to be within the same cluster, but we should expect that they cluster separately from normal cells. There are definitely cell states within the tumor cells that could be leading to different clusters, so the marker genes for those clusters won't necessarily be the same as the marker genes for all tumor cells. 

In regards to determining the other clusters, I am trying to just pull out tumor cells right now. I think the question of what other cell types might be present is a later question. The overall approach is to first just create a label indicating tumor vs normal and then we can annotate the normal cells separately. I would imagine just being able to use SingleR for annotating the normal cells. 

> On that front, it would similarly be useful to see what the automated annotation tools did here as a baseline. Do they have assignments that align with the clusters in any meaningful way? I know that some of that be in the QC report, but reproducing it here would be very useful for reference, and to see how the marker genes work with those classifications, not just the automated clustering (I might also redo some clustering here: relying on our base parameters seems like it might not be the best place to start. I might give [DBscan](https://rdrr.io/github/LTLA/bluster/man/DbscanParam-class.html) a chance, partly because I have an idea that this kind of clustering might be more useful for tumors.)

First, for clustering, I think finding the most "optimal" clustering is a totally different question that should be addressed elsewhere. For this, I think working with the original clusters is a good starting point. But actually assigning clusters that we are confident in will take a lot more exploration that I think is outside the scope of this PR. I think that's another notebook. 

For comparing to previous annotations, I did add a section that looks at how tumor/normal cells align with SingleR and CellAssign annotations. For the most part, we see that tumor cells are classified as muscle and normal cells are classified as endothelial and macrophages. The fibroblasts identified by SingleR encompass both tumor and normal and in CellAssign all fibroblasts are normal cells. 

> The last points might allow you to create a better reference for the CellAssign testing. I would expect you would still want immune cells and/or blood cells and/or bone (or other tissue depending on the sample: one is lung?) I think that exploration could be left for a separate PR, but I would definitely do it: I'm curious if your intuition about CellAssign failing because of only 2 cell types is correct.

I'm going to file another issue to specifically look at CellAssign in another notebook. The reason I didn't use immune cells here is because I am explicitly interested in identifying tumor cells. Also, the more marker genes that we include the longer CellAssign takes and I'm trying to avoid using computationally expensive methods to assign these cell types.

jashapiro:
> In regards to determining the other clusters, I am trying to just pull out tumor cells right now. I think the question of what other cell types might be present is a later question. The overall approach is to first just create a label indicating tumor vs normal and then we can annotate the normal cells separately. I would imagine just being able to use SingleR for annotating the normal cells.

My thought here was mostly that CellAssign might perform better if given other specific options of cell types with their own markers. Which is to say that I am not sure that trying to identify the tumor cells first, in isolation, is going to be the best approach. But we'll come back to that in #401, where you already have that outlined!

allyhawkins:
Just noting that I made a small update here to save the output tumor/normal classifications to a tsv file. This should now be present on S3 in `s3://researcher-211125375652-us-east-2/ewings-cell-type/SCPCL000822_tumor_normal_classifications.tsv`.

analyses/ewings-cell-type/01-marker-gene-tumor-classification.Rmd
```diff
@@ -0,0 +1,516 @@
...
+  dplyr::left_join(transformed_exp_df, by = c("barcodes", "gene_symbol"))
+```
+
+Now we can create the same density plot but looking at z-scores. 
+
+```{r}
+ggplot(umap_df, aes(x = transformed_gene_expression, fill = gene_symbol)) +
+  geom_density() +
+  theme_bw() +
+  facet_wrap(vars(gene_symbol))
```
jashapiro, line 186: Since the fill is just for show, no need for the legend
```suggestion
  facet_wrap(vars(gene_symbol)) +
  theme(legend.position = "none")
```

analyses/ewings-cell-type/01-marker-gene-tumor-classification.Rmd
```diff
@@ -0,0 +1,516 @@
...
+```
+
+Looking at this plot, there is definitely some variation in the expression of marker genes. 
+It looks like the highest gene expression is seen in the larger group of cells on the bottom right of the UMAP. 
+
+```{r}
+ggplot(umap_df, aes(x = gene_expression, fill = gene_symbol)) +
+  geom_density() +
+  theme_bw() +
+  facet_wrap(vars(gene_symbol))
```
jashapiro, line 155:
```suggestion
  facet_wrap(vars(gene_symbol)) +
  theme(legend.position = "none")
```

analyses/ewings-cell-type/01-marker-gene-tumor-classification.Rmd
```diff
@@ -0,0 +1,516 @@
...
+Now we can create the same density plot but looking at z-scores. 
+
+```{r}
+ggplot(umap_df, aes(x = transformed_gene_expression, fill = gene_symbol)) +
+  geom_density() +
+  theme_bw() +
+  facet_wrap(vars(gene_symbol))
+```
+
+It looks like some marker genes have distinct groups of cells with z-score > 0, while other marker genes may not be as informative.
```
jashapiro, line 189: I'm actually kind of impressed how well 0 splits the bimodal distributions!

analyses/ewings-cell-type/01-marker-gene-tumor-classification.Rmd
```diff
@@ -0,0 +1,516 @@
...
+    theme(
+      aspect.ratio = 1,
+      legend.position = "bottom",
+      axis.title = element_text(size = 9, color = "black"),
+      strip.text = element_text(size = 8),
+      legend.title = element_text(size = 9),
+      legend.text = element_text(size = 8)
+    ) +
+    guides(colour = guide_colorbar(title.position = "bottom", title.hjust = 0.5)) +
+  theme_bw()
```
jashapiro, line 145: Just a note that rather than adding `theme_bw` to every plot, you might add the following to a setup chunk:

```
theme_set(theme_bw())
```

analyses/ewings-cell-type/01-marker-gene-tumor-classification.Rmd
```diff
@@ -0,0 +1,516 @@
...
+We can then look at marker gene expression across clusters.
+Below we visualize the total number of cells in each cluster that are classified as tumor or normal using the sum of marker gene expression. 
+Then for each set of cells, how many of those actually express each marker gene. 
+
+We expect to see two things: 
+
+- Tumor cells and normal cells tend to separate by cluster. 
+- Good marker genes will be expressed in tumor cells but not normal cells. 
+
+```{r, fig.height=10, fig.width=10}
```
jashapiro, line 318: I don't think I have a good suggestion here, but I don't love this plot. My main complaint is mostly that the counts are so small for some of the clusters that it is hard to see the values. But transforming doesn't necessarily help, and using free scales for the facets would be very busy. It may be that this is just too much for one plot, but as this is an exploration notebook, it is probably fine to leave as is.

analyses/ewings-cell-type/01-marker-gene-tumor-classification.Rmd
```diff
@@ -0,0 +1,516 @@
...
+```{r}
+# plot marker gene expression across all clusters 
+ggplot(marker_sum_exp, aes(x = sum_exp, fill = cluster)) +
+  geom_density() +
+  facet_grid(rows = vars(cluster)) +
+  theme_bw()
+```
+
+Looking at this there are definitely clusters with higher sum expression of the marker genes than other cells. 
+If the cluster has a median expression > 0 then all cells in that cluster will get assigned as tumor cells.
```
jashapiro, line 353: This is where I start to get a bit uncomfortable with just using the initial clustering. If you are going to assign all cells in a cluster as tumor or not, then I would want the clusters to be something more meaningful than the defaults. I might just skip this for now, as I don't think it does anything more useful than the cell-based assignment, at least at this stage.

jashapiro (commented):
Looks good overall! I think this is a good level of detail for this phase of analysis.
I do have a few suggestions, but mostly minor and editorial.

The other thing that should happen here is an update to R 4.4/Bioc3.19 to sync up with everything else in the repo going forward. Annoyingly, the process for this can be a bit annoying. I managed these by manually updating the R and Bioconductor versions in `renv.lock` & `renv/settings.json`, followed by some combination of `renv::restore()`, `renv::hydrate()` and `BiocManager::install(version="3.19")` until everything was up to date. Much more hassle than I would have liked!

analyses/ewings-cell-type/01-marker-gene-tumor-classification.Rmd
```diff
@@ -0,0 +1,493 @@
...
+                              groups = sce$sum_classification,
+                              pval.type = "all") 
+```
+
+```{r}
+# pull out top markers for tumor cells 
+tumor_markers <- markers$Tumor |> 
+  as.data.frame() |> 
+  tibble::rownames_to_column("ensembl_gene_id") |> 
+  dplyr::left_join(marker_genes_df) |>
```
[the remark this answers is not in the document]
allyhawkins: No, I'm doing this so that I can print them out with the gene symbols rather than the IDs. Not all of us have them memorized! I'll add a comment here to state that.

analyses/ewings-cell-type/01-marker-gene-tumor-classification.Rmd
allyhawkins: The reason I made this plot is because I wanted to see how many of each cell type actually express the marker gene above our threshold of 0. I think this helps show which marker genes are going to be more specific to tumor cells, although not a complete analysis at all, I do think it's helpful to see. It was very busy though so I updated this plot to only separate by tumor/normal and not include the clusters. Hopefully it's a little bit easier to interpret this way.

allyhawkins:
@jashapiro I have updated this based on your comments so this should be ready for another look. 

- I updated to use R 4.4 and Bioconductor 3.19. To do this I actually just installed R 4.4 and the new version of Bioconductor and then did `renv::init()` again, which allows you to choose an option to update and re-generate the lock file. 
- I removed any of the clustering plots since they don't really give us much and I agree that we should just stick with classifying on a cell level. 
- I updated the plot that showed expression of marker genes across clusters to just look at tumor and normal cells (see https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/391#discussion_r1589383077)

analyses/ewings-cell-type/01-marker-gene-tumor-classification.Rmd
```diff
@@ -0,0 +1,469 @@
...
+  tibble::rownames_to_column("barcodes")
+
+# pull out the UMAP coordinates and make a data frame to use for plotting
+umap_df <- sce |> 
+  scuttle::makePerCellDF(use.dimred = "UMAP") |> 
+  # replace UMAP.1 with UMAP1
+  dplyr::rename_with(
+        \(x) stringr::str_replace(x, "^UMAP\\.", "UMAP"),
+        starts_with("UMAP")
+      ) |>
```
jashapiro, line 117: Just noting that because your regex matches from the start only, you should be able to leave off the `starts_with`
```suggestion
  dplyr::rename_with(
        \(x) stringr::str_replace(x, "^UMAP\\.", "UMAP")
      ) |> 
```

analyses/ewings-cell-type/renv/settings.json
```diff
@@ -1,5 +1,19 @@
...
+  "ignored.packages": [],
+  "package.dependency.fields": [
+    "Imports",
+    "Depends",
+    "LinkingTo"
+  ],
   "ppm.enabled": true,
-  "r.version": "4.3.3"
+  "ppm.ignored.urls": [],
+  "r.version": "4.3.3",
```
jashapiro, line 12:
```suggestion
  "r.version": "4.4.0",
```

jashapiro (approved):
LGTM. Just a couple little things I noticed along the way

PR #394, opened 2024-04-30
Tips for LSfR

sjspielman:
I'm filing this docs PR with a few additions based on my experiences using LSfR as part of OpenRRP prep. It had been a minute since I logged into the aws console to create/launch instances, and this information would have helped me along the way. I did a few spacing things while I was here as well, and I also updated a relative link to point to what I think is a more relevant docs page.


allyhawkins (approved):
👍

PR #406, opened 2024-05-01
Update LSfR docs

### LINKED ISSUE #395: Expand LSfR docs: Need two windows for VSCodium
We've discovered some LSfR behavior that we should add to the docs and modify existing content:

https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/4482e4f161a716281f7e18dc4f7eab122aba4d1d/docs/software-platforms/aws/creating-vcs.md?plain=1#L133

^ This is true, but it's only part of the story:

When you access the _operating system_ for VSCodium instances, you can't access VS Code itself. You need to have two windows open in browser: one launched directly as VSCodium and the other launched as accessing the operating system. You have to go back and forth between these windows to use VS Code and, e.g., GitKraken. There's no need to refresh a window when you work in the other. For example, I can touch a file from the VS Code terminal in the VSCodium window, and I can see it right away in a regular terminal in the OS window.

This situation is not the same for RStudio instances. RStudio _can_ be opened within an instance you launch by accessing the operating system. So, this is only a VSCodium consideration.

### LINKED ISSUE #396: LSfR docs: Stop the computer
It looks like we do not recommend anywhere that contributors formally stop their computers after a session. It's been my understanding (?) that we do want to make sure computers are stopped in between coding sessions to avoid extra charges. If I'm mistaken here please let me know, but otherwise we should include in the "Developing on LSfR" doc that they should stop the computer in between coding sessions.

sjspielman:
Closes #395 
Closes #396 

This PR updates a LSfR docs in a few spots:

1) For #395, I expanded instructions on launching an instance to clarify how to use VSCodium along with other apps. I also made the decision (please let me know what you think!) of creating a new docs page _just for_ accessing virtual computers. This makes it easier imo to find those accessing instructions again, since contributors will need to use them beyond just the initial creation of a virtual computer. As someone in this boat over the past 2 days working on LSfR (as part of OpenRRP prep), I would have appreciated it too!

2) For #396, I added some information about shutting down the computers into the developing for LSfR doc.


docs/software-platforms/aws/accessing-vcs.md
```diff
@@ -0,0 +1,50 @@
...
+
+    2. If applicable, use the `Launch RStudio` and/or `Launch VSCodium` button.
+        - Clicking this button will launch _only_ the `RStudio` or `VSCodium` application, respectively, and you will not be able to access other applications from the same browser window.
+        - If you need to access additional applications (e.g., GitKraken) you can open a separate window with `Access operating system`.
+        Both of these windows can be open simultaneously.
+        - Note that because the RStudio and VSCodium applications have built-in terminals, you can always use the terminal directly from them.
+
+        ??? tip "Working with `VSCodium` virtual computers"
+            When you launch a `VSCodium` computer via `Access operating system`, you will _not_ be able to use or access VSCodium itself.
+            You can only access VSCodium if you use the `Launch VSCodium` button.
```
allyhawkins, line 41: I read this and thought there is no way this is true. So I just created an instance just to check and indeed it is true, which is crazy!

allyhawkins (approved):
LGTM!

docs/software-platforms/aws/accessing-vcs.md
sjspielman: Yeah, this was indeed a wild discovery..!!!

PR #407, opened 2024-05-01
Update local conda setup docs

sjspielman:
Closes #404 

This PR updates conda setup docs so that local setup also creates an `openscpca` environment. I ended up making this into its own subsection to get all the details across in a way I thought was a bit more organized. Let me know what you think of the organization and of the content! I acknowledge that there are many admonitions, but since one is collapsed by default I think it's not too much?

PR #408, opened 2024-05-02
Make environment specs minimal

jashapiro:
Based on [my comment in a pr](https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/407#discussion_r1587667358) and seeing updates to things like `pre-commit` and `aws`, I thought we might want to make the base environment specification a bit looser, so I changed each of these tools to use `>=` for the versions. I left the `jupyter` spec alone, as that is really just a virtual package that does not specify the versions of the packages it depends on.

Note that the previous specification was already loose-ish, allowing for bugfix updates, but did not allow for updates from 3.6.x to 3.7.x, for example.


sjspielman (approved):
I agree this is a good idea generally. I rebuilt the environment with this file, and all good!

PR #426, opened 2024-05-10
Add  download-results.py and refactor download-data.py for modularity

### LINKED ISSUE #378: Update download-data.py to support downloading workflow results
The [`openscpca-nf` workflow](https://github.com/AlexsLemonade/OpenScPCA-nf) will generally deposit results files in the `openscpca-nf-workflow-results`. We would like these files to be available for further downstream analysis within analysis modules in this repository. 

We will most likely want to extend the current `download-data.py` script to support downloads from this results bucket, with the same kinds of options we have for downloading the main data, namely the ability to download only the results files from specific analysis modules, and the default to download the latest release (but the ability to download specified releases).

We will also want the ability to download from a public test bucket that contains outputs from running the workflow on the public simulated data. 

The structure of the workflow output directory will be determined in https://github.com/AlexsLemonade/OpenScPCA-nf/issues/15

### LINKED ISSUE #420: download-data.py is inconsistent when downloading multiplexed samples
### Describe the bug

When running `download-data.py` on  most multiplexed samples, no files will be downloaded, because the search prefix we use expects the prefix for those files to end with the id. Multiplexed samples are concatenated with `_`, so only the last id in a set will be found.  For example, specifying SCPCS000131 does not download any files, but the sample that is multiplexed with it, SCPCS000157, does work, as that us the end of the prefix.

### Steps to reproduce

`./download-data.py --samples SCPCS000131` does not download any files when it should!

### Consequences of the bug

No error message, just a silent failure

### Proposed solution

I think we may be able to update the include strings to include not only `*<sample_id>/*_<file_suffix>`, which is what we do now, but also `*<sample_id>_*/*<file_suffix>` to capture all possible positions for the sample id.

jashapiro:
Here I am adding the ability to download results data from the workflow output bucket to the download script.

Closes #378
Closes #420

This turned into a bigger reorganization, as the previous version of this script was not very modular. So now there are a bunch of functions to do various parts of the process. The `main()` function is mostly tasked with parsing arguments and delegating to the other functions.

The overall approach here is that you can download the results of a specific module by specifying the `--module` or `--module-results` argument and then naming a module or modules in a comma separated list. These results can be refined by project or sample, but not by other filters, as I didn't want to make assumptions about file names. You may want to look at how I am doing the searches for those identifiers. They may need refinement, but they do work for the test case I have for now...

Results files are downloaded to `data/<release>/results/<module>` and organized however they were organized in the results bucket. But note that the bucket itself organized without the "results" subdirectory, as that would be redundant. However, I figured we could use the same bucket for the test data and the simulated results, so I gave used the prefix `test-results` there, to contrast separate the modules from the simulated data that is stored in `test`.

You can also list the available results modules with the `--list-results` or `--list-modules` command. (why both? because I kept messing up when testing).

For dry runs I added a bit more output to allow for checking the size of the download. Or at least the total size of the data set, because I didn't want to get into parsing the full output to get the exact list of files to be downloaded and then try to match that up with the `ls` output (which is the only place with the sizes). The main purpose of this is so that people have some idea of the size of the data they are requesting, as some (the merge data, I mean) are big!

Finally, I changed the `--include` option that I never liked to `--process-stage` which is a bit odd for bulk data, but I think it works better overall by not being quite so generic. Our online docs do not directly cover this option, so the only places that needed updating for that change were a couple of readmes.

We _will_ need to document the `--module-results` flag though, but I thought that would require a full separate issue (which I will file shortly!)

Questions for review:
- Does the download structure make sense or do you have any other suggestions on that front?
  - note that I have not actually tested the test data version, as we don't have any results there yet!
- Please do test with `--dryrun` (and without) to see if things are working as expected. Note that the only  projects in the `merge-sce` module at the moment are `SCPCS000007` and `SCPCS000012` and the latter is much smaller than the former!
- do all the arguments make sense in how they work together?
- please let me know if you run into any bugs or think error messages should be cleaned up!
  - When I made things modular, I went back to raising exceptions a few places, which makes things a bit uglier that just exiting. I could catch some of the exceptions, and make things a bit prettier or go back to just exiting...


jaclyn-taroni:
@jashapiro - what do you mean by this? There are two buckets for test data set up already.

> However, I figured we could use the same bucket for the test data and the simulated results, so I gave used the prefix `test-results` there, to contrast separate the modules from the simulated data that is stored in `test`.

jashapiro:
> @jashapiro - what do you mean by this? There are two buckets for test data set up already.
> 
> > However, I figured we could use the same bucket for the test data and the simulated results, so I gave used the prefix `test-results` there, to contrast separate the modules from the simulated data that is stored in `test`.

Yes, I guess there are now. I hadn't looked at how the `*-public-data` were set up, and was only considering the `openscpca--sim-data` bucket that I had been working with in the workflow to this point and its mirror `openscpca-temp-simdata`. 

My thought here was we can probably get away with just the one public bucket, and it might simplify management a bit. But either way is fine, and if the buckets already exist, we can use them!

download-data.py
```diff
@@ -54,35 +367,18 @@ def main() -> None:
...
-        ),
-    )
-    parser.add_argument(
-        "--include-reports",
-        action="store_true",
-        help="Include html report files in the download.",
+        default="",
+        help="The results modules to download."
+        " A comma separated list of results modules to download."
+        "\nWhen this option is selected, the `--process-stage` and `--include-reports` options are ignored.",
```
allyhawkins, line 376: My understanding is that if these are used then you only download the results and not release, so I would specify that in this help message.

download-data.py
```diff
@@ -155,151 +477,124 @@ def main() -> None:
...
-    if (
-        args.test_data or args.release.lower() in ["current", "latest"]
-    ) and not args.dryrun:
-        # update the current symlink
-        current_symlink = args.data_dir / "current"
-        current_symlink.unlink(missing_ok=True)
-        current_symlink.symlink_to(release)
-        print(f"Updated 'current' symlink to point to '{release}'.")
+    ### Download the data ###
+    if not args.module_results:
```
allyhawkins, line 570: Looking at this I really do want a flag that specifies download type... I think the default could be release data, but I like being explicit about what type of data you are trying to get. Some people could think that by having a `--module_results` argument, you are adding download of results to download of a release, but these are separate. 

Another kind of wild thought, but what if we had two scripts... We can pull out the functions you have into a separate `utils` script that gets read in to each of the scripts. Then you have one that's just for downloading results from the workflow and one just for downloading release data.

download-data.py
```diff
@@ -4,20 +4,150 @@
...
+    ls_cmd = ["aws", "s3", "ls", f"s3://{bucket}/"]
+    if profile:
+        ls_cmd += ["--profile", profile]
+    if bucket == TEST_BUCKET:
+        ls_cmd += ["--no-sign-request"]
+    ls_result = subprocess.run(ls_cmd, capture_output=True, text=True)
+    if ls_result.returncode:
+        print(
+            "Error listing release versions from the OpenScPCA release bucket.",
+            " Ensure you have the correct AWS permissions to access OpenScPCA data.",
```
allyhawkins, line 36: Just because I've seen this error message myself, can we update it to remind people to login or use the correct profile since that might be why they don't have access 😬

download-data.py
```diff
@@ -26,12 +156,189 @@ def add_parent_dirs(patterns: List[str], dirs: List[str]) -> List[str]:
...
+    modules: Set[str],
+    data_dir: pathlib.Path,
+    projects: Set[str] = {},
+    samples: Set[str] = {},
+    dryrun: bool = False,
+    profile: str = "",
+    update_current: bool = True,
+) -> None:
+    """
+    Download results for a specific release of OpenScPCA.
```
allyhawkins, line 282:
```suggestion
    Download workflow results for a specific release of OpenScPCA.
```

download-data.py
```diff
@@ -4,20 +4,150 @@
...
+    """
+    Convert a size in bytes to something human readable.
+    """
+    for unit in ["B", "KiB", "MiB", "GiB", "TiB"]:
+        if size < 1024.0 or unit == "TiB":
+            break
+        size /= 1024.0
+    if unit == "B":
+        return f"{size:.0f} B"
+    return f"{size:.2f} {unit}"
```
allyhawkins, line 148: I don't think this is working as you expect? When I try to download the `merge-sce` results module with dry run I see a total size of 0 B, but the files are being listed.

allyhawkins: Update, if I include a `--projects` argument then the size is correct. So this doesn't work if you want to download results for an entire module.

jashapiro: Hmm...I am presumably doing something wrong with the globs, since I am trying to match the behavior of `aws`. Will look at this case!

jashapiro: I believe I fixed this in f1442f3. If you want to test again, that would be great! Thank you for the detailed report, it made it easy to find my error!

allyhawkins: Yes that fix works! I can now run a dry run for an entire module or a specific project within a module and the appropriate files are listed and the size is listed and appears to be correct. 🎉

download-data.py
```diff
@@ -54,35 +371,18 @@ def main() -> None:
         " To switch back, rerun this script with the `--release current` option.",
     )
     parser.add_argument(
-        "--format",
+        "--module-results",
```
jaclyn-taroni, line 374: Is this argument intended to work with `--format`? Because it doesn't appear to be respected in my testing, and the output of `--help` lead me to believe it would be supported.

jaclyn-taroni (commented):
I am returning an observation from testing, which, in combination with @allyhawkins comments elsewhere about the help message, does make me wonder if we should reconsider splitting the results download out to remove complexity? In weighing that question, maintainability and being able to explain this to users should both be considered.

download-data.py
jashapiro: No, it isn't, because I what format the results might take seemed too hard to predict. In this case we have both SCE and Anndata, but I don't expect it to be the norm, so I wasn't sure what to do to handle it at this stage. Maybe a future enhancement to allow more selectivity from results downloads?

jashapiro:
> I am returning an observation from testing, which, in combination with @allyhawkins comments elsewhere about the help message, does make me wonder if we should reconsider splitting the results download out to remove complexity? In weighing that question, maintainability and being able to explain this to users should both be considered.

I think splitting it up won't be too bad. Now that it is more modular, I can either pull out functions to a separate file or even more simply just import one script from the other to get the functions I need. That's the magic of `if __name__ == "__main__":`!

download-data.py
jaclyn-taroni: Then the help text should say something about:

> These results can be refined by project or sample, but not by other filters, as I didn't want to make assumptions about file names.

Rather than:

>  When this option is selected, the `--process-stage` and `--include-reports` options are ignored

jashapiro:
Okay, I have split this up into two separate scripts: `download-data.py` and `download-results.py` where the results script imports from the data script to get some of its functionality.

I also updated the AWS error message to tell people what is probably wrong (profiles and login), and just output the AWS error itself rather than the stack trace. Some other help text was also updated. 

This is probably worth testing one more time before it goes in.

.github/workflows/docker-image.yml
```diff
@@ -1,9 +1,9 @@
 name: Docker Image CI
 
 on:
-  push:
-    paths:
-      - 'Dockerfile'
+  # push:
+  #   paths:
+  #     - 'Dockerfile'
```
allyhawkins, line 6: Just checking did you mean to comment this out?

download-results.py
```diff
@@ -0,0 +1,318 @@
...
+    """
+    ls_cmd = ["aws", "s3", "ls", f"s3://{bucket}/{release}/"]
+    if profile:
+        ls_cmd += ["--profile", profile]
+    if bucket == TEST_RESULTS_BUCKET:
+        ls_cmd += ["--no-sign-request"]
+    ls_result = subprocess.run(ls_cmd, capture_output=True, text=True)
+    if ls_result.returncode:  # authentication errors, usually
+        print(
+            "Error listing release versions from the OpenScPCA bucket.\n\n"
```
allyhawkins, line 36:
```suggestion
            "Error listing modules available in OpenScPCA workflow results bucket.\n\n"
```
I think this should be updated to reflect that you are listing modules?

download-results.py
```diff
@@ -0,0 +1,318 @@
...
+
+def get_results_modules(bucket: str, release: str, profile: str) -> List[str]:
+    """
+    Get the list of available results modules from the OpenScPCA workflow results bucket.
+    """
+    ls_cmd = ["aws", "s3", "ls", f"s3://{bucket}/{release}/"]
+    if profile:
+        ls_cmd += ["--profile", profile]
+    if bucket == TEST_RESULTS_BUCKET:
+        ls_cmd += ["--no-sign-request"]
```
jaclyn-taroni, line 32: Reporting a result I was not expecting (have not looked into it further yet):
```
% ./download-results.py --test-data

Error listing release versions from the OpenScPCA bucket.

Make sure you have the correct profile active (or use the --profile option) and run `aws sso login` before running this script.

AWS Error:  
Unable to locate credentials. You can configure credentials by running "aws configure".
```

allyhawkins (approved):
I did some testing of both scripts and everything looks good! I just had one comment about an error message that needs to be updated, but I don't need to see this again.

download-results.py
allyhawkins: When I run this, I get the following which makes me think it might have something to do with the ability to access the test bucket? 

```
./download-results.py --test-data --dryrun --profile openscpca
Error listing release versions from the OpenScPCA bucket.

Make sure you have the correct profile active (or use the --profile option) and run `aws sso login` before running this script.

AWS Error:  
An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
```

jaclyn-taroni: This works just fine: `aws s3 ls s3://openscpca-test-workflow-results-public-access/test --no-sign-request`

`aws s3 ls s3://openscpca-test-workflow-results-public-access --profile openscpca` returns: 

```
An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
```

Which is what I would expect.

.github/workflows/docker-image.yml
jashapiro: I did for now. It was trying to run in my branch, so we will need to modify it for proper automation later.

download-results.py
allyhawkins: Confirming that this now works as expected with the updates.

allyhawkins (approved):
🚀

Issue #424, opened 2024-05-10
Initiate doublet detection module


sjspielman:
### Please link to the GitHub Discussion for this proposed analysis.

https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/364

### Describe the goals of this analysis module.

The ultimate goal of this module is to identify doublets across ScPCA libraries, where this identification will be represented in the form of a double score (or probability) based on one or more methods. In the longer-term, insights from this module will contribute to expanding the ScPCA pipeline to perform doublet detection and include results in objects for users to leverage in their research. 

### What software will you require?

This analysis will be based in R, and to start, will use the `scDblFinder` and `scds` packages (noting the latter of which contains three complementary approaches for flagging doublets). In the future we _may_ extend to a Python method, but based on the existing literature it seems that the R packages are the way to go (see linked discussion).

### What will your first pull request contain?

The first step in this module will be to perform benchmarking across methods (with real "ground truth" data, if possible) and assess computational requirements for different methods we could use. I plan to use `scDblFinder` and three methods (`cxds`, bcds`, and `hybrid`) from the `scds` package to run on several ScPCA libraries of different sizes and diagnoses. I will also likely explore some parameter space.

### What computational resources will you require?

According to existing literature benchmarks, the methods I'm exploring here are expected to be highly efficient and feasible on a laptop 🤞

sjspielman:
This module was initiated, and this issue is being closed in favor of #446

PR #429, opened 2024-05-10
Initiate doublet detection module

sjspielman:

This PR establishes a module for double detection with a `renv` environment and necessary dependencies, and a light README file which is mostly forthcoming. There is also a notebook whose contents (other than some core module library loads) are also forthcoming! I also added a Dockerfile, which I did test the build for and it indeed builds 👍 

Setting up for #424 

No science yet! Just initiating a module, since I would like a win today 🎉 

One question: Should I add a skeleton GHA file for this module as well, or are we waiting on those?

### DIFF
--- a/analyses/doublet-detection/Dockerfile
+++ b/analyses/doublet-detection/Dockerfile
@@ -1,0 +1,22 @@
+# Base image on the Bioconductor 3.19 image
+FROM bioconductor/r-ver:3.19
+
+# Labels following the Open Containers Initiative (OCI) recommendations
+# For more information, see https://specs.opencontainers.org/image-spec/annotations/?v=v1.0.1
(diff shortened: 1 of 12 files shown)


analyses/doublet-detection/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Base image on the Bioconductor 3.19 image
+FROM bioconductor/r-ver:3.19
```
jashapiro, line 2: Just noting that I discovered that this image unfortunately does not have `pandoc` installed. I recommend adding the following somewhere in here to install it (probably after labels, before renv):

```
RUN apt-get -y update &&  \
    DEBIAN_FRONTEND=noninteractive \
    apt-get install --no-install-recommends -y \
    pandoc \
    && rm -rf /var/lib/apt/lists/*
```

allyhawkins (approved):
LGTM

PR #430, opened 2024-05-10
Update run-cellassign to use raw counts and update notebook with new results

### LINKED ISSUE #423: Re-run CellAssign with raw.X on Ewing's samples

#292 

When originally running CellAssign for the Ewing's sample, we used the AnnData objects from the release bucket. These have logcounts stored in `X` and raw counts stored in `raw.X`. However, in the script for running `CellAssign` we use `X` when we should be using the raw counts in `raw.X`. This means we need to update the script to grab the raw counts, not log counts, and then re-run the analysis. 

Updating the `run-cellassign.py` to grab `raw.X`. 
https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/8a25222f6ed490bf6e8d4952ebc1c28aa5e7559d/analyses/ewings-cell-type/scripts/run-cellassign.py#L103

And then re-running the analysis and notebook and maybe adjusting conclusions/ findings in the notebook. 

_No response_

_No response_

_No response_

allyhawkins:
Closes #423 

Here I made the update to the `run-cellassign.py` script in the `ewings-cell-type` module to use the `raw.X` counts when running `CellAssign`. This is because we are using the processed `AnnData` objects from the Portal which have the logcounts stored in `X` and raw counts in `raw.X`. 

After doing that I regenerated all of the results from running `CellAssign` with the four different marker gene lists and then updated the notebook accordingly. This now has accurate and up to date results for running `CellAssign`. 

After updating the results, I re-ran the notebook and updated it based on the new results. 
This included: 

- Doing some more comparison of the marker gene classifications between manual annotation and CellAssign annotation using just the marker gene list. I added in a UMAP, confusion matrix, and exploration of marker gene expression. 
- I took some plots away from the filtered marker list section since using a smaller list actually means no cells are categorized as tumor, so there isn't much to plot there. 
- When looking at the annotations from using the Visser marker gene list, I added in a comparison to see if we are classifying the same cells as endothelial cells as we did with `CellAssign` in `scpca-nf`. 
- For the panglao reference I added in a UMAP showing the annotations and Jaccard heatmap comparing to the original `CellAssign` annotations in `scpca-nf`. 

_Results should be uploaded to your bucket so they are available during review._
_See here for instructions on how to upload to your bucket:_
https://openscpca.readthedocs.io/en/latest/software-platforms/aws/working-with-s3-buckets/

Updated predictions from running `CellAssign` can be found in: 
`s3://researcher-211125375652-us-east-2/ewings-cell-type/results/cellassign_predictions`. 

With these updated results, I actually think using just the marker genes with `CellAssign` does a pretty good job of pulling out tumor cells. At least in compared to the manual annotations (and the aneuploid detection, although I don't actually look at that here). 

I also found that decreasing that list of marker genes does worse and using the limited marker genes from Visser results in no annotation of tumor cells. Instead most cells are considered other. However, combining the Panglao reference list with marker genes does a good job of identifying endothelial cells and tumor cells. 

For just annotating tumor vs normal, I think using a set of marker genes could be promising. 
If we want to do normal too, then we may want to still try the approach of combining the full Panglao references with the tumor marker genes. This is definitely more computationally expensive though and I wonder how `CellAssign` would ultimately do trying to distinguish tumor cells vs. normal muscle cells, since those cells are currently classified as muscle. 

Here's an updated copy of the rendered notebook: 
[02-cellassign-tumor-cell-classification.html.zip](https://github.com/AlexsLemonade/OpenScPCA-analysis/files/15279788/02-cellassign-tumor-cell-classification.html.zip)

I was able to run everythning for this PR locally, but to avoid repeating `CellAssign`, grab the predictions results from the S3 bucket. 

I mentioned this in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/413#issuecomment-2105053809, but I'm not going to commit the rendered notebook. I'll remove the current rendered versions once we decide that is for sure a go. 

### Author checklists

_Check all those that apply._


.gitleaks.toml
```diff
@@ -0,0 +1,7 @@
+title = "gitleaks config for OpenScPCA"
+[extend]
+useDefault = true
+
+[allowlist]
+regexTarget = "line"
+regexes = ['''(?i)<img src="data:image\/.+;base64,.+?\s\/>''']
```
jashapiro, line 7: added a comment and slightly refined the regex
```suggestion
# skip base64 encoded images, which might have substrings that look like tokens
regexTarget = "line"
regexes = ['''(?i)<img src="data:image\/.+;base64,.+?".+\/>''']
```

jashapiro (approved):
LGTM!

I just suggested one slight change to your change to the script.

It is quite interesting (though maybe not surprising) that there are such strong effects of the other markers such that adding the Visser markers did so poorly. But that's the results!

PR #431, opened 2024-05-13
Add option to only download metadata

sjspielman:
I am proposing adding an argument to the data download script to only download metadata files. I found this useful while starting on #424, since I want to be able to select a couple libraries of varying sizes/diagnoses to run methods on. I added this option to quickly get the metadata so I can explore which specifies samples I want to download without waiting some hours for the full release to download. 

Let me know what you think of these changes! One other thought I had is that, since we're adding metadata downloads to the portal, maybe sometime soon-ish we might provide a single metadata file with the release here? But, this is what we've got for now.

Note - I recommend reviewing this in "split" view, since most of the diffs are just from tabbing which is not super clear in the "unified" view.

### DIFF
--- a/download-data.py
+++ b/download-data.py
@@ -72,7 +72,7 @@ def main() -> None:
             " One or more of 'processed', 'filtered', 'unfiltered', or 'bulk'."
             " Defaults to 'processed'."
             " For more than one level, use a comma separated list with no spaces."
-        ),
+        )
     )
     parser.add_argument(
         "--include-reports",
@@ -87,7 +87,7 @@ def main() -> None:
             "The project(s) to download."
             " A comma separated list of Project IDs to download."
             " Defaults to all. Can not be combined with `--samples`."
-        ),
+        )
     )
     parser.add_argument(
         "--samples",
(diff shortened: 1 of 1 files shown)


jashapiro:
I'm not going to review this now, because #426 Has a very large number of changes to the script (lots of refactoring), and I fear the conflicts with this will be end up quite substantial. After that goes in we can readdress this?

I also expect this to be largely mooted by the incoming portal changes, but convince me otherwise?

sjspielman:
> I also expect this to be largely mooted by the incoming portal changes, but convince me otherwise?

I have no compelling argument whatsoever to convince you otherwise on this one, aside from that I needed this info sooner than those portal changes will be incoming! Happy to just call this a "thing stephanie did for herself one day."

download-data.py
```diff
@@ -243,30 +249,36 @@ def main() -> None:
...
-        patterns += ["*.html"]
-
-    if "sce" in formats:
-        patterns += [f"*_{level}.rds" for level in includes]
+    # Only include other files if metadata_only is not false
+    # If metadata_only is true, overwrite other defaults for Download Summary printing
+    if args.metadata_only:
+        formats = ["N/A"]
+        includes = ["N/A"]
+    else:
```
jashapiro, line 257: For the future revision: I would rather have a guard that `--metadata_only` and other `--include` options are exclusive. That will allow you to avoid the `else:` and changing nesting levels.

I would also expect people might want to download metadata for only specific projects, so that part of the code would not be nested, though you would also want to guard against choosing samples, since w don't have individual metadata files.

sjspielman:
Closing since the script has been heavily refactored and strategy here needs to be changed.

Issue #427, opened 2024-05-10
Initiate a workflow to run scripts and notebooks in Ewings cell type module


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#292 

### Describe the goals of the changes to the analysis module.

Since starting to build up the `ewings-cell-type` module, we have now added a few scripts and notebooks and have enough to start a workflow of sorts. Although there are still missing steps and we are working on how things may link together, it would be helpful to at minimum have a shell script that runs all the scripts and notebooks in the required order. This will make it easier to run the same analysis on additional samples. 

### What will your pull request contain?

The addition of a shell script that will run the scripts and analysis notebooks in a specified order for classifying cells as tumor or normal.

allyhawkins:
Closed by #434

Issue #403, opened 2024-04-30
Use copy number inference to estimate tumor/normal cells in a Ewing sarcoma sample


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#292 

### Describe the goals of the changes to the analysis module.

Here we should test using different copy number inference methods that are available to annotate cells in Ewing sarcoma samples as tumor/normal. We should test out using: 

- [copyKat](https://github.com/navinlabcode/copykat)
- [InferCNV](https://github.com/broadinstitute/inferCNV/wiki)
- [SCEVAN](https://github.com/AntonioDeFalco/SCEVAN)

The PR should include running each method on a single sample and evaluating the results in comparison to the manual tumor/normal annotations obtained in #391. 

One thing to note is that `InferCNV` requires a group of normal cells as a reference and `SCEVAN` suggests using a vector of normal cells as a reference. For SCPCS000490, there are cells that are classified as normal types in SingleR and CellAssign, in particular, the endothelial cells. I think it's a safe bet that we can use these as a reference for normal cells. 

### What will your pull request contain?

This will contain a notebook that runs each method on a single `SingleCellExperiment` object. We will then plot the results from running each method and then use the Jaccard similarity index to compare cells that we manually annotated as tumor/normal to the annotations from each method. 

### Will you require additional software beyond what is already in the analysis module?

Yes, we will need to install each of these packages and then record them in the `renv.lock` file. 

### Will you require different computational resources beyond what the analysis module already uses?

I believe we should be able to run this locally.

allyhawkins:
I spent some time testing and playing around with using [`SCEVAN`](https://github.com/AntonioDeFalco/SCEVAN) to identify tumor cells and find potential CNVs as an alternative or supplement to using either `CopyKAT` or `InferCNV`. I wanted to summarize some of my thoughts here on using this package, since at this point I don't think it's worth spending more time on. Although in the future if we have difficulty with other samples with either `CopyKAT` or `InferCNV` (and if they fix the bug that I found) then I would consider using it. 

[...]

PR #432, opened 2024-05-13
Run InferCNV for Ewing's sample

allyhawkins:

Addresses #403 

This PR includes a script to run [`InferCNV`](https://github.com/broadinstitute/inferCNV/wiki) and an exploratory notebook looking at the results from running `InferCNV` on a Ewing's sample. The goal is to identify if using `InferCNV` is a reasonable approach for identifying tumor and normal cells in Ewing sarcoma. 

The first things I had to do was set up running `InferCNV`. To run this we need to create a [gene order file](https://github.com/broadinstitute/inferCNV/wiki/File-Definitions#gene-ordering-file) from the index we use for mapping in `scpca-nf`. They have a script on their repo that does it, but it was pretty easy to create directly from our gtf file, so I added a script here that copies over the gtf file in `s3://scpca-references` and then creates the gene order file from that gtf file. Note that these are both too big to commit to the repo, so as part of this I synced the `references` folder with my research bucket. So you can either grab the gene order file from that bucket or run the script to generate it. 

Then, I created a script to actually run `InferCNV`, since it can take a _long_ time to run. Using the settings in the script, it may take up to 2 hours to run. This script is very similar to the `run-copykat.R` script where the normal cells are provided as input and then used to run `InferCNV`. The main difference here is the creation of the annotations file, which right now I have as part of this script. `InferCNV` requires an annotation file where the normal or reference cells are annotated at minimum. We may want to move this to a separate script later, but for now I made it part of this script. Also, you can run the script without any normal cells, so if no normal cells are provided then InferCNV is run without them as a reference. 

I then ran `InferCNV` with the endothelial cells as a reference and with no cells as a reference. I evaluated the output in a notebook. One thing to note is that `InferCNV` mainly outputs the heatmap and then information on whether or not each cell has a gain or loss in a given chromosome. It does not call cells as tumor or normal, so we can use the output from `InferCNV` to help us identify which cells are more likely to be tumor, but no calls are provided by the program itself. This is different from `copyKAT` which actually predicts aneuploid or diploid cells. 

Also, it's unclear where the information for per cell CNVs actually lives and there is not one table that's output from `InferCNV`. Instead, you need to provide the output directory to all the results files to a function that will add that information to `meta.data` from a Seurat object. When running that function, a new file is created in the output directory that will contain the per cell information. I couldn't figure out how to generate this file or get this information without adding it to a Seurat object. 

In this notebook I show the following: 

- Heatmaps from running with and without references. From this it was pretty obvious to me that running without a reference may not be the way to go and if we have a known set of normal cells we should be using that. 
- Plots showing the distribution of total CNVs across cells. I then use this to set a cutoff and identify cells that are likely to be tumor cells. 
- Comparison of annotations to marker gene annotation and annotations with SingleR and CellAssign from scpca-nf. 
- Plots showing which cells have previously known CNVs likely to be found in Ewing's tumors. 

Just noting that I plan to tackle #427 next before continuing on any other analysis. I just had this mostly done before filing that issue so wanted to get it up into a PR. 

`s3://researcher-211125375652-us-east-2/ewings-cell-type`

Note that I included all results from running `InferCNV` in `results/infercnv`. 

The `references` folder contains any references present in this repo, including the gene order file. 

- Using `InferCNV` with a normal reference identifies a set of cells with CNVs that are not present in the reference normal cells. 
There is a group of cells that appear to have an increased number of CNVs present, and I labeled these as tumor cells.
- Because we do have to set our own cutoff to actually classify cells, I'm not sure how reproducible using InferCNV will be here. It may be beneficial to look at consistencies between two methods, like CopyKAT and InferCNV or manual annotation and InferCNV.  
- These cells are consistent with how we have been manually annotating tumor and normal cells with marker genes (and CellAssign and copyKAT, although I don't do any direct comparisons here). 
- `InferCNV` is helpful at visualizing specific CNVs that are likely to be present in Ewing's patients. If there are diseases with known alterations, this might be a good method since we can look for presence of that specific alteration in tumor cells. 

Here's a copy of the rendered report for easy review: 
[04-infercnv.html.zip](https://github.com/AlexsLemonade/OpenScPCA-analysis/files/15299062/04-infercnv.html.zip)

This can run locally and I recommend using 4 threads (that's the default for InferCNV). It does take some time, but it's not totally unreasonable to run locally. 
All packages have been added to the `renv.lock` to run these scripts and notebook. 

NA 

Just note that I had to increase the allowed max size for html files for this notebook to commit. 
I think it's probably because of the large heatmaps. 

I'm filing this as a draft since it contains changes also present in #413. I wanted to us the same identify normal cells script, so I will update this once that PR is in. 


allyhawkins:
> Also, it's unclear where the information for per cell CNVs actually lives and there is not one table that's output from InferCNV. Instead, you need to provide the output directory to all the results files to a function that will add that information to meta.data from a Seurat object. When running that function, a new file is created in the output directory that will contain the per cell information. I couldn't figure out how to generate this file or get this information without adding it to a Seurat object.

I updated this PR to grab the CNV cell level metadata within the `run-infercnv.R` script. I save that metadata to a TSV in the `results/infercnv` folder and keep all the other intermediate files that InferCNV outputs in the `scratch` directory. I also copy over the PNG with the final heatmap so the results will save only two files - the cell CNV metadata and the png.

analyses/cell-type-ewings/04-infercnv.Rmd
```diff
@@ -0,0 +1,433 @@
...
+
+Another option is to use the scaled proportion of CNVs which tells us the proportion of genes on that chromosome that are part of the CNV.  
+
+```{r}
+proportion_cols <- names(cnv_df)[startsWith(colnames(cnv_df), "proportion_scaled_cnv")]
+
+# get mean proportion of cnvs per cell and then scale that number 
+cnv_df <- cnv_df |> 
+  dplyr::mutate(mean_proportion = rowMeans(across(proportion_cols)),
+                scaled_mean_proportion = scale(mean_proportion))
```
jashapiro, line 226: Since proportions are already bounded 0-1, I probably would not use scaling here. It probably doesn't quite apply to counting chromosomes either, now that I think about it.

analyses/cell-type-ewings/04-infercnv.Rmd
```diff
@@ -0,0 +1,433 @@
...
+It looks like maybe we could again use 0 as the cutoff here and anything above 0 is a tumor cell. 
+Let's see how this matches up with the tumor/normal classifications we defined just using total CNVs. 
+
+```{r}
+ggplot(cnv_df, aes(x = scaled_mean_proportion, color = cnv_classification)) +
+  geom_density() +
+  theme_bw()
+```
+
+Interestingly, the cells we labeled as tumor have a higher mean proportion value then those that we labeled as normal.
```
jashapiro, line 255: Isn't this what we would expect?

analyses/cell-type-ewings/04-infercnv.Rmd
```diff
@@ -0,0 +1,433 @@
...
+
+ggplot(cnv_df, aes(x= UMAP1, y = UMAP2, color = proportion_scaled_dupli_chr1)) +
+  geom_point(alpha = 0.5, size = 0.5) +
+  labs(title = "Chr1 gain") +
+  theme_bw() +
+  scale_color_viridis_c()
+```
+
+Now when we look at the proportion, we see that only the group of tumor cells on the bottom seem to have a higher proportion of genes in chr8 that have a gain in expression. 
+Based on the papers, we should see a gain in the whole chr8 and chr12, not just one arm or specific genes, so we expect a high value for the proportion.
```
jashapiro, line 414: This makes sense to me, as it seems like the method has a high chance of calling small CNVs somewhere in the chromosome. This is why I would expect that the proportion is going to be a better measure overall. There may also be some value in looking at the _variance_ in the proportions across chromosomes. We might expect that tumor cells have greater variance based on the fact that they are likely to have some larger duplications/deletions than normal cells.

jashapiro (commented):
Sorry for the delay in getting this to you! Overall, I like the approach you took with dividing up the scripts, so I don't have much direct comment there, other than that I would probably include adding them to the workflow script as part of this PR. I think there are also some other little updates that probably need to be made to line things up with PRs that have gone in since this was submitted as well. 

I started with some comments on the scripts and such, which are mostly minor. The biggest thing is that I think you _may_ be able to get rid of the Seurat conversion by just using the `add_to_seurat()` function with no Seurat object set. It's a little unclear whether you get all the same data when you do that though, so maybe not. I didn't set up to do a full test...

Looking at the overall results, my inclination is to focus on the `proportion` statistics, as the method seems to be biased toward calling _some_ CNV on a chromosome more often than you might expect in normal cells, but those will mostly be small. So I might expect to get a bigger separation with that measure, averaged across chromosomes. For the averaging, you might want to scale by chromosome size though? Or if there are any stats where that (or other whole-genome measure) is already calculated?

For any of these (and this probably goes for copyKAT as well) you should probably exclude cells that went into the method as "normal" from any statistics that you are calculating and displaying. Doing so will create a bias, as those are the training set, and we would have a strong prior expectation about the CNV content that the method would calculate for those. So I would be sure to set those values as NA for any plots and before doing any scaling (which is probably not needed here, given the fact that proportions are already bounded 0-1)

allyhawkins:
> Sorry for the delay in getting this to you! Overall, I like the approach you took with dividing up the scripts, so I don't have much direct comment there, other than that I would probably include adding them to the workflow script as part of this PR. I think there are also some other little updates that probably need to be made to line things up with PRs that have gone in since this was submitted as well.


I went ahead and made some updates based on changes that were made since this was filed. This included updating the script to handle the reference cells as a dataframe instead of a list and adding the script and notebook to the workflow. Just a note that the template notebooks are _slightly_ different then these exploratory notebooks since I add a few params and remove any concluding text. It does make this PR kind of big, so I can separate out those commits into a different PR if you would prefer that. 

> I started with some comments on the scripts and such, which are mostly minor. The biggest thing is that I think you _may_ be able to get rid of the Seurat conversion by just using the `add_to_seurat()` function with no Seurat object set. It's a little unclear whether you get all the same data when you do that though, so maybe not. I didn't set up to do a full test...

I did test this and it does work! It just creates the table and saves as a `.txt` file which I copy over to the final output directory. 

> Looking at the overall results, my inclination is to focus on the `proportion` statistics, as the method seems to be biased toward calling _some_ CNV on a chromosome more often than you might expect in normal cells, but those will mostly be small. So I might expect to get a bigger separation with that measure, averaged across chromosomes. For the averaging, you might want to scale by chromosome size though? Or if there are any stats where that (or other whole-genome measure) is already calculated?

I agree that I think the proportion statistics are going to be more informative here. I also updated the averaging of the scaled proportion across all chromosomes to weight by the number of genes in each chromosome. To do this I read in the full InferCNV results which has a table of all genes used and their chromosome. I ended up identifying a cutoff based on looking at the bimodal distribution, but I don't know how well this cut off will apply to additional samples. With the value being bounded between 0-1 I think it will be okay. 

I also noted your idea of using variance in CNVs as a way to identify tumor cells. I didn't do anything to address that here, but I think if we have trouble identifying cells in future samples then we may want to revisit that. 


> For any of these (and this probably goes for copyKAT as well) you should probably exclude cells that went into the method as "normal" from any statistics that you are calculating and displaying. Doing so will create a bias, as those are the training set, and we would have a strong prior expectation about the CNV content that the method would calculate for those. So I would be sure to set those values as NA for any plots and before doing any scaling (which is probably not needed here, given the fact that proportions are already bounded 0-1)

I updated this to exclude any "normal" cells before doing any scaling for the total sum of CNVs and then set to NA for most of the plots. 

There was also a comment about the colors on the heatmap and I'm honestly not sure why both reference and unknown were getting the same color. I did add some more comments in the notebook to explain interpreting the heatmap, including how the reference cells are used. 

Updated rendered notebook for review: 
[04-infercnv.html.zip](https://github.com/AlexsLemonade/OpenScPCA-analysis/files/15381801/04-infercnv.html.zip)

analyses/cell-type-ewings/README.md
```diff
@@ -11,6 +11,10 @@ Currently, we use a variety of methods to annotate tumor cells independently:
 The full list of marker genes can be found in `references/tumor-marker-genes.tsv`.
 - Use the list of tumor marker genes to classify tumor cells with [`CellAssign`](https://docs.scvi-tools.org/en/stable/user_guide/models/cellassign.html).
 - Identify copy number variations and annotate tumor cells using [`CopyKAT`](https://github.com/navinlabcode/copykat).
+- Identify copy number variations using [`InferCNV`](https://github.com/broadinstitute/inferCNV/wiki).
+This returns a proportion of each chromosome with a CNV detected.
+We then calculate the mean proportion for each cell across all chromsomes and weight by the number of genes in a chromosome.
+Cells with a mean proportion > 0.1 are called as tumor cells.
```
jashapiro, line 17: I think this is fine for now, but I do wonder if there may be a way to find a split in a bimodal distribution, should one exist. (I have not yet gotten to the data... just thinking ahead)

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,485 @@
...
+  with_ref = file.path(results_dir, "endothelial_reference")
+)
+
+# png ref files 
+png_files <- all_results_dir_list |>
+  purrr::map_chr(\(dir) file.path(dir, glue::glue("{params$library_id}_infercnv.png")))
+
+png_files |> 
+  # make sure png files exists
+  purrr::map(\(file)
```
jashapiro, line 80: Prevent extra output
```suggestion
  purrr::walk(\(file)
```

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,485 @@
...
+
+From the [`InferCNV` documentation](https://github.com/broadinstitute/inferCNV/wiki/Interpreting-the-figure): 
+
+> The normal cells in the top heatmap define baseline expression for genes in normal cells. This baseline distribution of normal gene expression is subtracted from both the normal cells from which was defined as well as the tumor cells. Afterwards, the normal cell expression heatmap should be largely devoid of signal, with the exception of certain outlier gene expression values in certain cells. Removing this baseline of normal expression signal from the tumor cells should reveal those chromosomal regions that have significantly more or less expression than the normal cells, highlighting likely amplified or deleted whole chromosomes or large chromosomal regions.
+
+The dendogram on the left side is hierarchical clustering of all non-reference cells. 
+The color annotation next to it indicates subdivisions of the dendogram by cutting it in `k_obs_groups` groups, with the default `k_obs_groups` = 1.
```
jashapiro, line 122: Which color annotation is this? There are two, it seems to me, and I can't tell what the rainbow is: it certainly isn't `k_obs_groups = 1`!

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,485 @@
...
+reference_cells_only <- all_cnv_df |> 
+  dplyr::filter(stringr::str_detect(subcluster, "reference")) |> 
+  dplyr::mutate(has_cnv = NA,
+                z_scaled_cnv = NA)
+
+unknown_cells_only <- all_cnv_df |>
+  dplyr::filter(stringr::str_detect(subcluster, "unknown")) |> 
+  # get total number of chromosomes with cnvs per cell and then scale that number 
+  dplyr::mutate(has_cnv = rowSums(across(has_cnv_cols)),
+                z_scaled_cnv = scale(has_cnv))
```
jashapiro, line 208: Rather than splitting these up and then rejoining, I would probably do something like this:
```suggestion
# don't include the normal cells that were provided as reference in scaling 
cnv_df <- all_cnv_df |>
  # get total number of chromosomes with cnvs per cell and then scale that number 
  dplyr::mutate(
    has_cnv = ifelse(
      stringr::str_detect(subcluster, "reference"), 
      NA_real_, 
      rowSums(across(has_cnv_cols))
    ),
    z_scaled_cnv = scale(has_cnv))
```

`scale` unlike `mean` and `sum` seems to handle `NA` values as you would want it to.

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,485 @@
...
+
+```{r}
+proportion_cols <- names(cnv_df)[startsWith(colnames(cnv_df), "proportion_scaled_cnv")]
+
+# get total number of genes per chromosome to use in weighted mean 
+chr_weights <- final_cnv_obj@gene_order |> 
+  as.data.frame() |> 
+  dplyr::count(chr) |> 
+  # only keep chr 1-22, drops MT and GL chr
+  dplyr::filter(chr %in% glue::glue("chr{seq(1,22)}")) |>
```
jashapiro, line 264: Do you also want X?

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,485 @@
...
+  dplyr::count(chr) |> 
+  # only keep chr 1-22, drops MT and GL chr
+  dplyr::filter(chr %in% glue::glue("chr{seq(1,22)}")) |> 
+  dplyr::pull(n)
+
+# get mean proportion of cnvs per cell weighted by chromsome size (number of genes)
+cnv_df <- cnv_df |> 
+  dplyr::rowwise() |> 
+  # calculate weighted mean, excluding normal cells 
+  dplyr::mutate(mean_proportion = ifelse(!is.na(has_cnv), weighted.mean(across(proportion_cols), num_genes_df$n), NA)) |>
```
jashapiro, line 271: You did it in one step this time, but I would break up across lines for clarity

jashapiro: You are using `num_genes_df` here, but I think this should be `chr_weights` after the code? I also think this is probably not quite right, as the MT genes have not been excluded from the proportion columns.

Rather than filtering out those chromosomes above, I might just set their weights to 0 so you don't have to worry about mismatched lengths?

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,485 @@
...
+  theme_bw() +
+  scale_color_viridis_c()
+```
+
+```{r}
+ggplot(cnv_df, aes(x = mean_proportion)) +
+  geom_density() +
+  theme_bw()
+```
+It looks like we could use a cutoff of 0.1 here to classify tumor/normal cells.
```
jashapiro, line 287: For formatting
```suggestion

It looks like we could use a cutoff of 0.1 here to classify tumor/normal cells. 
```

analyses/cell-type-ewings/scripts/run-infercnv.R
```diff
@@ -0,0 +1,160 @@
...
+                              HMM=T ,
+                              save_rds = FALSE, # don't save the intermediate rds files
+                              num_threads = opt$threads
+)
+
+# Save final results -----------------------------------------------------------
+
+# create table with barcodes and CNVs for each chromosome
+infercnv::add_to_seurat(seurat_obj = NULL,
+                                      infercnv_output_path = scratch_dir)
```
jashapiro, line 147: Just spacing
```suggestion
infercnv::add_to_seurat(seurat_obj = NULL,
                        infercnv_output_path = scratch_dir)
```

analyses/cell-type-ewings/scripts/run-infercnv.R
```diff
@@ -0,0 +1,160 @@
...
+
+# define path to output cnv metadata file saved in scratch dir
+scratch_metadata_file <- file.path(scratch_dir, "map_metadata_from_infercnv.txt")
```
jashapiro, line 150: define this path at the top?

I also wonder if we want to keep these named by library, as with all the other files.

analyses/cell-type-ewings/template_notebooks/04-infercnv.Rmd
```diff
@@ -0,0 +1,455 @@
...
+  geom_point(alpha = 0.5, size = 0.5) +
+  labs(title = "Chr1 gain") +
+  theme_bw()
+
+ggplot(cnv_df, aes(x= UMAP1, y = UMAP2, color = has_loss_chr16)) +
+  geom_point(alpha = 0.5, size = 0.5) +
+  labs(title = "Chr16 loss") +
+  theme_bw()
+```
+
```
jashapiro, line 409: As these plots are kind of big, I think I might either use `patchwork` to join them into a grid or maybe even just remove the `has_dupli` plots from the template altogether, and only show the proportion plots, which I think are more informative overall. (Still might use patchwork for the proportion plot... or pivot and facet)


Also, as with copykat, it is probaby worth showing this data in the context of the chromosomes that are not expected to have CNVs: A distribution for each chromosome, faceted, is probably worth doing (which would make making the UMAP plots faceted by chromosome easier as well, so you don't need patchwork)

jashapiro (commented):
Looking good. I think it _is_ a bit much to have both the exploratory and templates in the same PR, but it is fine for now.

I do think you may want the template to be a bit different: some of the plots in the exploratory notebook are probably overkill, and I think you could reduce the amount to look through in the template. That said, I do want you to add one plot, which is the distributions of cnv proportions for all chromosomes, not just the ones we expect to see in Ewings. Having negative controls is good!

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
allyhawkins: I'm having trouble understanding the docs here, because it sounds like the bar immediately next to the dendogram (the rainbow one) is the subdivisions of the dendogram. I think there is an error in the docs, because you are correct that is not `k_obs_groups=1`. And the right most bar is supposed to be the original annotation provided.

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
allyhawkins: ChrX isn't in the output as an option.

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
allyhawkins: Sorry the `num_genes_df` is leftover from when that's what it was named. Also I have removed the MT genes from all columns so the dimensions should match up correctly.

analyses/cell-type-ewings/scripts/run-infercnv.R
allyhawkins: We can't name this file. This is just defining the path to the file that's saved by InferCNV and then when we copy it over we add the library ID.

jashapiro: Yeah, sorry, I was confused!

allyhawkins:
> I do think you may want the template to be a bit different: some of the plots in the exploratory notebook are probably overkill, and I think you could reduce the amount to look through in the template. That said, I do want you to add one plot, which is the distributions of cnv proportions for all chromosomes, not just the ones we expect to see in Ewings. Having negative controls is good!

So I ended up removing the UMAPs that weren't showing the proportion and were showing just the `has_cnv` all together. I don't think they gave us that much information and if we are using the proportion we should use that. I updated this section in both exploratory and template notebooks to include a faceted density plot where we look at the proportion of cnvs for each chromosome and color by whether or not we classified the cell as tumor or normal. I think this is helpful in showing which chromosomes are affected. 

I then added a faceted UMAP for just the chromosomes commonly affected in Ewings. To me making 21 UMAPs seemed like overkill and I didn't get a lot from looking at all of them on a UMAP. Let me know if you feel differently. 
[04-infercnv.html.zip](https://github.com/AlexsLemonade/OpenScPCA-analysis/files/15395582/04-infercnv.html.zip)

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,480 @@
...
+  )
+
+```
+
+```{r, fig.height=10, fig.width=10}
+ggplot(proportion_df, aes(x = proportion, color = cnv_mean_classification)) +
+  geom_density() +
+  theme_bw() +
+  facet_wrap(vars(chromosome), ncol = 3)
+```
```
jashapiro, line 440: A couple adjustments to the density plot here to make it more readable: 

One is to add bounds to the  density calculation, which should help make the plot not extend past (0,1), but you may still want to set the width of the plot as well with `coord_cartesian()`. I also set a bandwidth (which was just a guess, definitely see if you can adjust) to make the plots more comparable. The goal here is to make the plots less spikey than the default, with a secondary goal of making them more similar in the y-range. 

The other thing you could do is add scaling to the density plots so they all have a max of 1: `geom_density(aes(y = after_stat(scaled))`, which is not quite kosher, but does often make comparing shapes easier.

````suggestion
```{r, fig.height=10, fig.width=10}
ggplot(proportion_df, aes(x = proportion, color = cnv_mean_classification)) +
  geom_density(bounds = c(0,1), bw = 0.05) +
  theme_bw() +
  facet_wrap(vars(chromosome), ncol = 3)
```
````

File: analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd [hunk 15]
```diff
@@ -0,0 +1,480 @@
...
+# don't include the normal cells that were provided as reference in scaling 
+cnv_df <- all_cnv_df |>
+  # get total number of chromosomes with cnvs per cell and then scale that number 
+  dplyr::mutate(
+    has_cnv = ifelse(
+      stringr::str_detect(subcluster, "reference"), 
+      NA_real_, 
+      rowSums(across(has_cnv_cols))
+    ),
+    z_scaled_cnv = scale(has_cnv))
```
jashapiro, line 207: I was not super comfortable with this before, and on a bit more reflection, I think we want to just remove it. The reason is that the values of `has_cnv` are not continuous, but rather in the range [0-22], and that is not going to change. So I think we probably want to just use that knowledge straight. So below, I would say you should use histograms rather than density plots.

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,480 @@
...
+  scale_color_viridis_c()
+```
+This shows that the total CNVs are highest in the group of cells that correspond to the same cells that are manually annotated as tumor cells, which is promising. 
+Those cells in grey (bottom left) are cells used as a reference. 
+If we were to use total CNVs we would need some sort of cut off for the number. 
+Let's look at the distribution of total CNVs and also the scaled total CNVs to see if we can identify a cut off. 
+
+```{r}
+ggplot(cnv_df, aes(x = has_cnv)) +
+  geom_density() +
```
jashapiro, line 225: Try a histogram here instead? (binwidth and center to make things line up over the integers)
```suggestion
  geom_histogram(bindwith = 1, center = 0, alpha = 0.5, position="identity") +
```

alternatively: 
```suggestion
  geom_freqpoly(binwidth = 1, center = 0) +
```

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,480 @@
...
+  theme_bw() +
+  labs(x = "Total number of CNVs")
+
+ggplot(cnv_df, aes(x= z_scaled_cnv)) + 
+  geom_density() +
+  theme_bw() +
+  labs(x = "z-scaled chromosome CNV count")
+```
+
+Looking at the scaled number of CNVs it does look like we could use 0 as the cutoff and call any cell that has > 0 CNVs (after scaling) as tumor and all other cells as normal.
```
jashapiro, line 235: This is the equivalent of just using the mean as the cutoff, so let's make that explicit?

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,480 @@
...
+
+Looking at the scaled number of CNVs it does look like we could use 0 as the cutoff and call any cell that has > 0 CNVs (after scaling) as tumor and all other cells as normal. 
+
+In the below plot we label all non-reference cells as either tumor or normal. 
+Cells that are NA are those that were already classified as reference cells. 
+
+```{r}
+# add cnv classification 
+cnv_df <- cnv_df |> 
+  dplyr::mutate(cnv_classification = dplyr::if_else(z_scaled_cnv > 0, "Tumor", "Normal"))
```
jashapiro, line 243: Explicitly using the mean.
```suggestion
  dplyr::mutate(cnv_classification = dplyr::if_else(has_cnv > mean(has_cnv), "Tumor", "Normal"))
```

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
```diff
@@ -0,0 +1,480 @@
...
+  scale_color_viridis_c()
+```
+
+```{r}
+ggplot(cnv_df, aes(x = mean_proportion)) +
+  geom_density() +
+  theme_bw()
+```
+
+It looks like we could use a cutoff of 0.1 here to classify tumor/normal cells.
```
jashapiro, line 288: I might also look at using the mean here? This will be a more productive discussion with more samples, of course.

analyses/cell-type-ewings/scripts/make-gene-order-file.R
```diff
@@ -0,0 +1,53 @@
...
+)
+
+# Parse options
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# create ref directory if doesn't already exist 
+fs::dir_create(opt$local_ref_dir)
+
+# sync gtf file to local directory 
+gtf_filename <- stringr::word(opt$gtf_file, -1, sep = "/")
```
jashapiro, line 30: simpler and probably more robust
```suggestion
gtf_filename <- basename(opt$gtf_file)
```

analyses/cell-type-ewings/README.md
jashapiro: Added a thought below that maybe just the mean is the way to do this

jashapiro (commented):
Looks pretty good, though I had a few more thoughts on plots and scaling. I'd approve, but I kind of want to make sure I see something of what the results of my suggestions look like!

analyses/cell-type-ewings/README.md
allyhawkins: So just the mean of the scaled proportion values across all chromosomes rather than weighting by chromosome? Or just taking the mean of the sum of total chromosomes with CNVs?

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
allyhawkins: We are taking the weighted mean here so want to check that you just want to use the mean alone? Or compare calling tumor cells using the weighted mean to the mean of the CNV sum?

analyses/cell-type-ewings/README.md
jashapiro: Two different answers, which are really the same:
- For using the total chromosomes with CNVs, just the mean # of chromosomes with CNVs as the dividing line. (This is the same as scaling and then using 0).
- For the weighted mean proportion CNV values, again using the mean across cells as the dividing line. The idea here is that if there is a bimodal distribution, that should generally lie _somewhere_ between the two peaks. There are going to be more sophisticated ways of choosing a dividing line, but I expect this to be a reasonable place to start.

analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd
jashapiro: I think yes. Specifically, the cutoff I am proposing is `mean(cnv_df$mean_proportion)`

allyhawkins: Sorry I was confused and thought you meant to look at the distribution of the mean and wasn't clear that you meant to use it as the cutoff. But that makes sense so I'm updating it to use that!

allyhawkins:
[04-infercnv.html.zip](https://github.com/AlexsLemonade/OpenScPCA-analysis/files/15404977/04-infercnv.html.zip)

@jashapiro I updated this to use the mean(sum_cnv) and the mean(proportion_cnv) as cutoffs rather than setting the arbitrary cutoffs. The results don't change, but I agree this might be more robust to calling cells in future samples. 

I also removed any scaling all together and any plots that showed the scaled data. I added in the histogram that you recommended and I adjusted the density plots. Thanks for that suggestion! They are much easier to interpret. 

This should be ready for another look.

PR #434, opened 2024-05-15
Initiate workflow for annotating tumor cells in Ewing's samples

allyhawkins:

Addresses #427 

The goal of this PR is to begin a workflow that can be used to obtain annotations of tumor and normal cells across multiple Ewing's samples. We started with a group of exploratory notebooks and scripts that were run to explore use of a variety of methods for tumor cell type annotation, but we want to be able to run this analysis on additional samples and generate the same notebooks for a few samples. The purpose of this workflow is mainly to be able to generate the same rendered notebooks but on multiple samples. This includes using manual annotation, `CellAssign`, `CopyKAT` and `InferCNV`. 

I imagine that this will be revised to get a "final" workflow that identifies tumor cells in a sample using the methods we choose to continue with. We also may revise this further to include other annotations besides tumor cells, but right now it's just running the same methods we have already implemented in the module. 

In this PR, I focused on getting the workflow set up, defining inputs and outputs, and creating a step to run the manual classifications. I anticipate having separate PRs for the addition of the next steps (CellAssign, CopyKAT, and InferCNV once #432 is done). 

**Workflow setup:**

- I moved all of the current notebooks and rendered notebooks to an `exploratory_analysis` folder. This keeps the copies of the original notebooks with all the conclusions made within those notebooks in tact. 
- I created a copy of all those notebooks and added to a `template_notebooks` folder. These are the same notebooks _mostly_ with some modifications. I have made these modifications to only the marker gene notebook since I was trying to keep changes minimal. I will change the other notebooks as we add each step. 
  - I took out any sample/library specific conclusions from the text. I kept text that was descriptive about methods, but anything that concluded something from the plot that wasn't general was removed since it won't be applicable to all future samples we run. 
  - I added additional params to the notebook for the `results_dir` and then also a reference cell table. In the marker gene notebook we pull out specific annotated cell types from `SingleR` and `CellAssign` and treat those as either "normal" or "tumor". However, the annotations are not always consistent across samples... so I added an option to read in a table that has normal cells and tumor cells (more on creating that table later). 

**The workflow:**

I created an initial workflow that takes as input the sample ID, library ID, a list of normal cell types and a list of tumor cell types. 

  - The first step of the workflow is to create a table with cells to use as references for later steps in the workflow. Here I modified the previous `select-normal-cells.R` to create a reference table rather than output just a list of normal cells. You can now use the `--normal_cells` and `--tumor_cells` arguments to provide a list of cell types that should be "normal" or "tumor" for reference. These annotations can come from either `SingleR` or `CellAssign`. The output is a data frame with the barcode, the reference classification (tumor or normal), and associated annotation for that cell from `SingleR` and `CellAssign`. This does mean we need some minor changes to how the references get read in to `CopyKAT`, but I can easily address that when we add that script to the workflow. 
  - The second step is rendering the marker gene notebook. This notebook also produces a final table with barcodes and marker gene classification which get used as input into later steps. I think ultimately we might want to create a separate script that actually produces just this file, but for now I think it's fine to leave in the notebook. 
  - The outputs to the workflow get saved to `results/annotate_tumor_cells_output/sample_id` and include a marker gene report and marker gene classifications for each library. I think I want the final output to just include each of the reports and one final table that contains tumor/normal classifications from each method in one file. But that will get modified at a later time. 

Yes, I plan on filing additional PRs that add running `CellAssign` and `CopyKAT` to this workflow. 

_Results should be uploaded to your bucket so they are available during review._
_See here for instructions on how to upload to your bucket:_
https://openscpca.readthedocs.io/en/latest/software-platforms/aws/working-with-s3-buckets/

I have only run the workflow on one sample thus far, generating a report and table of marker gene classifications. 

No new results that aren't already present in the `exploratory_analysis` folder have been produced yet. 

Just noting that I ran this locally without any issue and no special requirements. 

Just noting that the only change from the original notebooks will be in `template_notebooks/01-marker-genes.Rmd`. The other notebooks that got added to that folder were copied exactly from the original and will be modified in later PRs. 
Specifically, the only code changes I introduced were updats to the parameters and lines 104-109 and 394-432. I also deleted some concluding text. 

I know there was talk about just making a bash script to run the analysis as it was, but there were some changes that needed to be made to be able to run things on other samples, which is what I would like to do. So that's why I started this way, in an attempt to make things more flexible. I still tried to keep things pretty simple for now since we only plan on running this on a few samples, so I don't want to spend too much time perfecting the "workflow" aspect. I do anticipate as we finalize methodology that this will change. 

### Author checklists

_Check all those that apply._


analyses/cell-type-ewings/annotate-tumor-cells-workflow.sh
```diff
@@ -0,0 +1,104 @@
...
+A set of reports summarizing the tumor cell annotations and a TSV file containing annotations from all used methods will be saved to the `results_dir`.
+The default `results_dir` is `results/annotate_tumor_cells_output/<sample_id>`.
+
+Example of running the workflow:
+
+./annotate-tumor-cells-workflow.sh \
+  --sample_id "SCPCS000490" \
+  --library_id "SCPCL000822" \
+  --normal_celltypes "Endothelial cells,endothelial cell" \
+  --tumor_celltypes "Pulmonary vascular smooth muscle cells,smooth muscle cell"
```
jashapiro, line 34: I haven't done a full review, obviously, but I think we want to avoid having arguments like this to a shell script. I just don't think it is very maintainable and can easily end up with hidden bugs with misspelled option names and whatnot. The two approaches I might take would be to either convert this to a python/R script where you can use a full argument parser (or read in a sample sheet that is part of the repo), or use arrays of samples/libraries coded into the script itself and looping through those. 

The main goal is to allow a single command without any options to be used for testing, etc. That might call a script that does a sample a time, but we want a "full" test where we can.

(I might also try to avoid specifying both sample & library, if you can avoid it: if we can discover the library ids by what is present in the sample dir, that would be easier to keep track of, assuming you don't need separate different cell types for each of the libraries within a sample)

allyhawkins: > The two approaches I might take would be to either convert this to a python/R script where you can use a full argument parser (or read in a sample sheet that is part of the repo), or use arrays of samples/libraries coded into the script itself and looping through those.

I was trying to keep things simple in this first PR and just have a script that will run every step on a single sample so had avoided making a sample sheet. I was planning on modifying this workflow to be more formal later to be able to run on multiple samples. One option is to have a script that wraps around this and parses in the arguments, the other is to move to using snakemake and provide a sample sheet as input. I wasn't sure that either of those things were necessary yet, but I can make that change here if you prefer that.

jashapiro: > I was trying to keep things simple in this first PR and just have a script that will run every step on a single sample so had avoided making a sample sheet. I was planning on modifying this workflow to be more formal later to be able to run on multiple samples. One option is to have a script that wraps around this and parses in the arguments, the other is to move to using snakemake and provide a sample sheet as input. I wasn't sure that either of those things were necessary yet, but I can make that change here if you prefer that.

I appreciate that. It's really just having the argument parsing code in bash at all in here that I am probably reacting to. Bash just isn't good at it! If you had the script use only internal variables, that would probably be fine for now. But I would try to make it work at the sample level, if that is possible.

allyhawkins: Okay for now I did end up adding in a sample sheet and then reading that in as the main argument of the workflow. I think we are going to ultimately want that to help track what cell types are used for which sample as normal cells, so mine as well create it now.

analyses/cell-type-ewings/scripts/select-cell-types.R
```diff
@@ -0,0 +1,107 @@
...
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# Set up -----------------------------------------------------------------------
+
+# make sure path to sce file exists
+stopifnot("sce_file does not exist" = file.exists(opt$sce_file))
+
+# check that either normal or tumor cell types are provided 
+stopifnot("Either normal_cells or tumor_cells must be provided" = 
+            !is.null(opt$normal_cells) || !is.null(opt$tumor_cells))
```
jashapiro, line 52: All checks can be done in a single call to `stopifnot()`
```suggestion
stopifnot(
  # make sure path to sce file exists
  "sce_file does not exist" = file.exists(opt$sce_file)),

  # check that either normal or tumor cell types are provided 
  "Either normal_cells or tumor_cells must be provided" = 
    !is.null(opt$normal_cells) || !is.null(opt$tumor_cells)
)       
```

analyses/cell-type-ewings/scripts/select-cell-types.R
```diff
@@ -0,0 +1,107 @@
...
+  as.data.frame() |> 
+  dplyr::select(barcodes, all_of(annotation_columns)) |> 
+  tidyr::pivot_longer(all_of(annotation_columns),
+                      names_to = "celltype_method",
+                      values_to = "celltype")
+
+# get list of normal and tumor cell types to save as refernences 
+if(!is.null(opt$normal_cells)){
+  normal_cell_types <- stringr::str_split(opt$normal_cells, pattern = ",") |>
+    unlist()
```
jashapiro, line 80:
```suggestion
  normal_cell_types <- stringr::str_split_1(opt$normal_cells, pattern = ",")
```

analyses/cell-type-ewings/scripts/select-cell-types.R
```diff
@@ -0,0 +1,107 @@
...
+} else {
+  normal_cell_types <- ""
+}
+
```
jashapiro, line 84: Since you are going to set these to "", you could remove a lot of logic by just making that the default value when reading in. You would only need:
```suggestion
normal_cell_types <- stringr::str_split_1(opt$normal_cells, pattern = ",") 
```

Of course, you would also have to update the check to check whether the args `== ""` instead of `is.null`, but that should be pretty easy too:

```
"Either normal_cells or tumor_cells must be provided" = 
    opt$normal_cells != "" || opt$tumor_cells != ""
```

analyses/cell-type-ewings/scripts/select-cell-types.R
```diff
@@ -0,0 +1,107 @@
...
+# label cells as either tumor or normal based on which cell type was annotated 
+ref_table_df <- coldata_df |> 
+  dplyr::mutate(reference_cell_class = dplyr::case_when(
+    celltype %in% normal_cell_types ~ "Normal",
+    celltype %in% tumor_cell_types ~ "Tumor"
+  )) |> 
+  # remove any that weren't tumor or normal 
+  dplyr::filter(!is.na(reference_cell_class)) |>
```
jashapiro, line 99: A small modification to make the default case explicit. 
```suggestion
    celltype %in% tumor_cell_types ~ "Tumor",
    .default = "NA"
  )) |> 
  # remove any that weren't tumor or normal 
  dplyr::filter(reference_cell_class != "NA") |> 
```

analyses/cell-type-ewings/scripts/select-cell-types.R
```diff
@@ -0,0 +1,107 @@
...
+option_list <- list(
+  make_option(
+    opt_str = c("--sce_file"),
+    type = "character",
+    help = "Path to RDS file containing a processed SingleCellExperiment object from scpca-nf"
+  ),
+  make_option(
+    opt_str = c("--normal_cells"),
+    type = "character",
+    default = NULL,
```
jashapiro, line 26: See other comments for why I think this can simplify things a lot!
```suggestion
    default = "",
```

jashapiro (commented):
The general logic and movements of code here make sense to me. 

I had some small suggestions for the `select-cell-types.R` script, many of which come down to a realization that I have had in the last few months that dealing with `NULL` values as defaults for arguments often makes things more complicated later, and the list of cells types here is a prime example: simply setting the default as `""` removes a whole bunch of logic, with exactly the same results.

But my main comment at this stage remains that I think you are trying to do too much with bash! 😄 I don't mean to be a noodge about it, but it just seems like trying to stitch too much together there is just going to lead to trouble down the road... If you find yourself needing  more than one or two arguments for a bash script, I would immediately switch to something which can handle those more robustly, with help text, and if you are trying to parse a file, there are lots of potential gremlins! And if you plan to switch to something else later, don't worry about implementing something you are going to have to rip out later anyway!

So my recommendation is to either keep it very simple, or take the plunge to something more robust. 

The one suggestion I would make if you do keep it in bash though is to use something like: 
```
for sce in $data_dir/$sample_id/*_processed.rds
do
  Rscript $scripts_dir/select-cell-types.R --sce_file "$sce" 
  ...
done
```
so that you do not have to keep track of both libraries and samples...

If you really need the library id, you can get it with something like:
```
library_id=$(basename $sce | sed 's/_processed.rds$//')
```

All that said: if your intent is to move to something like snakemake or nextflow to handle the processing, I would leave this script as simple as possible: Just set the sample id and anything else that might need modification later as a bash variables at the top of the script. If you want, you can set them with the `$VAR=${VAR:-default}` syntax and then somebody calling can always run the script as `VAR=newval annotate.sh` to modify on the command line. I feel that that format makes it more clear that the script is not meant to be a fully robust solution.

To be clear, what I think we really want at this stage is just an easy, runnable record of the steps you are taking and how they fit together. We can add layers on top of it later, but the simpler it is now, the easier that will be.

analyses/cell-type-ewings/annotate-tumor-cells-workflow.sh
```diff
@@ -0,0 +1,103 @@
...
+cd $(dirname "$0")
+module_directory=$(pwd)
+
+# build paths to directories for input/output
+workflow_results_dir="$module_directory/results/annotate_tumor_cells_output"
+sample_metadata="$module_directory/sample_metadata.tsv"
+
+# define paths to store ref and grab data
+ref_dir="$module_directory/references"
+data_dir="$module_directory/../../data/current/SCPCP000015"
```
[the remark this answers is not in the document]
allyhawkins: Just noting that I need the full paths because the scripts and notebooks are not in the same directory as this workflow. So you can't provide a path to those scripts or notebooks that are relative to where the workflow lives or the module root.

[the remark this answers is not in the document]
jashapiro: If those scripts are being run from the a different directory, they should still take in paths relative to the calling location, not where the files live. The only exception would be Rmd files, but for that I would modify your script so that when you pass in the params to the notebook, you use `normalizePath()` to get the full path. So it might look something like this (This is also a good spot for a heredoc for a bit of cleanup):
```
 Rscript -e << EOF
rmarkdown::render('$notebook_dir/01-marker-genes.Rmd', 
  clean = TRUE, 
  output_dir = '$sample_results_dir',
  output_file = '${library_id}_marker-gene-report.html',
  params = list( 
    sample_id = '$sample_id', 
    library_id = '$library_id', 
    results_dir = normalizePath('$sample_results_dir'), 
    reference_cell_file = normalizePath('$reference_cell_file') 
  ),
  envir = new.env()
)
EOF

**Comment by allyhawkins** [comment]
@jashapiro I updated this to try and simplify it as much as possible in terms of inputs. To just make sure we are on the same page, the purpose of this workflow at this stage is to serve two purposes - document the order of how things were run and make it simple to re-run the same steps for 2-3 additional samples and that's it. Once we evaluate those results then we can finalize which steps should actually remain and create a workflow that's more robust to running across more samples. I definitely don't anticipate using a shell script for that. 

I did update it here to use only three arguments, the `sample_id`, `normal_celltypes` and `tumor_celltypes`. There isn't really a way to just make it use `sample_id`, since the exact cell types to use as a reference are somewhat sample specific. I thought you didn't want to use the cell types as an argument so that's why I had moved to having a sample sheet. But if you are okay with this approach then I'll keep that and we will have to specify which cell type to use for each sample at the command line (we can also do that with a wrapper script to run just the handful of samples we want at a later time). Regardless I would like to keep the sample metadata for now since I think it will be helpful to know which reference cell types were used for which sample. 

If you are okay with the changes I've made here then I'll update the description at the top of the workflow and the README file. But I want to make sure this is what you are looking for before I do that?

**Comment by jashapiro** [comment]
> To just make sure we are on the same page, the purpose of this workflow at this stage is to serve two purposes - document the order of how things were run and make it simple to re-run the same steps for 2-3 additional samples and that's it. Once we evaluate those results then we can finalize which steps should actually remain and create a workflow that's more robust to running across more samples. I definitely don't anticipate using a shell script for that.

Yes, I think we are on the same page now!

> If you are okay with the changes I've made here then I'll update the description at the top of the workflow and the README file. But I want to make sure this is what you are looking for before I do that?

This looks good to me. It probably doesn't even need the for loop for the samples here (as long as there is one library per sample), I'd guess, but safer to have it just in case.

**Comment by allyhawkins** [comment]
Okay, I updated the documentation throughout to reflect the current state of the workflow. I believe this should be ready for another look @jashapiro!

File: analyses/cell-type-ewings/scripts/select-cell-types.R [hunk 8]
```diff
@@ -0,0 +1,96 @@
...
+# get list of normal and tumor cell types to save as refernences 
+normal_cell_types <- stringr::str_split_1(opt$normal_cells, pattern = ",")
+
+tumor_cell_types <- stringr::str_split_1(opt$tumor_cells, pattern = ",")
+
+# label cells as either tumor or normal based on which cell type was annotated 
+ref_table_df <- coldata_df |> 
+  dplyr::mutate(reference_cell_class = dplyr::case_when(
+    celltype %in% normal_cell_types ~ "Normal",
+    celltype %in% tumor_cell_types ~ "Tumor",
```
jashapiro, line 84: The one thing I might do here to make things a bit more robust to accidental errors is to just normalize everything to lower case. 

```suggestion
# get list of normal and tumor cell types to save as refernences 
normal_cell_types <- stringr::str_split_1(opt$normal_cells, pattern = ",") |> tolower()

tumor_cell_types <- stringr::str_split_1(opt$tumor_cells, pattern = ",") |> tolower()

# label cells as either tumor or normal based on which cell type was annotated 
ref_table_df <- coldata_df |> 
  dplyr::mutate(reference_cell_class = dplyr::case_when(
    tolower(celltype) %in% normal_cell_types ~ "Normal",
    tolower(celltype) %in% tumor_cell_types ~ "Tumor",
```

Probably not required (and you still have potential errors from presence or absence of `s` in `cell` vs. `cells`, but a thought.

jashapiro (approved):
Looks good. I only have some minor comments left at this stage, and I don't think I should hold you up any longer!

I would expect you should be able to add the cellassign and copykat steps in a single PR, unless you think there are a lot of changes to the scripts for those parts?

PR #440, opened 2024-05-17
Select libraries and run initial benchmarking

sjspielman:

#424

This PR begins the doublet detection benchmarking process with a script to run four different methods, from two packages `scDblFinder` and `scds`. 
The goal here is to generate some TSV file results from benchmarking, including actual doublet calls and runtimes, to be further analyzed in a notebook in a subsequent PR.

I wrote a script that takes the following steps, considering only non-multiplexed projects at this time:

- First, select a set of processed libraries from each project that have a range of library sizes (`processed_cells`)
  - Select ~7 (where possible) libraries from each project: For all libraries with >=50 cells, select the two samples with the smallest number of cells, the two with the largest number of cells, and the three in the middle. This gives us a range of dataset sizes across diagnoses.
  - Note that because I anticipate this module will eventually run on all SCEs, I didn't add anything yet to the README about input data required - should I?
- Run the four methods on each SCE, and capture results and runtime for each. 
- Export a single TSV for each of all the runtimes and results. These have been uploaded to my S3 bucket

Yup!

### Results

There are two main result files here: One TSV with runtimes for each library, and one TSV with doublet results for each library.

`s3://researcher-654654257431-us-east-2`, and results are in:
`s3://researcher-654654257431-us-east-2/doublet-detection/results/benchmark-results`

Two TSV files

These are going to be analyzed in a notebook in a subsequent PR, so not much to report here yet. 

Dependencies are in the  `renv.lock` file and the code can be run on a laptop. Note that the script has a default of 4 cores, but this can be changed with the `--cores` input argument.  

- Are my docs sufficient for this code addition?
- Some aspects of this code could probably be more modular, especially if we want to add more methods into the mix? For now I diudn't think this was something to strongly focus on, but please let me know if I should! 
- Should finding libraries to run on be in a separate script from actually running the detection, and/or should I somehow be exporting this list of libraries?

N/A

### Author checklists

_Check all those that apply._


allyhawkins:
@sjspielman before I give this a full review, can you give me a little bit more context, specifically: 

- Why are you starting with 7 libraries per project and why all projects? I understand there may be differences across projects, sure, but that's a lot of samples to start with, especially without any knowledge of how these methods might work in the context of cancer. I think I might just pick a few samples to start with and run this on those samples. You could simplify this by creating a metadata sheet with the sample ID, library ID, and any other relevant metadata and use that as input to the benchmarking script and just run through the benchmarking with those libraries. 
- Can you give me an idea of how you plan to use these results? With no information on how these libraries are going to look with doublet detection, how are we going to know what good performance is? Even if you can't find a cancer dataset as a benchmarking dataset, then I would encourage you to find a normal dataset where you are confident in the doublets so that we have something to compare to. Also, you might consider using the tabula muris dataset we use in [training to show doublet detection](https://github.com/AlexsLemonade/exercise-notebook-answers/blob/45cf50c96120dc569ce2c8405f8196e7087d3334/scRNA-seq/exercise_01-scrna_quant-solved.Rmd#L475). I don't think this is a positive control, but we already have a sense of which cells may be doublets, so that could be helpful in evaluating these methods.

analyses/doublet-detection/scripts/00_benchmark-methods.R
```diff
@@ -0,0 +1,251 @@
...
+# Run doublet detection methods and export results ------------------------------------
+
+sce_files <- libraries_to_benchmark$file_path
+names(sce_files) <- libraries_to_benchmark$scpca_library_id
+
+doublet_output <- sce_files|>
+  purrr::imap(benchmark_doublets, scratch_dir)
+
+runtime_df <- doublet_output |>
+  purrr::map(\(x) x$runtimes) |>
```
jashapiro, line 228: purrr has you covered here! You want to just select the same element? It's got you!
```suggestion
  purrr::map("runtimes") |>
```

jashapiro (commented):
I was just scanning through and had a couple little tips...

But following up on this point form Ally:
> Why are you starting with 7 libraries per project and why all projects?

If you have time to do this much benchmarking, we probably don't need the timing info. Its fine to have, but unless there was a dramatic difference, I doubt we would make a decision based on that. And if there is a dramatic difference, presumably other benchmarking papers have seen it already?

sjspielman:
> Why are you starting with 7 libraries per project and why all projects? I understand there may be differences across projects, sure, but that's a lot of samples to start with, especially without any knowledge of how these methods might work in the context of cancer. I think I might just pick a few samples to start with and run this on those samples. You could simplify this by creating a metadata sheet with the sample ID, library ID, and any other relevant metadata and use that as input to the benchmarking script and just run through the benchmarking with those libraries.

Yeah, this is more reasonable... I was hoping to track down if there might be salient differences across projects which might be meaningful to know about, but you're right that it's overkill for this stage and would be challenging to interpret as well.
How many libraries do you think would be reasonable to explore here? Maybe like 5-10? Also to @jashapiro's comment above, all runtimes are fairly quick. preview - 
Screenshot 2024-05-17 at 2 09 29 PM


> Can you give me an idea of how you plan to use these results? With no information on how these libraries are going to look with doublet detection, how are we going to know what good performance is

This has been an open question for me, and you are right that it needs to be nailed down now before I take this further - or even take it to where it is now (!). Aside from measuring runtime (spoiler: it's all fast), I was planning on comparing the specific inferences to one another e.g. with an upset plot. All this said, over the past 15 minutes I've started to question whether I've identified a fair set of methods to compare. This selection was based on best-performing methods in the literature, but three of them are from the same package (`scds`). I did not fully appreciate that this package does not actually predict doublet/singlet, vs return a score that has no objective or obvious threshold. Previous assessments of its performance have all relied on comparing to ground truth with AUC (roc and prc), so a general threshold to call results was not strictly needed. How to actually make these calls is therefore an open challenge in and of itself.

allyhawkins:
> This has been an open question for me, and you are right that it needs to be nailed down now before I take this further - or even take it to where it is now (!). Aside from measuring runtime (spoiler: it's all fast), I was planning on comparing the specific inferences to one another e.g. with an upset plot. All this said, over the past 15 minutes I've started to question whether I've identified a fair set of methods to compare. This selection was based on best-performing methods in the literature, but three of them are from the same package (scds). I did not fully appreciate that this package does not actually predict doublet/singlet, vs return a score that has no objective or obvious threshold. Previous assessments of its performance have all relied on comparing to ground truth with AUC (roc and prc), so a general threshold to call results was not strictly needed. How to actually make these calls is therefore an open challenge in and of itself.

> TLDR - maybe I need to hit the breaks and return to GitHub Discussions on this analysis 😬 !

I think you've done a good job of looking into methods, but I think the key thing to think about is how you are going to evaluate the results from these methods and what controls are you going to use? I think it would be helpful to comment on the discussion with a proposed analysis plan with steps that you plan to take. I'm sure they might change, but having a starting point of what am I measuring and how am I going to evaluate these results would be helpful. 

I would also encourage you to make a notebook with one sample and just look at some of these methods and their outputs and explore how you would use them. For this it would be good to use a control dataset rather than an ScPCA dataset at first just to get an idea of what we expect the output from each method to look like. This will also help you figure out how you can take scores from some of the methods and actually turn them into a prediction or way to classify doublets.

sjspielman:
I'm going to close this PR in favor of something better scoped for initial analysis.

PR #450, opened 2024-05-21
Fix a small typo in error message for download scripts

allyhawkins:
This just fixes a small typo I found when using the data download script where `SCPCS` was accidentally spelled with `SCPS`.


jashapiro (approved):
Oops!

Issue #443, opened 2024-05-20
Create a module-specific Docker builder workflow template in `templates/workflows`


jaclyn-taroni:
Each analysis module will have its own Docker image. We want to build and push that image anytime relevant files are altered (e.g., `analyses/{module name}/Dockerfile`, relevant lock files) to the `openscpca` public registry on ECR. We will use the module name as the image name. 

To facilitate setting this up for each module, we should create a template in `templates/workflows` that builds and pushes the images to ECR.

It should have the following features:

- The `on:` information is commented out to start, but we likely want to include:
  - Changes in PRs to `main` in the module  `Dockerfile` and the relevant files if using `renv` or `conda` trigger building
  - `workflow_dispatch`
  - `workflow_call` (not sure if we'll end up using this one, but it didn't seem like there's harm in including it!)
- We'll log in using `aws-actions/configure-aws-credentials` and `aws-actions/amazon-ecr-login` (see: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/docker-image.yml)
- Tag images using `latest` and something like `${{ github.run_number }}` or `${{ github.sha }}`
- We should only _push_ to ECR upon merge. My current thinking is that we can use dependencies between jobs for this!
  - We'll need to handle creating a repository if it doesn't exist yet. As an example:
    ```sh
    aws ecr-public create-repository --repository-name {repository} || echo "{repository} already exists!"
    ``` 

It makes sense to work through these ideas with `simulate-sce` as an example! We may want to look into Docker actions for some of these steps.

jashapiro:
closed by #474

PR #451, opened 2024-05-22
Add Docker image building for simulate-sce

jashapiro:
As part of #443, I am making a test workflow for building the docker image for the simulate-sce workflow and pushing it to Amazon ECR.

This PR should build only and not require logging in to AWS. Then on push an image should be built and pushed.

### DIFF
--- a/.github/workflows/docker_simulate-sce.yml
+++ b/.github/workflows/docker_simulate-sce.yml
@@ -15,13 +15,12 @@ on:
       - analyses/simulate-sce/conda-lock.yml
   push:
     branches:
-      - jashapiro/docker-simulate-sce
       - main
-    # paths:
-    #   - analyses/simulate-sce/Dockerfile
-    #   - analyses/simulate-sce/.dockerignore
-    #   - analyses/simulate-sce/renv.lock
-    #   - analyses/simulate-sce/conda-lock.yml
+    paths:
+      - analyses/simulate-sce/Dockerfile
+      - analyses/simulate-sce/.dockerignore
+      - analyses/simulate-sce/renv.lock
+      - analyses/simulate-sce/conda-lock.yml
 
 permissions:
   id-token: write # This is required for requesting the JWT
@@ -31,7 +30,6 @@ jobs:
   build:
     name: Build Docker Image
     runs-on: ubuntu-latest
-    environment: "prod"
 
     steps:
       - name: Configure AWS Credentials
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_simulate-sce.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: public.ecr.aws/${{ env.REGISTRY_ALIAS }}/${{ env.MODULE }}
+          # tag with semver for releases latest for main branch, ref for prs (tags with branch name)
+          tags: |
+            type=raw,value=latest,enable={{is_default_branch}}
+            type=semver,pattern={{raw}}
+            type=ref,event=pr
```
allyhawkins, line 68: Does this mean a new image is pushed every time there is a change, and we aren't using any versioned tags?

allyhawkins (approved):
This looks good, just one question I wanted to clarify.

.github/workflows/docker_simulate-sce.yml
```diff
@@ -0,0 +1,78 @@
...
+  pull_request:
+    branches:
+      - main
+    paths:
+      - analyses/simulate-sce/Dockerfile
+      - analyses/simulate-sce/.dockerignore
+      - analyses/simulate-sce/renv.lock
+      - analyses/simulate-sce/conda-lock.yml
+  push:
+    branches:
```
jashapiro, line 17:
```suggestion
  push:
    tags:
      - "v*.*.*"
    branches:
```

.github/workflows/docker_simulate-sce.yml
allyhawkins: Okay makes sense 👍

.github/workflows/docker_simulate-sce.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: public.ecr.aws/${{ env.REGISTRY_ALIAS }}/${{ env.MODULE }}
+          # tag with semantic version for releases/tags, "latest" for main branch pushes, and the branch name for PRs
```
jashapiro, line 66:
```suggestion
          # tag with 'latest' for main branch pushes, semantic version for releases/tags, and the branch name for PRs
```

.github/workflows/docker_simulate-sce.yml
jashapiro: The real test will be if it works as expected on merge!

PR #452, opened 2024-05-22
Allyhawkins/cellassign to workflow

### LINKED ISSUE #435: Add running CellAssign to workflow for annotating tumor cells in Ewings

#292 

After completing #434, we will have the start of a workflow for annotating tumor cells in Ewing's cells. This workflow is being used to run various methods for annotating tumor cells on a single sample. 

This issue is to track addition of running the script for `CellAssign` using the various marker gene refs and then reading all of those in to render the notebook for exploring `CellAssign`. 

This PR will contain new steps in the workflow that both run `CellAssign` and render the exploratory notebook. Part of this will also include pulling out creating the reference matrices to a separate script rather than doing that within the exploratory notebook. 

When doing this we will also have to update the way we handle grabbing the normal cells to create the reference from PanglaoDB since the normal cells identified in each sample aren't exactly the same. We should use the output cell reference table created in the first step of the workflow in #434 to do this. 

No since we already have the setup to run `CellAssign`. 

_No response_

_No response_

### LINKED ISSUE #436: Add running CopyKAT to workflow for annotating tumor cells in Ewings

#292 

After completing https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/434, we will have the start of a workflow for annotating tumor cells in Ewing's cells. This workflow is being used to run various methods for annotating tumor cells on a single sample.

This issue is to track addition of running the script for `CopyKAT` with and without a reference of normal cells and rendering the notebook looking at `CopyKAT` output. 

This will include new steps in the workflow to run the `CopyKAT` script twice - once with normal cells as reference and once without normal cells. We may have to slightly adjust the script to take as input a reference table rather than list of normal cells following changes in #434. 

This will also include a step to render the notebook for `CopyKAT`. 

No, we already have everything set up to run `CopyKAT`. 

_No response_

_No response_

allyhawkins:

Closes #435 
Closes #436 

This PR adds running both `CellAssign` and `CopyKAT` to the `annotate-tumor-cells.sh` workflow. This includes both running the `run-cellassign.py` and `run-copykat.R`  scripts prior to rendering each of the notebooks. I also made some small adjustments to the workflow and removed some things that I didn't think were necessary to run on all samples in the notebooks. 

- There is a new script for generating the references to use for `CellAssign`. This includes a reference using just marker genes, the all visser marker genes, and then one that is endothelial cells and fibroblasts from panglao and tumor marker genes. Although not all of the samples in SCPCP000015 have a solid group of normal cells that are endothelial and/or fibroblast, I am fairly confident that at least one of those cell types _should_ be present in each sample so I think this is a good reference to start with. Like I said previously, we may want to consider using a much larger reference generated from Panglao and incorporates tumor marker genes, but that's a different question outside the scope of this PR. 
- I added running `CellAssign` for each of the references for a given sample and then rendering the report. I spent far too much time trying to get bash to do too much with associative arrays, but in the end decided to just define the paths to each reference and output predictions file here. I don't think we need to get that complicated at this stage. 
- When adding running both `CellAssign` and `CopyKAT` I decided to only run them if the output object/ predictions file didn't already exist yet. This should make it easier to re-run and re-render notebooks if necessary. Also, it's helpful if something fails to not have to re-run all the time consuming steps. 
- I removed anything related to the filtered marker genes reference in the `CellAssign` notebook since that wasn't really useful. Also because we are now creating the refs and running `CellAssign` separately, all of that code was removed from the notebook. 
- I moved storing the GTF file to the `scratch` directory. 
- For `CopyKAT`, I updated the script and notebook to use the data stored in the final rds object output by `CopyKAT` rather than reading in the larger matrices. 
- In the `CopyKAT` plots I am labeling any cells that were used as reference, similar to what we just added for `InferCNV`. 
- For all methods, I am now saving a tsv file with the final annotations from the specified method. 

Yes

The output from running the workflow on SCPCS000490. 

No additional results beyond what is already in the exploratory notebook at this point. 

You will need at least 4 CPUs to run this and I expect that the CopyKAT and InferCNV steps will take a few hours each. 

### Author checklists

_Check all those that apply._


jashapiro (approved):
This all looks good to me. 

Now that you have a script that runs all the parts, I think you should next make a GHA to run the shell script on a sample of simulated data, which we will use for CI testing. File an issue for this first, but I don't want it to wait long before we implement it. 

You should be able to follow the pattern of https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/run_hello-R.yml for most of it, I hope, though I guess you will also need a step to install the conda dependencies, so you may need part of `run_hello-python.yml` as well.

You probably first will want to do some of your own testing with the simulated data, if you have not already, just in case there are errors that pop up due to smaller cell counts, etc.

Issue #446, opened 2024-05-21
Explore doublet detection methods


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#424
Also related to #425 

### Describe the goals of the changes to the analysis module.

This issue tracks implementing the first steps in the existing (but largely empty!) `doublet-detection` module. The specific goal is to compare performance on a few "ground truth" datasets used in previous benchmarking studies (including [this one](https://doi.org/10.1016/j.cels.2020.11.008)), available from this [Zenodo repository](https://doi.org/10.5281/zenodo.4562782).

We'll aim to use these four datasets, chosen due to their varying library sizes and putative cell types of origin:

- `hm-6k` (N=6806): Mixture of human HEK293T and mouse NIH3T3 cells
	- Previous benchmarking showed excellent performance on these datasets potentially since ground-truth doublet annotation focused on species differences
	- Although this is cell line data which may not be directly comparable to our scpca data*, it does provide a "best case scenario" 
- `pdx-MULTI` (N=10296): PDX of human breast cancer, with mouse immune
- `HMEC-orig-MULTI` (N=26426): Human primary mammary epithelial cells
- `pbmc-1B-dm` (N=3790): PBMCs from patient with systemic lupus erythematosus
	- Previous benchmarking showed poorer performance on this dataset potentially since ground-truth doublet annotation may have considered homotypic doublets

### What will your pull request contain?

This issue encompasses two main goals, each of which is expected to be its own PR:

1. Run doublet detection methods on each dataset using an R and python script. We'll use three methods, each of which operates on a raw counts matrix:
	- `scDblFinder` (R)
	- `cxds` (R)
		- Specifically, `scDblFinder` [calculates a version of this this score](https://github.com/plger/scDblFinder/blob/99b947a49a4f3f3bd1a2203fe2296a6dc143fe11/R/misc.R#L266-L290) that is more robust to low sparsity which we'll use here. The main reason we'll use this score is that it's normalized to `[0,1]`, which is much more interprettable than the unbounded scores reported by the `scds::cxds()` function.
	- `scrublet` (python)
2. (Noting this could be >1 PR) Explore doublet inferences:
	- Explore distribution of scores and how they relate to the applied threshold (provides insight into the threshold itself)
	- Measure balanced accuracy for each method at a given chosen threshold
	- Visualize singlet/doublet calls in PC space
[...]

PR #454, opened 2024-05-22
Run doublet detection on ground truth data

sjspielman:

#446

This PR runs two (with a secret third!) doublet detection methods across four ground-truth datasets. The methods are `scDblFinder` (which runs a version of `cxds` for free) and `scrublet`. Results are saved in `results/benchmark_results` for further exploration in a subsequent PR.

I wrote a couple different scripts to make this all happen, as documented in `scripts/README.md`. 
First, I have a shell script to download the files from Zenodo, which also calls an R script to format into SCE and AnnData. My reasoning here was getting files into ScPCA-esque formats will make code more adaptable for future actual use.
Then, I have an R script that runs `scDblFinder`, and a python script that runs `scrublet`. At one point I thought about using `reticulate`, but then decided I'd rather not get into the R vs python environment weeds, and again keep things modular for each language for future portability.

There is also an overall script `run_doublet-detection.sh` to wrap everything.

I sure do!

Everything is here: `s3://researcher-654654257431-us-east-2/doublet-detection`

- SCE files with scDblFinder doublet results embedded
- TSV files with scrublet results 

None yet - that's for the next PR where I analyze these outputs in a notebook.

- Can be run on a laptop.
- The module has both `renv` and `conda` environments, and directions for the latter are in `README.md`.

Have I gotten docs everywhere they need to be so far?

Worth noting that the `Dockerfile` is _not_ currently up-to-date with the environment, and currently considers only `renv`. I will need to get `conda` dependencies in here too, but waiting to do that until I take some time to evaluate a good base environment that could be used for this case.

### Author checklists

_Check all those that apply._


analyses/doublet-detection/environment.yml
```diff
@@ -0,0 +1,14 @@
+name: openscpca-doublet-detection
+channels:
+  - conda-forge
+  - bioconda
+  - defaults
+dependencies:
```
jashapiro, line 6: Add platforms for easier `conda-lock` usage and slightly smaller files (we don't want the default list that includes windows)
```suggestion
platforms:
  - linux-64
  - osx-64
  - osx-arm64
dependencies:
```

analyses/doublet-detection/scripts/01b_detect-doublets.py
```diff
@@ -0,0 +1,48 @@
...
+        input_anndata = dataname + "_anndata.h5ad"
+        result_tsv = dataname + "_scrublet.tsv"
+
+        adata = anndata.read_h5ad( Path(data_dir / input_anndata) )
+        scrub_results = run_scrublet(adata)
+
+        with open(Path(result_dir / result_tsv), "w") as f:
+            f.write("barcode\tdoublet_score\tpredicted_doublets\n")
+            for x in range(len(scrub_results[0])):
+                f.write("\t".join( [scrub_results[0][x], scrub_results[1][x], scrub_results[2][x]] ) + "\n")
```
jashapiro, line 46: This is a "more pythonic" way of doing the same thing.

The `*` is "splat" to give zip each list/tuple in `scrub_results` as an argument. Then zip gives the aligned tuples.

I use the splat again with `print` because that allows me to join numbers (or anything with a string method), which `"\t".join()` doesn't like.
```suggestion
            for values in zip(*scrub_results)):
                print(*values, sep = "\t", file = f)
```

jashapiro (commented):
I gave some python style comments...

analyses/doublet-detection/scripts/00b_format-benchmark-data.R
```diff
@@ -0,0 +1,56 @@
...
+# Each is an RDS object containing a list of two items:
+#  [[1]] A raw counts matrix
+#  [[2]] A vector of doublet/singlet calls for each barcode
+# The exported files are named `<dataset name>-<sce/anndata>.<rds/h5ad>.
+# Each has a variable `ground_truth_doublets` representing the singlet/doublet calls:
+#   - In the SCE file, this is in the colData slot
+#   - In the AnnData file, this is in the obs slot
+
+# Load renv
+renv::load("../")
```
allyhawkins, line 15: If you run this script from anywhere else this will break... I would define the project root and then load that.

allyhawkins:
```suggestion
project_root <- here::here()
renv::load(project_root)

```

analyses/doublet-detection/scripts/00b_format-benchmark-data.R
```diff
@@ -0,0 +1,56 @@
...
+
+library(optparse)
+library(SingleCellExperiment)
+library(zellkonverter)
+
+
+option_list <- list(
+  make_option(
+    c("-d", "--dir"),
+    type = "character",
```
allyhawkins, line 25: You could use `project_root` to define the default directory here.

analyses/doublet-detection/scripts/00b_format-benchmark-data.R
```diff
@@ -0,0 +1,56 @@
...
+# We're only focusing on these datasets for benchmarking
+datanames <- c("hm-6k", "pbmc-1B-dm", "pdx-MULTI", "HMEC-orig-MULTI")
+
+# Perform conversion
+datanames |>
+  purrr::walk(
+    \(dataname) {
+      input_file <- file.path(opts$dir, glue::glue("{dataname}.rds"))
+      output_sce_file <- file.path(opts$dir, glue::glue("{dataname}_sce.rds"))
+      output_anndata_file <- file.path(opts$dir, glue::glue("{dataname}_anndata.h5ad"))
```
allyhawkins, line 42: I would save these to a different directory? Maybe create folders for each dataset and then have the sce and anndata inside those folders. That might make it easier to parse those into a notebook later on.

analyses/doublet-detection/scripts/00b_format-benchmark-data.R
```diff
@@ -0,0 +1,56 @@
...
+      mat <- dat[[1]]
+      calls <- dat[[2]]
+
+      # Create and export SCE
+      sce <- SingleCellExperiment(assays = list(counts = mat))
+      sce$ground_truth_doublets <- calls
+      readr::write_rds(sce, output_sce_file)
+
+      # Export AnnData version
+      writeH5AD(sce, output_anndata_file)
```
allyhawkins, line 54:
```suggestion
      zellkonverter::writeH5AD(sce, output_anndata_file)
```
I think you are going to need the package here?

analyses/doublet-detection/scripts/00b_format-benchmark-data.R
```diff
@@ -0,0 +1,56 @@
...
+      input_file <- file.path(opts$dir, glue::glue("{dataname}.rds"))
+      output_sce_file <- file.path(opts$dir, glue::glue("{dataname}_sce.rds"))
+      output_anndata_file <- file.path(opts$dir, glue::glue("{dataname}_anndata.h5ad"))
+
+      dat <- readRDS(input_file)
+      mat <- dat[[1]]
+      calls <- dat[[2]]
+
+      # Create and export SCE
+      sce <- SingleCellExperiment(assays = list(counts = mat))
```
allyhawkins, line 49: make sure you add `library(SingleCellExperiment)` to the top of the script.

analyses/doublet-detection/scripts/01a_detect-doublets.R
```diff
@@ -0,0 +1,90 @@
...
+                                    ...)
+  }
+
+  return(sce)
+}
+
+# Infer doublets --------------
+
+# Benchmark dataset names
+datanames <- c("hm-6k", "pbmc-1B-dm", "pdx-MULTI", "HMEC-orig-MULTI")
```
allyhawkins, line 77: In thinking about being able to reuse these scripts with SCE objects, they aren't going to have these dataset names. So I think I would actually provide an argument where I input a list of the SCE files to use. Alternatively you could make this work on a single SCE file and then in the bash script run it 4x, once for each dataset.

analyses/doublet-detection/scripts/01a_detect-doublets.R
```diff
@@ -0,0 +1,90 @@
...
+
+
+# Functions -----------
+
+#' Run scDblFinder on an SCE object
+#'
+#' @param sce SCE object to run scDblFinder on
+#' @param cores Number of cores to specify to BiocParallel::MulticoreParam(
+#' @param sample_var Required for multiplexed data, the SCE colData variable that indicates sample of origin
+#' @param random_seed Required for multiplexed data, a random seed to ensure reproducibility when running on multiplexed data.
```
allyhawkins, line 37: Why do we only want a seed for multiplexed data?

analyses/doublet-detection/scripts/01a_detect-doublets.R
```diff
@@ -0,0 +1,90 @@
...
+# Detect doublets and export SCE file with inferences
+datanames |>
+  purrr::walk(
+    \(dataname) {
+      input_sce_file <- file.path(data_dir, glue::glue("{dataname}_sce.rds"))
+      output_sce_file <- file.path(result_dir, glue::glue("{dataname}_sce.rds"))
+
+      readRDS(input_sce_file) |>
+        run_scdblfinder(cores = opts$cores) |>
+        readr::write_rds(output_sce_file)
```
allyhawkins, line 88: rather than save the RDS file, why don't we just grab the results as a dataframe? We don't need the whole object and then it would match the AnnData object.

analyses/doublet-detection/scripts/01b_detect-doublets.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # Define directories
+    openscpca_base = pyprojroot.find_root(pyprojroot.has_dir(".git"))
+    module_base = Path(openscpca_base / "analyses/doublet-detection")
+
+    data_dir = Path(module_base / "scratch/benchmark_datasets")
+    result_dir = Path(module_base / "results/benchmark_results")
+    result_dir.mkdir(parents = True, exist_ok = True)
+
+    datanames = ["hm-6k", "pbmc-1B-dm", "pdx-MULTI", "HMEC-orig-MULTI"]
```
allyhawkins, line 34: I think same thing here. You can get rid of a lot of wrangling of filenames by just providing the filename as the input.

analyses/doublet-detection/scripts/01b_detect-doublets.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+
+    scrub = scr.Scrublet(adata.X)
+    doublet_scores, predicted_doublets = scrub.scrub_doublets()
+
+    # stringify
+    doublet_scores_str = [str(x) for x in doublet_scores]
+    predicted_doublets_str = [("doublet" if x else "singlet") for x in predicted_doublets]
+
+    return( (list(adata.obs_names), doublet_scores_str, predicted_doublets_str) )
```
allyhawkins, line 22:
```suggestion
    results_df = pd.DataFrame(barcodes = adata.obs_names, scrublet_score = doublet_scores_str, srublet_prediction = predicted_doublets_str)

    return(results_df)
```
Why not just turn this into a dataframe and then you can directly save to tsv file?

analyses/doublet-detection/scripts/01b_detect-doublets.py
allyhawkins: I think you can avoid this all together if you use pandas following my suggestion in the function. 

```suggestion
scrub_results.to_csv(result_tsv, sep="\t", index=False)
```

allyhawkins (commented):
Overall organization looks pretty good. My main comment is about making the decision to have these scripts run on any SCE or any AnnData, rather than hard coding in the filenames for benchmarking. I think I would add a few more options to the scripts, including the SCE file and an output file and then have the script run on a single sample. You already have a bash script for running the benchmarking datasets so you can loop through the script with the specific files you are interested in there. That will make it easier to re-use these scripts when we get to ScPCA data.

analyses/doublet-detection/scripts/01b_detect-doublets.py
jashapiro: I would agree with this (but don't convert scores to strings). I thought about suggesting it, but wasn't sure whether you had `pandas` in the environment already.

analyses/doublet-detection/scripts/00b_format-benchmark-data.R
sjspielman: It was loaded at the top! But I can unload it and just use this?

analyses/doublet-detection/scripts/01a_detect-doublets.R
sjspielman: I linked this below on line 56: see section 1.5.6  here https://bioconductor.org/packages/3.19/bioc/vignettes/scDblFinder/inst/doc/scDblFinder.html#usage. Apparently `set.seed()` in the R script is enough for single samples only, and our main circumstance for that would be multiplexed data (that said, we don't actually _perform_ demultiplexing currently to be able to identify this, but who knows maybe it'll happen during OpenScPCA!). Either way, I should update this doc to say "multiple samples" instead of "multiplexed data"

analyses/doublet-detection/scripts/01a_detect-doublets.R
sjspielman: > Alternatively you could make this work on a single SCE file and then in the bash script run it 4x, once for each dataset.

I think this is the move!

analyses/doublet-detection/scripts/00b_format-benchmark-data.R
allyhawkins: Sorry I thought I looked for where you loaded libraries and couldn't find it, but I see now that it's there so you can ignore this comment and the one about `SingleCellExperiment`.

analyses/doublet-detection/scripts/01a_detect-doublets.R
sjspielman: Worth noting I was indeed missing a seed more generally though, and in the python script(!)

analyses/doublet-detection/scripts/01a_detect-doublets.R
```diff
@@ -0,0 +1,126 @@
...
+if (!file.exists(input_sce_file)) {
+  stop(
+    glue::glue("Could not find input file, expected at: `{input_sce_file}`.")
+  )
+}
+output_tsv_file <- file.path(opts$results_dir, glue::glue("{opts$dataset_name}_scdblfinder.tsv"))
+
+readRDS(input_sce_file) |>
+  run_scdblfinder(cores = opts$cores) |>
+  readr::write_tsv(output_tsv_file)
```
sjspielman, line 126:
```suggestion
  readr::write_tsv(output_tsv_file)
 
```

sjspielman:
Code has now been heavily modularized and, I think, much improved; thanks for all the reviews! And bonus: the code now uses seeds 😄 Results are now all TSV files in my bucket.

I am hoping the code in here is "self-explanatory" from the docs 🤞 - if I'm very wrong, please stop reviewing and let me know!

analyses/doublet-detection/scripts/01b_detect-doublets.py
```diff
@@ -0,0 +1,88 @@
...
+    if not args.data_dir.exists():
+        print(
+            "A correct path to the input data must be provided with --data_dir.",
+            file=sys.stderr
+        )
+        sys.exit(1)
+    args.results_dir.mkdir(parents = True, exist_ok = True)
+
+    # Run scrublet and export the results
+    input_anndata = args.dataset_name + "_anndata.h5ad"
```
allyhawkins, line 80: Maybe a good idea to explicitly check that this file exists rather than the directory.

allyhawkins (commented):
I think the reorg looks much better and a lot easier to follow! I just had a few remaining comments that I think should be addressed before merging.

analyses/doublet-detection/scripts/01b_detect-doublets.py
```diff
@@ -0,0 +1,88 @@
...
+
+def main() -> None:
+
+    parser = argparse.ArgumentParser(
+        description="Detect doublets on a set of AnnData objects using scrublet.",
+    )
+    parser.add_argument(
+        "--dataset_name",
+        type=str,
+        default="",
```
jashapiro, line 43: You may want to set this as `required=True` rather than setting the default like this?

analyses/doublet-detection/scripts/01b_detect-doublets.py
```diff
@@ -0,0 +1,88 @@
...
+    parser.add_argument(
+        "--random_seed",
+        type=int,
+        default=2024,
+        help="Random seed to ensure reproducibility."
+    )
+    args = parser.parse_args()
+
+    # Prepare input arguments
+    if args.dataset_name == "":
```
jashapiro, line 65: Or if you want to handle this yourself... an empty string is evaluated as falsey, so you can just use that.
```suggestion
    if not args.dataset_name:
```

analyses/doublet-detection/scripts/00_format-benchmark-data.R
```diff
@@ -0,0 +1,61 @@
...
+# Each has a variable `ground_truth_doublets` representing the singlet/doublet calls:
+#   - In the SCE file, this is in the colData slot
+#   - In the AnnData file, this is in the obs slot
+
+# Load renv
+project_root <- here::here()
+renv::load(project_root)
+
+library(optparse)
+
```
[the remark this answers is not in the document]
sjspielman: I ended up going with `SingleCellExperiment::` when used since the only thing we're doing here is creating the object.

analyses/doublet-detection/scripts/01b_detect-doublets.py
jashapiro, line 80: I'd also just set these as paths here, since you don't need the separate file names anywhere: 
```suggestion
    input_anndata = args.data_dir / args.dataset_name + "_anndata.h5ad"
```

analyses/doublet-detection/scripts/01a_detect-doublets.R
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env Rscript
+
+
+# Load libraries and renv environment ------
+library(SingleCellExperiment)
+library(optparse)
+
+project_root <- here::here()
+renv::load(project_root)
```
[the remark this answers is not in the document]
jashapiro: There is a circularity here in that you won't have these libraries unless you already loaded the project or already installed separately...

jashapiro (commented):
A few more little thoughts

analyses/doublet-detection/scripts/01b_detect-doublets.py
sjspielman: noting this needed more `Path()` - `args.data_dir / Path(args.dataset_name + "_anndata.h5ad")`

analyses/doublet-detection/scripts/01b_run-scrublet.py
```diff
@@ -0,0 +1,85 @@
...
+        default=2024,
+        help="Random seed to ensure reproducibility."
+    )
+    args = parser.parse_args()
+
+    args.results_dir.mkdir(parents = True, exist_ok = True)
+
+    # Run scrublet and export the results
+    input_anndata = args.data_dir / Path(args.dataset_name + "_anndata.h5ad")
+    result_tsv = args.results_dir / Path(args.dataset_name + "_scrublet.tsv")
```
jashapiro, line 70: `Path / str` is perfectly valid, but you do need to account for order of operations, as `Path + str` is not: 
```suggestion
    input_anndata = args.data_dir / (args.dataset_name + "_anndata.h5ad")
    result_tsv = args.results_dir / (args.dataset_name + "_scrublet.tsv")
```

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
```diff
@@ -0,0 +1,120 @@
...
+}
+
+# Run doublet finder 
+result_df <- scDblFinder::scDblFinder(
+  sce,
+  samples = sample_var # Default is NULL so use whatever was provided by the user or default 
+  BPPARAM = bp,
+  returnType = "table", # return df, not sce
+  ...
+)
```
allyhawkins, line 49: I think this all needs to be indented.

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
```diff
@@ -0,0 +1,120 @@
...
+opts <- parse_args(OptionParser(option_list = option_list))
+
+# Check input arguments
+if (opts$dataset_name == "") {
+  stop("Must provide a dataset name with --dataset_name.")
+}
+if (!dir.exists(opts$data_dir)) {
+  stop("Must provide an input path to data files with --data_dir.")
+}
```
allyhawkins, line 100: You probably don't need to check both that the data directory exists and the input file exists. I think just the input file is fine, but if you want to leave this also okay.

allyhawkins (approved):
A few small comments, but otherwise looks good!

sjspielman:
@allyhawkins I ended up doing one more thing in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/454/commits/2b1410c8ac5fcf20015f6efc45ee1e8ca4eccb95 - added a `sample_var` arg to the script and pass it along with the seed into the scdblfinder function, in case future us would like this.

PR #455, opened 2024-05-22
Add script to move counts in simulated data

jashapiro:

The `simulate-sce` module did not output processed AnnData files in the same format as the scpca-nf script

https://github.com/AlexsLemonade/OpenScPCA-nf/issues/47

Make it uniform!

I ported the [`move_counts_anndata.py` script](https://github.com/AlexsLemonade/scpca-nf/blob/development/bin/move_counts_anndata.py) script we have in `scpca-nf`, but decided to use `basilisk/reticulate` to perform the actions to prevent having to create another python environment and to allow us to use the same docker image.

Nope, though I will port this over to `openscpca-nf`

Does it seem to work as expected? Look good?

### Author checklists

_Check all those that apply._

### DIFF
--- a/analyses/simulate-sce/scripts/move-anndata-counts.R
+++ b/analyses/simulate-sce/scripts/move-anndata-counts.R
@@ -1,0 +1,48 @@
+#!usr/bin/env Rscript
+
+# Script to move the logcounts layer to the X slot in an anndata object
+
+library(basilisk)
+library(optparse)
+
+# Parse arguments --------------------------------------------------------------
+# set up arguments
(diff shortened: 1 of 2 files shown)


jashapiro:
This module does not currently have a GHA... should we add one? Probably not really needed, since this code will run in openscpca-nf, but we could...

allyhawkins (approved):
LGTM. I tested this locally and then opened up the file in Python to make sure that the new format was correct. 

As for a GHA, I think for consistency it would be nice to have but isn't a priority.

Issue #453, opened 2024-05-22
Create GHA for cell typing ewings module


allyhawkins:
Now that you have a script that runs all the parts, I think you should next make a GHA to run the shell script on a sample of simulated data, which we will use for CI testing. File an issue for this first, but I don't want it to wait long before we implement it. 

You should be able to follow the pattern of https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/run_hello-R.yml for most of it, I hope, though I guess you will also need a step to install the conda dependencies, so you may need part of `run_hello-python.yml` as well.

You probably first will want to do some of your own testing with the simulated data, if you have not already, just in case there are errors that pop up due to smaller cell counts, etc.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/452#pullrequestreview-2071957780_

PR #456, opened 2024-05-23
Fix bugs in annotating tumor cells to account for simulated data

allyhawkins:

Related to #453 

Here I am prepping the workflow to be able to run with simulated data so that we can create a GHA. In doing so, I had to fix a few bugs and make some changes to account for scenarios that I hadn't seen in the sample that I was testing. 

I downloaded the test data for SCPCS000490 and ran it through the workflow and fixed bugs I found along the way. 

- I had to add two new command line arguments to the workflow for the data directory and the results directory. I didn't want to overwrite my real results with test results and then I needed the data directory to be directed to the test data. 
- The main change I had to make throughout the notebooks was accounting for the scenario where there is no marker gene expression. I took a pretty simple approach where if no marker genes are found in the dataset, none of the plots looking at marker gene expression are created. Additionally, if there are any comparisons to marker gene annotations, those get skipped in the other notebooks. 
- For `CellAssign`, I had to make a minor change to the check for `raw.X`. For some reason just checking `raw` was not working as expected with the real data, but changing to `raw.X` fixes that problem. 
- Also `CellAssign` complains if none of the marker genes are expressed in the dataset and everything is 0 in the input, which makes sense. So I added a check to make sure they aren't all 0. I also updated the report to account for any issues with running `CellAssign`. 
- In this scenario, `CopyKAT` reports aneuploid and diploid low confident cells. I assume this actually might show up in our real data too. The main change I made there was just filtering out any low confident calls in the confusion matrix and only comparing the confident calls. 

Next up is a GHA that runs this workflow on the simulated data. 

### Author checklists

_Check all those that apply._

### DIFF
--- a/analyses/cell-type-ewings/annotate-tumor-cells.sh
+++ b/analyses/cell-type-ewings/annotate-tumor-cells.sh
@@ -30,26 +30,26 @@ sample_id="SCPCS000491" ./annotate-tumor-cells.sh
 
 set -euo pipefail
 
-# input variables
-sample_id=${sample_id:-"SCPCS000490"}
-normal_celltypes=${normal_celltypes:-"Endothelial cells,endothelial cell"}
-tumor_celltypes=${tumor_celltypes:-"Pulmonary vascular smooth muscle cells,smooth muscle cell"}
-threads=${threads:-4}
-
 # this script lives in the root of the module directory
 # use this to define other default paths
 cd $(dirname "$0")
 module_directory=$(pwd)
 
-# path to input data folder
-data_dir="../../data/current/SCPCP000015"
+# input variables
+sample_id=${sample_id:-"SCPCS000490"}
+normal_celltypes=${normal_celltypes:-"Endothelial cells,endothelial cell"}
+tumor_celltypes=${tumor_celltypes:-"Pulmonary vascular smooth muscle cells,smooth muscle cell"}
+data_dir=${data_dir:-"../../data/current/SCPCP000015"}
+workflow_results_dir=${workflow_results_dir:-"${module_directory}/results/annotate_tumor_cells_output"}
+threads=${threads:-4}
+
+echo $workflow_results_dir
 
 # define paths to notebooks and scripts run in the workflow
 notebook_dir="template_notebooks"
 scripts_dir="scripts"
 
 # define results directories
-workflow_results_dir="${module_directory}/results/annotate_tumor_cells_output"
 sample_results_dir="${workflow_results_dir}/${sample_id}"
 
 cellassign_results_dir="${sample_results_dir}/cellassign"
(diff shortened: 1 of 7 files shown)


analyses/cell-type-ewings/annotate-tumor-cells.sh
```diff
@@ -30,26 +35,26 @@ sample_id="SCPCS000491" ./annotate-tumor-cells.sh
...
+# this script lives in the root of the module directory
+# use this to define other default paths
+cd $(dirname "$0")
+module_directory=$(pwd)
+
 # input variables
 sample_id=${sample_id:-"SCPCS000490"}
 normal_celltypes=${normal_celltypes:-"Endothelial cells,endothelial cell"}
 tumor_celltypes=${tumor_celltypes:-"Pulmonary vascular smooth muscle cells,smooth muscle cell"}
+data_dir=${data_dir:-"../../data/current/SCPCP000015"}
```
jashapiro, line 47: Just noting for the purposes of GHA, this will still be the correct directory, as the link `current->test` will be created.

analyses/cell-type-ewings/template_notebooks/01-marker-genes.Rmd
```diff
@@ -147,6 +147,15 @@ umap_df <- sce |>
 
 
 ```{r}
+if(sum(umap_df$gene_expression) == 0){
```
jashapiro, line 150: Just adding a comment here, for clarity. 
```suggestion
# test if _any_ marker genes have expression
if(sum(umap_df$gene_expression) == 0){
```

analyses/cell-type-ewings/template_notebooks/01-marker-genes.Rmd
```diff
@@ -147,6 +147,15 @@ umap_df <- sce |>
...
+  has_marker_gene_exp <- FALSE
```
jashapiro, line 151: Why don't we add a warning here (and in similar blocks) to indicate why plots are missing.

jashapiro (commented):
LGTM. A bit sad that so much gets thrown out, but that is not unexpected, I guess. 

My only real suggestion is to add some warning text for the case where plots will be skipped.

analyses/cell-type-ewings/annotate-tumor-cells.sh
allyhawkins: Oh good to know!

allyhawkins:
> My only real suggestion is to add some warning text for the case where plots will be skipped.

I added messages to print at the top of the notebook if marker genes aren't found. I think all the other cases (no CellAssign results), I had messages so should be good for another look.

jashapiro (approved):
LGTM!

PR #457, opened 2024-05-23
Add GHA to cell-type-ewings module

### LINKED ISSUE #453: Create GHA for cell typing ewings module

allyhawkins:

Closes #453 

Here I am adding a workflow to run the `annotate-tumor-cells.sh` script to GHA. 

I mostly used the steps that were used in the `hello-R` and `hello-python` module to set up both `renv` and `conda`. 

While working on this, I realized that the conda lock file was generated with a previous version of the environment and hadn't been updated since we started using `pip` for installing `scvi`.  When I went to update the lock file using the existing environment, I was getting errors that it couldn't locate `torch`. After some searching, it looks like conda lock will search for dependencies in `conda` and not in `pip`, unless you specifically list the `pip` dependency. So I added in `torch` and then was able to successfully create the conda lock file. 

Yes 

I want to file this so I can test it and will request review once things complete successfully. 


.github/workflows/run_cell-type-ewings.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Set up conda
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-activate-base: true
+          activate-environment: ""
+
+      - name: Install and activate locked conda environment
+        run: |
+          conda install conda-lock
+          conda-lock install --name test ${MODULE_PATH}/conda-lock.yml
```
jashapiro, line 63: My guess about what is going in is some kind of incomaptibility with `basilisk` or another R package.  Two solutions I can think of would be to move the conda install before the R install, or what I am suggesting here: explicitly activating the `test` environment. `setup-miniconda` is supposed to do that, but seems to be failing for some reason.

```suggestion
      - name: Set up conda
        uses: conda-incubator/setup-miniconda@v3

      - name: Install and activate locked conda environment
        run: |
          conda activate test
          conda install conda-lock
          conda-lock install --name test ${MODULE_PATH}/conda-lock.yml
```

jashapiro: Oh, never mind, I remembered the subtle thing that has to happen if you are using conda: setting the shell to `bash`: Removing my suggestion above, and reverting to what you had first, then adding the line you need:
```suggestion
      - name: Set up conda
        uses: conda-incubator/setup-miniconda@v3

      - name: Install and activate locked conda environment
        run: |
          conda install conda-lock
          conda-lock install --name test ${MODULE_PATH}/conda-lock.yml
```

.github/workflows/run_cell-type-ewings.yaml
```diff
@@ -0,0 +1,68 @@
...
+  pull_request:
+    branches:
+      - main
+    paths:
+      - analyses/cell-type-ewings/**
+jobs:
+  run-module:
+    runs-on: ubuntu-latest
+
+    steps:
```
jashapiro, line 29: This is the fix!
```suggestion
    runs-on: ubuntu-latest
    defaults:
      run:
        shell: bash -el {0}

    steps:
```

jashapiro (commented):
Took me a second to remember the problem here: conda needs `bash`!

allyhawkins:
Finally got this to pass 🎉  This should be ready for review!

.github/workflows/run_cell-type-ewings.yaml
```diff
@@ -0,0 +1,68 @@
...
+        with:
+          working-directory: ${{ env.MODULE_PATH }}
+
+      - name: Set up conda
+        uses: conda-incubator/setup-miniconda@v3
+
+      - name: Install and activate locked conda environment
+        run: |
+          conda install conda-lock
+          conda-lock install --name openscpca-cell-type-ewings ${MODULE_PATH}/conda-lock.yml
```
jashapiro, line 63: I see that in your script you are using `conda run`
https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/8937a0aed3e36837b2f785e85dbf38a9dbe705a2/analyses/cell-type-ewings/annotate-tumor-cells.sh#L136

I think this is probably a bit of overhead and specificity that should not be needed. If you activate the conda environment before running the script, would this work the same way? The only way I would potentially see a problem would be if your conda environment also included an R installation. Is that the case? 

If not, I would probably leave the `name` here as `test` and modify the script to call the component scripts directly, without `conda run`.

analyses/cell-type-ewings/template_notebooks/02-cellassign.Rmd
```diff
@@ -9,11 +9,11 @@ output:
...
-  marker_gene_classification: "../results/annotate_tumor_cells_output/SCPCS000490/SCPCL000822_tumor-normal-classifications.tsv"
-  tumor_marker_predictions: "../results/annotate_tumor_cells_output/SCPCS000490/cellassign/SCPCL000822_tumor-only-predictions.tsv"
-  visser_marker_predictions: "../results/annotate_tumor_cells_output/SCPCS000490/cellassign/SCPCL000822_visser-predictions.tsv"
-  panglao_predictions: "../results/annotate_tumor_cells_output/SCPCS000490/cellassign/SCPCL000822_panglao-predictions.tsv"
-  results_dir: "../results/annotate_tumor_cells_output/SCPCS000490"
+  marker_gene_classification: "../results/sim_annotate_tumor/SCPCS000490/SCPCL000822_tumor-normal-classifications.tsv"
+  tumor_marker_predictions: "../results/sim_annotate_tumor/SCPCS000490/cellassign/SCPCL000822_tumor-marker_predictions.tsv"
+  visser_marker_predictions: "../results/sim_annotate_tumor/SCPCS000490/cellassign/SCPCL000822_visser-all-marker_predictions.tsv"
+  panglao_predictions: "../results/sim_annotate_tumor/SCPCS000490/cellassign/SCPCL000822_panglao-endo-fibro_predictions.tsv"
+  results_dir: "../results/sim_annotate_tumor/SCPCS000490"
```
jashapiro, line 16: I feel like you probably want to leave these defaults as the full output directory and change them in the rendering script itself as needed?

I also expect you may not want all of these defined as `params`, but you might instead want to build these paths within the notebook from `sample_id`, `library_id` and `results_dir`? I see from the outer script why you might want to pass these in, but in that case I think it might make sense to leave the defaults as generic names, rather than explicit paths.

For the outer script, I might add a variable that designates if this is a test run, so you can change the path for test data on the fly without having to modify any other defaults.

jashapiro (commented):
Great that you got it running!

I had a suggestion about simplifying the outer script just a bit, because I don't think we want scripts to assume conda environments in general. If you can't call the full script by activating the environment first and then having no conda calls inside, then your solution may be the only way that we can go, but that should be documented a bit more.

The other comment is about some of the complexity of the file paths that have to be passed around, and what the defaults should be. In practice, if running from the outer script, the defaults should not matter, but it is better if they can be not _too_ complicated.

.github/workflows/run_cell-type-ewings.yaml
allyhawkins: > I think this is probably a bit of overhead and specificity that should not be needed. If you activate the conda environment before running the script, would this work the same way? The only way I would potentially see a problem would be if your conda environment also included an R installation. Is that the case?

No I don't have R in the conda environment right now, which is why I couldn't just activate the environment and then run the script. I can add R into the environment and try again?

analyses/cell-type-ewings/template_notebooks/02-cellassign.Rmd
allyhawkins: The reason I have full paths here is because the predictions files have to be defined in the outer script since they are used both to run cell assign and to render the notebook. I thought it would be easier to maintain if the full paths to these files were defined only in the outer script and not in both the script and the notebook. 

I can adjust the defaults to be just generic file names, this is mostly left over from me testing things and trying to figure out why something was failing. 

> For the outer script, I might add a variable that designates if this is a test run, so you can change the path for test data on the fly without having to modify any other defaults.

I'm not totally sure what you mean here? You would just change the `data_dir` argument if you wanted to explicitly specify test vs real data. I don't think adding another argument really helps here.

.github/workflows/run_cell-type-ewings.yaml
jashapiro: It should use the system R version if there is no R in the conda environment, which is what you want! There should not be any other changes needed.

analyses/cell-type-ewings/template_notebooks/02-cellassign.Rmd
jashapiro: Probably not, but you would want to change both `data_dir` and `workflow_results_dir`, right? Or really just `workflow_results_dir` as the download script should handle switching the input directories?

https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/8937a0aed3e36837b2f785e85dbf38a9dbe705a2/analyses/cell-type-ewings/annotate-tumor-cells.sh#L47C1-L48C104

.github/workflows/run_cell-type-ewings.yaml
allyhawkins: But to run it locally you would need to activate the environment and restore renv before being able to run the whole script. I just thought it was simpler to keep the R and conda things separate because I was getting issues just trying to run it locally, regardless of GHA.

analyses/cell-type-ewings/template_notebooks/02-cellassign.Rmd
allyhawkins: You would only need to change `data_dir` if you want to switch back and forth between what's considered "current". So if you have both test data and real data downloaded, then you would want to specify which input to use. Otherwise it's directed to whatever you downloaded more recently. For the purposes of GHA, it's only downloading the test data, so the "current" directory will automatically link to the test data. 

You would only need to change the `workflow_results_dir` if you want to save the output to a different folder. This would also only need to happen if you are working locally with both test and regular data from the same samples. In the GHA, there is only one output so this doesn't need to be changed. I guess you could create a test flag rather than changing the `workflow_results_dir`, but I thought we were keeping the number of arguments here to a minimum?

.github/workflows/run_cell-type-ewings.yaml
jashapiro: I would expect the following commands run in the module directory to set everything up as needed:

```
Rscript -e "renv::restore()"
conda activate openscpca-cell-type-ewings

./annotate-tumor-cells.sh
```

When I do that, it is still using the system R and the correct conda environment. The renv restore is needed either way, and the conda environment should not interfere. What errors were you running into?

allyhawkins:
```
(base) allyhawkins@allys-work-mac cell-type-ewings % Rscript -e "renv::restore()"
- The library is already synchronized with the lockfile.
(base) allyhawkins@allys-work-mac cell-type-ewings % conda activate openscpca-cell-type-ewings
(openscpca-cell-type-ewings) allyhawkins@allys-work-mac cell-type-ewings % ./annotate-tumor-cells.sh 
/Users/allyhawkins/Documents/ALSF/forked_repos/OpenScPCA-analysis/analyses/cell-type-ewings/results/annotate_tumor_cells_output
Starting workflow for SCPCS000490, SCPCL000822
Saving cell references...
- One or more packages recorded in the lockfile are not installed.
- Use `renv::status()` for more details.
- One or more packages recorded in the lockfile are not installed.
- Use `renv::status()` for more details.
Error in library(SingleCellExperiment) : 
  there is no package called ‘SingleCellExperiment’
Calls: suppressPackageStartupMessages -> withCallingHandlers -> library
Execution halted
```

jashapiro (approved):
Okay, LGTM as it is working!

Well, mostly... For some reason cellassign does not seem to want to run on my computer (I get a segfault), but that seems to be some kind of problem with pytorch that I may have to figure out on my own... 

*followup: seems like it runs fine on my machine with one thread, but not multithreaded*

The only thing I am going to ask you to do is to update the readme a bit with instructions about activating the conda environment and running renv::restore() before running the workflow.

allyhawkins:
> followup: seems like it runs fine on my machine with one thread, but not multithreaded

Ah yes, I had this issue too but was able to run just fine on the server or a machine with more threads to start. I updated the readme to also mention this in addition to the note about activating the environment.

PR #459, opened 2024-05-24
Add a single workflow for docker image builds

jashapiro:

In #451, I added a test workflow for building a single docker image. That was good for testing, but it is going to be annoying to have many workflows doing basically the same thing, and keeping track of them was likely to be difficult. 

Here I am creating a single workflow that _should_ be able to build all docker images on a merge to main, or when a release is tagged (or on manual dispatch)

Part of, or a partial alternative to #443 

Simplify management of docker building

The workflow has two jobs: an initial job to get the list of modules with changed docker images and a secondary matrix job to build and push all of the images. 

The initial job check out the full repo, then uses `git diff` to get a list of files that have changed. This list is filtered with `grep` to only the files that are likely to affect the docker image, then reduced to just a list of the modules.

For `push` actions, this module list is then compared to the `DOCKER_MODULES` list that is defined (as a JSON string) at the top of the workflow using `jq` and the result is saved to `$GITHUB_OUTPUT`. Github matrix actions don't like empty arrays, so there is a fallback to the array `["None"]` if there are no modules returned.

For `workflow_dispatch` actions, the full `DOCKER_MODULES` list is saved to `$GITHUB_OUTPUT`.

The second job takes the matrix of modules and builds all of the docker images, pushing them to AWS ECR with appropriate tags. If the module is `None`, all steps are skipped.

I also simplified the current workflow for `simulate_sce` to _just_ do the build on PRs, with no pushing, and hence no need to interact with AWS at all.

There are also changes to the renv for the `hello-r` module for the purposes of making sure this runs when merged as a test.

Yes: see further discussion.

At a minimum, I will still need to create a template. (This was done in #464)

I did some testing to make sure that this worked as expected, at least as far as , but I also expect to encounter bugs when this gets into the wild.

I can't decide whether to consolidate all of the PR actions into one in a similar way to this, but that might be worth doing too, as an alternative to #443.

(The required `git diff` command for PRs is a bit different from merges, so it is going to be easier to keep those separate, especially as the PR action can skip all of the AWS stuff)

One other thing we might want to add is some kind of alert for when a build fails. We also might want to allow an input at `workflow_dispatch` to specify a single module?

### Author checklists

_Check all those that apply._


jashapiro:
> The concerns that come immediately to mind for me (not DRY, multiple workflows) either seem minor or easily surmountable with good naming, which is what the "templatization" should accomplish.
> 
> Looking at this, I think it would be harder for me to debug than what we originally planned (and ease of debugging is why we're in GHA-land). You also state that you expect bugs.

These are fair points, and I agree that more explanation is needed! 

There are two components in this action that I think we can separate out discussion of, "find changed modules" and "build and push the images".

Taking the build/push images part first: I think this is generally straightforward, and matches almost exactly the current approach in https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/docker_simulate-sce.yml. 
One thing I discovered when writing that one though was that in order for it to run on PRs I had to remove the `environment` restriction, as only the `main` branch can use that. But I don't think there is any additional security afforded by that right now, as removing that did not affect the ability to run the workflow using the github openid login. If we can lock that down (I think we should) that would mean we would need separate jobs for build tests and pushing. This is not terrible: we would have multiple copies of the second job in this action, without the matrix component. The biggest problem is that if we make a change, it has to be made in all versions of the file... 

The first part is where I expected potential bugs, and where most of the complexity lies, to be sure. The goal here was to prevent having to rerun all of the docker builds on every push to main, but maintain the ability to build all of the images if we wanted to. The annoying part is that there is no easy way to  get a list of changed files from GHA, and no way I could find to pass an array directly between jobs. So there is some ugly stuff in there, to be sure, and making that simpler would be good.

I'm trying to think about the right alternative... I have a thought that it may be possible to do this with separate actions per module, each with two jobs, one of which runs only on PRs that matches this PR for `simulate_sce` and the other on pushes and workflow dispatches (or workflow calls) that matches the matrix job here. Then we could have a separate action that runs only on dispatch to rebuild all, if needed. 

Let me code that up as an alternative PR, and we can compare how they would look in practice. I will move this back to draft status.

jaclyn-taroni:
I'm not sure this will assuage your concerns because I'm not sure I follow yet (so the alternative PR is very welcome)... I'm pretty sure we could push the image to ECR under the following conditions:

- On `push` to `main` when certain files in a module are impacted
- When using `workflow_dispatch`
- When using `workflow_call`

For example:

```yaml
on:
  push:
    branches:
      - main
    paths: [ {relevant paths} ]
  pull_request:
    branches:
      - main
    paths: [ {relevant paths} ] 
  workflow_dispatch:
  workflow_call:
```

Then the build push step would contain:

```
      - name: Build Docker image
        uses: docker/build-push-action@v5
        with:
          push: ${{ github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.name == 'workflow_call' }}
```

I would expect that `push` is typically true when you are on the `main` branch, and then you could lock down the OIDC context to just the `main` branch.

One thing I don't know, though, is how we expect that would impact the build cache situation if you're building on a PR (i.e., not on the `main` branch).

jashapiro:
closing in favor of #474

Issue #461, opened 2024-05-24
Add GHA for doublet-detection module


sjspielman:
### Describe the goals of the changes to the analysis module.

This issue tracks adding a GitHub Actions workflow to run the `doublet-detection` module.

### What will your pull request contain?

A workflow yaml file

### Will you require additional software beyond what is already in the analysis module?

N/A

### Will you require different computational resources beyond what the analysis module already uses?

N/A

sjspielman:
Closed by #462

PR #462, opened 2024-05-24
GHA for doublet-detection module

sjspielman:

#461 

Note that should be considered "stacked on" #454 

This PR adds a GHA for the module, including a comment placeholder for we'll probably want to download data eventually, later in this module's development.

It may be too early in this module's development to be filing this PR, so it's fine if it hangs out for a little bit since there shouldn't be any conflicts!.


.github/workflows/run_doublet-detection.yml
```diff
@@ -0,0 +1,62 @@
...
+  workflow_dispatch:
+  workflow_call:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - analyses/doublet-detection/**
+jobs:
+  run-module:
+    runs-on: ubuntu-latest
```
jashapiro, line 29: This is needed for conda environments to work properly
```suggestion
  run-module:
    runs-on: ubuntu-latest
    defaults:
      run:
        shell: bash -el {0}
```

sjspielman:
> this is also an argument for using rprojroot::find_root(rprojroot::is_renv_project) or similar for more explicit root

Indeed!!! Will wait for after this build. If `renv` doesn't get cached...oooof.

sjspielman:
@jashapiro do you have any insight into why the conda-lock environment doesn't seem to be fully activated/available here? https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/9270012159/job/25502032871#step:9:15143

allyhawkins (approved):
LGTM!

sjspielman:
Wanted to get an opinion here before merging @allyhawkins @jashapiro - running this workflow takes quite a bit of time! I'm wondering if, for now, we want to remove the trigger to run on every change to this module, and instead just add to the [workflow that runs all modules](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/run_all-modules.yml). Notably, I did not actually add to the run-all workflow in this PR, so I suppose I'm also seeking opinions on should I go ahead and do that!

jashapiro:
> Wanted to get an opinion here before merging @allyhawkins @jashapiro - running this workflow takes quite a bit of time! I'm wondering if, for now, we want to remove the trigger to run on every change to this module, and instead just add to the [workflow that runs all modules](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/run_all-modules.yml). Notably, I did not actually add to the run-all workflow in this PR, so I suppose I'm also seeking opinions on should I go ahead and do that!

You are running the workflow on all your benchmarking datasets, in full. This is not something we want to have happen in general. The goal of these actions is to see that the code works with example data, so they should not be running with real data in GHA.

If you have a small dataset that you can download and run instead, that would be the preferred solution: you can set a ENV variable to select that option when running your wrapper script. At a minimum, you should run the script on only _one_ dataset, but it would be better to find something completely different, if you can. 

For now, I would probably completely disable this action. In general, I don't think benchmarking is  going to be something we want running repeatedly in GHA, so it will not be a big problem. When you get to running on the ScPCA samples, you will use simulated data anyway. 

For now, I would probably disable this workflow entirely.

sjspielman:
> If you have a small dataset that you can download and run instead, that would be the preferred solution: you can set a ENV variable to select that option when running your wrapper script. 

I had been wondering about this too. I think this is a separate PR, to not automatically download the full benchmark zip, but include an option to use smaller `zip` file with dummy data. I'd also have to change the dataset array here, too. 

For now, I'll just turn it off. Thanks!

Issue #444, opened 2024-05-20
Add GitHub Action workflow creation as part of `create-analysis-module.py`


jaclyn-taroni:
Blocked by #443.

We already have two workflow templates in `templates/workflows`. We are adding one for building and pushing Docker with #443. We should create workflows for a module when someone creates a module via the `create-analysis-module.py` script.

An expected follow-up to this issue is documenting this new capability in `docs/`.

sjspielman:
The module creation script now creates the workflow, and this behavior has been documented. 

Going to leave this issue open for now given this active PR: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/459

jashapiro:
Closed with #474

PR #464, opened 2024-05-24
Create workflow templates on module creation

jashapiro:

Part of #444 , in combination with #459

Adds workflow templates to `create-analysis-module.py`

Pretty simple addition of a few more copy with modification entries. 

I added a very basic docker workflow that only runs on PRs to test that the image builds, which goes along with #459 which does all the building. We _may_ want to change this to an integrated workflow for all modules, but that would be a separate PR.

The run workflows are likely to vary more, so I expect those to stay distinct. I did do some renaming for consistency and sorting.

All workflows are inactive on pull requests by default.


create-analysis-module.py
```diff
@@ -259,6 +260,40 @@ def main() -> None:
...
+        gha_template_source = "run_renv-module.yml"
+    else:
+        gha_template_source = "run_conda-module.yml"
+
+    gha_template = workflows_dir / f"run_{args.name}.yml"
+    copy_file_with_tag_replacement(
+        src=workflow_template_dir / gha_template_source,
+        dest=gha_template,
+        tag="openscpca_module",
+        replacement=args.name,
```
sjspielman, line 278:
```suggestion
        replacement=args.name
```

more generally though I've noticed that you often have commas at the end of a set of arguments into a function - is this intentional or otherwise strategic?

sjspielman (commented):
This looks good to me overall. Can you also update `templates/workflows/README.md` to match these changes?

Also FYI, I'm going to open a separate issue(s) about docs we'll need to update based on these changes.

PR #466, opened 2024-05-24
Add download script option to download metadata

### LINKED ISSUE #445: Add data download script option to download metadata
This issue tracks adding an option to the `download-data.py` script to just download the sample metadata files.

jashapiro:
Closes #445

I did this a pretty simple way, by just setting the formats and stages to empty sets before download if the metadata option was set. It seems to work, but may need more testing.

While I was at it, I let the argument validation complete before exiting, so people will see multiple errors at once, rather than having to fix one only to get another.

### DIFF
--- a/analyses/simulate-sce/scripts/move-anndata-counts.R
+++ b/analyses/simulate-sce/scripts/move-anndata-counts.R
@@ -1,4 +1,4 @@
-#!usr/bin/env Rscript
+#!/usr/bin/env Rscript
 
 # Script to move the logcounts layer to the X slot in an anndata object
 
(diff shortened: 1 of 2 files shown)


download-data.py
```diff
@@ -311,10 +311,18 @@ def main() -> None:
         help="Download test data from the test bucket and direct the `current` symlink to the test data directory."
         " To switch back, rerun this script with the `--release current` option.",
     )
+    parser.add_argument(
+        "--metadata-only",
+        action="store_true",
+        help="Download only the metadata files and not the data files."
+        " To also download QC reports, combined with the --include-reports option."
```
sjspielman, line 318:
```suggestion
        " To also download QC reports, combine with the --include-reports option."
```

download-data.py
```diff
@@ -382,6 +390,17 @@ def main() -> None:
             "Using both `--projects` and `--samples` options together is not supported.",
             file=sys.stderr,
         )
+        validation_error = True
+
+    if args.metadata_only and args.samples:
+        print(
+            "Using both `--metadata-only` and `--samples` options together is not supported.",
```
sjspielman, line 397: I think either "both" or "together", but not (sorry) both?

```suggestion
            "Using `--metadata-only` and `--samples` options together is not supported.",
```

sjspielman (commented):
This all looks good to me, and works fine when run. The only thing I might also change is the Download Summary. Currently, it's (for example having just downloaded metadata for project `SCPCP000001`) - 
```
Download Summary
Release: 2024-05-01
Data Format: 
Processing levels: 
Projects: SCPCP000001
Data download location: /Users/sjspielman/ALSF/open-scpca/OpenScPCA-analysis/data/2024-05-01
Total download size (includes previous downloads): 24.68 KiB
```

I think we might want something like the following for these two lines?
```
Data Format: Metadata only 
Processing levels: N/A
```

analyses/simulate-sce/scripts/move-anndata-counts.R
jashapiro, line 1: Bug fix + some autoformatting.

sjspielman (approved):
🚀

PR #468, opened 2024-05-28
439 Update sync script module flag

### LINKED ISSUE #439: Update sync script module flag
The -m module argument is currently needed to specify the module name in the results syncing script. We should document that behavior, or instead update the script to make it a required argument sans flag.

jashapiro:
Closes #439

While the modules flag was already required, I decided that for simplicity we should just make it a positional argument, so that is what I did here.

I updated error messages as well, and the docs to match the new format.

Finally, I removed a script that we do not need that was meant to be part of the openscpca-nf repo.


scripts/sync-results.py
```diff
@@ -27,10 +27,9 @@ def main() -> None:
         description="Sync data from an analysis module to your S3 bucket.",
     )
     parser.add_argument(
-        "--module",
-        "-m",
+        "module",
+        metavar="MODULE",
```
sjspielman, line 31: noticed while I was here - maybe change the help below to "The name of the analysis module to sync **RESULTS** from.", (not data)

sjspielman (approved):
LGTM! Script behaves as expected.

I left 1 small comment on code not formally in this PR.

scripts/sync-results.py
sjspielman: hmmm... "to sync results _for_"? either way.

allyhawkins (approved):
I would incorporate Stephanie's suggestion but otherwise LGTM

scripts/sync-results.py
```diff
@@ -27,10 +27,9 @@ def main() -> None:
...
         type=str,
-        required=True,
         help="The name of the analysis module to sync data from.",
```
jashapiro, line 33:
```suggestion
        help="The name of the analysis module to sync results from.",
```

sjspielman (approved):
actually approving!

PR #469, opened 2024-05-28
Explore benchmark results, part 1

sjspielman:

Part of #446

The goal of this PR is to explore the performance of doublet detection methods on the benchmarking datasets, one dataset at a time.
The PR does not attempt to _compare_ results - that's next up!

This PR adds a notebook template, `exploratory-analyses/02_explore-benchmark-results.Rmd`, to explore the benchmarking results on one dataset at a time. This notebook is run from `run_doublet-detection.sh`, and produces 1 notebook per dataset saved in `results/benchmark-results/rendered-notebooks` (naming comments are generally welcome!). Since they are saved to `results`, the HTMLs are not under version control.

I also updated the script that prepares benchmark data to also export TSVs of the ground truth results, so I can read those into the notebook too.

To start, I added two main things to the notebook:
- Jitter plots to visualize scores and calls for each method
- I used `caret` to calculate various performance stats

We can certainly expand on these analyses here, and I hope to get some feedback during PR review on what else might be useful!
For example, UMAPs might actually not be a bad thing here (or, we can save that for the next notebook which will _compare_ methods directly?).

It's also worth noting that `scDblFinder` and `scrublet` apply their own threshold, which I used here. But, these thresholds are known to not necessarily be great for every dataset! Tailoring this threshold to a given dataset, however, is not very straightforward, and certainly not in an automated fashion, so I "trusted" their thresholds, for now.
Along these lines, there is no threshold for calling `cxds` scores; I ended up testing out all of 0.5, 0.75, and 0.9 here.

I sure do.

### Results

_Delete this section if no results are associated with your pull request._

_In this section, tell reviewers about what kinds of results (if any) your code produces._

Everything is here:

`s3://researcher-654654257431-us-east-2/doublet-detection/results/benchmark-results/`

A few notebooks, available at:
`s3://researcher-654654257431-us-east-2/doublet-detection/results/benchmark-results/rendered-notebooks/`

I've made this table with a brief overview of the results from the notebooks. `scrublet` is a clear winner here, based on this metric at least!

| Dataset           | Cell count | Method        | Balanced accuracy |
| ----------------- | ---------- | ------------- | ----------------- |
| `HMEC-orig-MULTI` | 26426      | `scDblFinder` | 0.7046            |
|                   |            | `scrublet`    | 0.7805            |
|                   |            | `cxds, 0.5`   | 0.8166            |
|                   |            | `cxds, 0.75`  | 0.8553            |
|                   |            | `cxds, 0.9`   | 0.8787            |
| `pdx-MULTI`       | 10296      | `scDblFinder` | 0.7086            |
|                   |            | `scrublet`    | 0.8455            |
|                   |            | `cxds, 0.5`   | 0.5446            |
|                   |            | `cxds, 0.75`  | 0.5446            |
|                   |            | `cxds, 0.9`   | 0.5446            |
| `hm-6k`           | 6806       | `scDblFinder` | 0.7046            |
|                   |            | `scrublet`    | 0.9685            |
|                   |            | `cxds, 0.5`   | 0.9980            |
|                   |            | `cxds, 0.75`  | 0.9919            |
|                   |            | `cxds, 0.9`   | 0.9879            |
| `pbmc-1B-dm`      | 3790       | `scDblFinder` | 0.6635            |
|                   |            | `scrublet`    | 0.6919            |
|                   |            | `cxds, 0.5`   | 0.5673            |
|                   |            | `cxds, 0.75`  | 0.6510            |
|                   |            | `cxds, 0.9`   | 0.4828            |

- Any feedback on naming?
- Any additional visualizations or calculations you'd recommend for this notebook?
- How should I document these results? For example, should I add something like the table I wrote for this PR to the results README?

There is a `renv` and `conda` environment to set up, as described in the README.

-

-


analyses/doublet-detection/exploratory-analyses/02_explore-benchmark-results.Rmd
```diff
@@ -0,0 +1,267 @@
...
+
+doublet_df <- dplyr::inner_join(
+  scdbl_df,
+  scrub_df, 
+  by = "barcodes"
+) |>
+  dplyr::inner_join(
+    true_df, 
+    by = "barcodes"
+  )
```
allyhawkins, line 158: Why use `inner_join` here? Theoretically every barcode should have a ground truth and this is fine, but in the case that there is no ground truth for a barcode, I think we want to know that and not drop that barcode.

analyses/doublet-detection/exploratory-analyses/02_explore-benchmark-results.Rmd
```diff
@@ -0,0 +1,267 @@
...
+
+assess_doublets(
+  doublet_df,
+  "scrublet_prediction"
+) |>
+  print()
+```
+
+
+### cxds results
```
allyhawkins, line 204: I think it would be helpful here to show the distributions of the score before doing any cutoffs? Maybe a density plot colored by ground truth so you can compare what scores look like for true singlets and true doublets.

analyses/doublet-detection/exploratory-analyses/02_explore-benchmark-results.Rmd
allyhawkins: Maybe because you don't want any NAs when creating the confusion matrix, but I think we want to see those in the jitter plot. You can filter any NAs out before creating the confusion matrix?

analyses/doublet-detection/run_doublet-detection.sh
```diff
@@ -8,15 +8,19 @@ set -euo pipefail
 MODULE_DIR=$(dirname "${BASH_SOURCE[0]}")
 cd ${MODULE_DIR}
 
+NB_DIR=exploratory-analyses # directory with exploratory notebooks
```
allyhawkins, line 11: I think it would make more sense to store this in `teplate-notebooks`. I feel like exploratory analysis is something that you are running without parameters and we want to commit both the Rmd file and the rendered notebook. For this, you have just created a template so I would move that there.

allyhawkins (commented):
This looks like a pretty good start. I had a few overall comments that I think need to be addressed: 

- Somewhere in the docs can you include some information about what each dataset being used for benchmarking contains? It's hard to tell what type of data I'm looking at based on the name alone. 
- I think we want PCA plots here where you color the PCA plot by the score produced by the tool, the final call output by the tool, and then the ground truth. 
- I think it might be helpful also to see the total number of cells that are classified as doublet vs. singlet for the ground truth. I might add a section first where you just look at the ground truth. This would include a table showing you how many doublets vs singlets there are and the PCA showing where the doublets and singlets are located. My concern here is that some of these benchmarking datasets appear to have a large amount of doublets in them. I think some of the methods we are using are based on an assumption that doublets will be minimal, so because of that I would assume they won't perform as well on those datasets. I expect our datasets to be a situation where we have a small proportion of doublets. 
- Then for each of the methods sections I would add those same things - how many of each cells are identified and what does the PCA look like. Then show the jitter plots and confusion matrices. 
- Specifically for cxds I would plot the distribution of the score and color by ground truth (and add in the PCA) before going into trying to set cutoffs. 

Just a note here that some of the jitter plots here are really concerning to me where you see a lot of cells that have a ground truth of doublets and a score that is associated with a singlet and vice versa.

analyses/doublet-detection/exploratory-analyses/02_explore-benchmark-results.Rmd
sjspielman: No real reason for this to be inner_join, will update!

sjspielman:
This has now been heavily updated to incorporate reviews, including some more context and plots! Thanks for the suggestions on what else to explore here.

There are two changes in addition to the notebook:

- I reverted a previous change I made to export a TSV with ground truth values in favor of just using these values as saved in the SCE, since we'll need PC coordinates for the notebook. So, I re-modified that initial formatting script to also normalize & calculate PCA before SCE export.
	- I could also have kept that TSV and exported PC coordinates in that TSV file?
	- Currently, the SCE files are only in `scratch`, not `results`. Do we think this is ok?
- I updated the `scDblFinder` code to only return results for _real_ doublets, not also the artificial ones simulated during inference. These extra rows were the reason I had had `inner_join()`, and now there's no longer a need.

allyhawkins (commented):
Thanks for adding in the new plots! They are really helpful and definitely informative. The biggest take home for me looking at this is that the performance of methods varies across dataset types... The hm-6k is the dataset that looks most like what we would expect when there are clear doublets. That was expected since it's human/mouse mixture so it's going to be the easiest to find doublets. 

But I will say all the other datasets generally don't look good for any methods. The HMEC-orig-multi dataset looks particularly bad for all methods. This dataset also has the highest percentage of doublets to begin with which is also expected given the fact that HMECs are sticky. Although you are seeing high balanced accuracies, what concerns me is the score distributions for the ground truth calls and the jitter plots. There are a lot of cells that have very low scores that are called as doublets in the ground truth but singlets by the method. 

How were the ground truths for each of these datasets identified? I'm just wondering how much we should trust the ground truths here. 

I left some minor comments throughout that should be addressed before merging, but I also think it might be worth thinking about next steps here. I know you want to compare the methods which should probably still be done, but it just worries me that on the more real datasets they all look bad.

sjspielman:
> How were the ground truths for each of these datasets identified? I'm just wondering how much we should trust the ground truths here.

I can add some of this into the notebook dataset descriptions. Here's the information I have from the paper, which is a little thin (but could go further digging). Based on these descriptions, the ground truth may indeed not be entirely perfect...:
- `hm-6k`: "A droplet was annotated as a doublet if its barcode was associated with both human and mouse."
- `pbmc-1B-dm`: "Doublets were annotated by demuxlet"
- `pdx-MULTI`:  "Doublet [sic] were annotated by following the data processing pipeline avail- able at https://github.com/chris-mcginnis-ucsf/MULTI-seq."
- `HMEC-orig-MULTI`: "The GEO accession and preprocessing pipeline of this dataset are the same as those of the pdx-MULTI dataset." (aka, they used the `MULTI-seq` pipeline)


> I also think it might be worth thinking about next steps here. I know you want to compare the methods which should probably still be done, but it just worries me that on the more real datasets they all look bad.

:100: all around here. Some thoughts:

1. It might not hurt to run through the rest of the ground truth datasets? It doesn't take very long, and it's not too much to look over manually. Maybe we'll see some informative performance differences... 
2. I have some optimism that comparisons will also be actually useful. For example, does the intersection among all methods have a closer relationship to the ground truth? In this analysis, I'd plan to do: i) upset plots on each dataset, ii) PCAs colored by method overlap. I'll probably use a `cxcs` cutoff of 0.5, iii) confusion matrices between ground truth and "intersection among all methods" calls.
	- less technically, I'm on the fence if this should be a separate notebook so each notebook doesn't get too big, or if it's better to have it in the same notebook for one-stop-shop viewing? What do you think?

analyses/doublet-detection/template-notebooks/02_explore-benchmark-results.Rmd
```diff
@@ -0,0 +1,587 @@
...
+First, we assess the ground truth doublet distributions to contextualize the doublet calls.
+
+
+```{r}
+percent <- print_doublet_table(doublet_df, "ground_truth", is_ground_truth = TRUE)
+```
+
+```{r results = "asis"}
+percent
+```
```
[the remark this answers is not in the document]
sjspielman: Ugh, turns out the double printing is that "bug" we've seen before where notebooks just like to double print certain values when rendering...

allyhawkins:
> > How were the ground truths for each of these datasets identified? I'm just wondering how much we should trust the ground truths here.
> 
> I can add some of this into the notebook dataset descriptions. Here's the information I have from the paper, which is a little thin (but could go further digging). Based on these descriptions, the ground truth may indeed not be entirely perfect...:
> 
> * `hm-6k`: "A droplet was annotated as a doublet if its barcode was associated with both human and mouse."
> * `pbmc-1B-dm`: "Doublets were annotated by demuxlet"
> * `pdx-MULTI`:  "Doublet [sic] were annotated by following the data processing pipeline avail- able at https://github.com/chris-mcginnis-ucsf/MULTI-seq."
> * `HMEC-orig-MULTI`: "The GEO accession and preprocessing pipeline of this dataset are the same as those of the pdx-MULTI dataset." (aka, they used the `MULTI-seq` pipeline)

Ah so this is really helpful! They are just using other methods to call doublets and I wouldn't really call this a "ground truth".  Those methods are designed to demultiplex any samples that have been multiplexed in a library so I'm confused how they are using it to call doublets in scenarios where all cells come from the same sample. I think the ground truth is probably not as reliable as we think. 

> > I also think it might be worth thinking about next steps here. I know you want to compare the methods which should probably still be done, but it just worries me that on the more real datasets they all look bad.
> 
> 💯 all around here. Some thoughts:
> 
> 1. It might not hurt to run through the rest of the ground truth datasets? It doesn't take very long, and it's not too much to look over manually. Maybe we'll see some informative performance differences...
> 2. I have some optimism that comparisons will also be actually useful. For example, does the intersection among all methods have a closer relationship to the ground truth? In this analysis, I'd plan to do: i) upset plots on each dataset, ii) PCAs colored by method overlap. I'll probably use a `cxcs` cutoff of 0.5, iii) confusion matrices between ground truth and "intersection among all methods" calls.

I think looking at the intersection definitely makes sense! I think generally we are going to take an approach where we look for the consensus across multiple methods, so this plan works for me. 

>    * less technically, I'm on the fence if this should be a separate notebook so each notebook doesn't get too big, or if it's better to have it in the same notebook for one-stop-shop viewing? What do you think?

This depends on if you plan on comparing methods one sample at a time or making a combined notebook where you read in calls from all methods and all benchmarking datasets. If you are still doing it one sample at a time, you probably could add a section below these plots and look at the merge, but if you want to look at it all in one notebook to see how the intersection looks across methods and datasets, then I would do a separate one.

sjspielman:
>  I think the ground truth is probably not as reliable as we think.

Agree, I added a sentence to the top of the notebook to caveat this.

> I think looking at the intersection definitely makes sense! I think generally we are going to take an approach where we look for the consensus across multiple methods, so this plan works for me.

Do you have any particular thoughts on running more of these real datasets through? We can of course circle back to that later, as well and keep it these four for now though.

> but if you want to look at it all in one notebook to see how the intersection looks across methods and datasets, then I would do a separate one.

I think I prefer this approach!

allyhawkins:
> Do you have any particular thoughts on running more of these real datasets through? We can of course circle back to that later, as well and keep it these four for now though.

I think we probably can keep it at this 4 for now. It will be easy to add more later if we want to look at more.

sjspielman:
Noting I was able to get all your review comments, except for the double printing :/

PR #470, opened 2024-05-29
Make sync verbose again

jashapiro:
In #468 I thought I was making the sync script be a bit nicer about errors, but I was really removing all of the output, which was not great. So here I am reverting a line of that change.


sjspielman (approved):
🦎

PR #471, opened 2024-05-29
Add Dockerfile for cell-type-ewings module 

### LINKED ISSUE #463: Update docker image for cell-type-ewings module

_No response_

This issue is to track updating the `Dockerfile` in `cell-type-ewings` to include all dependencies. This should include both dependencies captured in the `renv.lock` and `conda.lock`. 

An update to the existing Dockerfile which is currently just a template. 

_No response_

_No response_

_No response_

allyhawkins:

Closes #463 

Here I updated the Dockerfile for the `cell-type-ewings` module to contain both the renv and conda dependencies needed to run the scripts and analysis present in the module. 

I started with a Bioconductor base image and added libraries specified in `renv.lock` and created and activated the conda environment. A few hiccups that I ran into: 

- For some reason I was continually getting an error about `GenomicAlignments` v1.40.0 not being available. I was originally using the `bioconductor/r-ver:3.19` base image, but that wasn't working as I expected. Although I noticed we do use this same base image in `scpcaTools` and also have the same version of `GenomicAlignments`? For now, I am using the `bioconductor/r-ver:devel-R-4.4.0`, but I'm open to any suggestions to fix this. 
- I originally tried to use `conda-lock` to install and activate the environment, but I had the `COPY conda-lock.yml conda-lock.yml` command present and I kept getting an error that the `conda-lock.yml` file didn't exist. I copied the file name exactly and still no luck. So I decided to just use conda and create an environment from the `environment.yml` file. 

I was able to successfully build the image and then run it interactively to confirm that everything that was expected to be present is there. 

Yes. 

Just noting that I used `docker build -t cell-type-ewings .` to build the image and then `docker run -ti cell-type-ewings` to run the image in my terminal. 

### Author checklists

_Check all those that apply._

### DIFF
--- a/analyses/cell-type-ewings/Dockerfile
+++ b/analyses/cell-type-ewings/Dockerfile
@@ -1,9 +1,61 @@
-# A template docker file for creating a new analysis
-FROM ubuntu:22.04
+# Dockerfile for cell-type-ewings module
+FROM bioconductor/r-ver:devel-R-4.4.0
 
 # Labels following the Open Containers Initiative (OCI) recommendations
 # For more information, see https://specs.opencontainers.org/image-spec/annotations/?v=v1.0.1
+LABEL org.opencontainers.image.title="openscpca/cell-type-ewings"
+LABEL org.opencontainers.image.description="Docker image for the OpenScPCA analysis module 'cell-type-ewings'"
 LABEL org.opencontainers.image.authors="OpenScPCA scpca@ccdatalab.org"
-LABEL org.opencontainers.image.source="https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/templates/analysis-module"
+LABEL org.opencontainers.image.source="https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-ewings"
 
+# Disable the renv cache to install packages directly into the R library
+ENV RENV_CONFIG_CACHE_ENABLED FALSE
+# set a name for the conda environment
+ARG ENV_NAME=openscpca-cell-type-ewings
 
+# set environment variables to install miniconda
+ENV PATH="/root/miniconda3/bin:${PATH}"
+ARG PATH="/root/miniconda3/bin:${PATH}"
+
+### Renv setup
+
+# Copy the renv.lock file from the host environment to the image
+COPY renv.lock renv.lock
+
+# Install renv to enable later package installation
+RUN Rscript -e "install.packages('renv')"
+
+# Install dependencies for renv library
+RUN apt-get -y update &&  \
+    DEBIAN_FRONTEND=noninteractive \
+    apt-get install --no-install-recommends -y \
+    pandoc \
+    wget \
+    && rm -rf /var/lib/apt/lists/*
+
+# restore from renv.lock file and clean up to reduce image size
+RUN Rscript -e 'renv::restore()' && \
+  rm -rf ~/.cache/R/renv && \
+  rm -rf /tmp/downloaded_packages && \
+  rm -rf /tmp/Rtmp*
+
+### Conda setup
+# Install miniconda
+RUN wget \
+    https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh \
+    && mkdir /root/.conda \
+    && bash Miniconda3-latest-Linux-x86_64.sh -b \
+    && rm -f Miniconda3-latest-Linux-x86_64.sh
+
+COPY environment.yml environment.yml
+
+# restore from conda-lock.yml file and clean up to reduce image size
+RUN conda env create -f environment.yml -n ${ENV_NAME} && \
+  conda clean --all --yes
+
+# Activate conda environment on bash launch
+RUN conda init
+RUN echo "conda activate ${ENV_NAME}" >> ~/.bashrc
+
+# Set entrypoint to bash to activate the environment for any commands
+ENTRYPOINT ["/bin/bash"]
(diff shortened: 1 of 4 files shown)


analyses/cell-type-ewings/Dockerfile
```diff
@@ -1,9 +1,61 @@
...
+
+### Conda setup
+# Install miniconda
+RUN wget \
+    https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh \
+    && mkdir /root/.conda \
+    && bash Miniconda3-latest-Linux-x86_64.sh -b \
+    && rm -f Miniconda3-latest-Linux-x86_64.sh
+
+COPY environment.yml environment.yml
```
jashapiro, line 50: The reason you could not copy the `conda-lock` file in is because it was not in the `.dockerignore` file as an excluded file. Add the following line to that file and you should be able to copy it in (newer templates have this already, but yours did not):

```
!/conda-lock.yml
```

analyses/cell-type-ewings/Dockerfile
```diff
@@ -1,9 +1,61 @@
-# A template docker file for creating a new analysis
-FROM ubuntu:22.04
+# Dockerfile for cell-type-ewings module
+FROM bioconductor/r-ver:devel-R-4.4.0
```
jashapiro, line 2: This is going to end up using Bioc 3.20, so we really don't want to do this...

analyses/cell-type-ewings/Dockerfile
```diff
@@ -1,9 +1,61 @@
...
+LABEL org.opencontainers.image.source="https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-ewings"
 
+# Disable the renv cache to install packages directly into the R library
+ENV RENV_CONFIG_CACHE_ENABLED FALSE
+# set a name for the conda environment
+ARG ENV_NAME=openscpca-cell-type-ewings
 
+# set environment variables to install miniconda
+ENV PATH="/root/miniconda3/bin:${PATH}"
+ARG PATH="/root/miniconda3/bin:${PATH}"
```
jashapiro, line 18: I don't think we should need both of these: ENV sets environment variables within the final image, while ARG only sets them during the build, so having both is redundant. We want the former in this case, and I am going to suggest we follow the pattern of conda install from https://github.com/ContinuumIO/docker-images/blob/main/miniconda3/debian/Dockerfile, so change this to: 
```suggestion
ENV PATH="/opt/conda/bin:${PATH}"
```

analyses/cell-type-ewings/Dockerfile
```diff
@@ -1,9 +1,61 @@
...
+  rm -rf /tmp/downloaded_packages && \
+  rm -rf /tmp/Rtmp*
+
+### Conda setup
+# Install miniconda
+RUN wget \
+    https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh \
+    && mkdir /root/.conda \
+    && bash Miniconda3-latest-Linux-x86_64.sh -b \
+    && rm -f Miniconda3-latest-Linux-x86_64.sh
```
jashapiro, line 48: It is better to have all of the "static" installation steps at the top, and only the part that changes toward the end, for efficient rebuilding. So I would move this to right after the `apt install` step.

I'm also going to suggest a few more steps to make this match the miniconda docker image:

```suggestion
# Install miniconda
# adapted from https://github.com/ContinuumIO/docker-images/blob/main/miniconda3/debian/Dockerfile
RUN curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh \
  && bash Miniconda3-latest-Linux-x86_64.sh -b -p /opt/conda \
  && rm -f Miniconda3-latest-Linux-x86_64.sh \
  && ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh  \
  && echo ". /opt/conda/etc/profile.d/conda.sh" >> ~/.bashrc  \
  && find /opt/conda/ -follow -type f -name '*.a' -delete \
  && find /opt/conda/ -follow -type f -name '*.js.map' -delete

# add conda-lock
RUN conda install conda-lock
```

analyses/cell-type-ewings/Dockerfile
```diff
@@ -1,9 +1,61 @@
...
+
+COPY environment.yml environment.yml
+
+# restore from conda-lock.yml file and clean up to reduce image size
+RUN conda env create -f environment.yml -n ${ENV_NAME} && \
+  conda clean --all --yes
+
+# Activate conda environment on bash launch
+RUN conda init
```
jashapiro, line 57: This line is made redundant by the bash additions I suggested above

jashapiro (commented):
I did some testing locally, and it looks like the error is actually coming from the installation of `Rhtslib`, which may have a corrupted tar file on a server somewhere. At least that is my current guess. I was able to get the Docker build to work with 3.19 by installing `Rhtslib` separately with `RUN Rscript -e 'BiocManager::install("Rhtslib")'` on a line before the `renv::restore()`. I would expect this to be a temporary fix, but it won't hurt but for a bit of time if it stays around. I would leave a comment with it though!

My other main comment is that you want to reorganize the Dockerfile a bit. In general, we want the parts that are less likely to change to be first because that makes for more efficient caching, so my recommendation is to follow the following general order:

- `apt` installations
- install miniconda and conda-lock (see comments about following the installation steps from the miniconda docker repo)
- install `renv`
- copy `conda-lock.yml` and `conda-lock install`
- copy `renv.lock` and `renv::restore`

PR #472, opened 2024-05-30
Add workflow files to module creation docs

### LINKED ISSUE #465: Update analysis creation docs to account for workflows
### What is the documentation improvement or update you wish to see?

In #464, we're updating `create-analysis-module.py` to also add two module workflows: one to build the module dockerfile, and one to run the module. We'll want to update docs to reflect these changes.

Specifically, we'll want to update https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/contributing-to-analyses/analysis-modules/creating-a-module.md to: 
- explain that these workflows get created, but that they are inactive which is expected
- explain that they should be added/committed when establishing the module (maybe also talk about this in commit docs?)

### Is there any additional context you would like to provide?

_No response_

### If this request is for an existing documentation page, please provide the link here.

_No response_

sjspielman:
Closes #465 

This PR adds a small section of documentation explaining that the module creation script now also makes some workflow files. I tried to keep it pretty surface level and not get into the weeds too much, but please let me know if you think more detail would be better!

### DIFF
--- a/docs/contributing-to-analyses/analysis-modules/creating-a-module.md
+++ b/docs/contributing-to-analyses/analysis-modules/creating-a-module.md
@@ -56,8 +56,18 @@ To run this script, take these steps:
     ./create-analysis-module.py --help
     ```
 
-## Module creation script flags
+### Module workflows
 
+The `create-analysis-module.py` script will also create two additional files besides your analysis module.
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
This looks fine to me, but we might want to mention that these are disabled by default?

PR #473, opened 2024-05-31
Add more samples to metadata sheet used for annotating tumor cells in Ewing's

### LINKED ISSUE #437: Run tumor cell annotation workflow for Ewing's samples on more samples

#292 

This issue is just to track running the completed workflow for annotating tumor cells in Ewing's on more samples. We will want to run on 2-3 more samples and then evaluate the results. 

I may add a small script that runs the workflow on multiple samples just so we can track it, but the PR should not contain much else. All new results will be added to a results bucket on S3. 

_No response_

_No response_

_No response_

allyhawkins:

Closes #437 

Here I'm adding additional samples to the sample metadata sheet that have been run through the tumor annotation workflow. This includes picking the cells that should be treated as normal cells for the reference. I used the cell type reports for these samples to identify the cells that I felt confident were normal cells. 

I added the new samples to the metadata sheet and then ran the workflow using the information in the metadata sheet to specify normal/tumor cells for each sample. 

- SCPCS000491 - The "normal" cell types in SingleR were spread out all over the place, but CellAssign had identified a group of T cells. I ended up using that as the reference because there really wasn't a great option. Note that this ended up being a PDX sample, so I actually expect most cells to be tumor cells. 
- SCPCS000492 - The only group of normal cells identified by SingleR that seemed reliable was the epithelial cells, but that same group of cells are classified as trophoblasts in CellAssign. I originally used these as the normal cells along with the Endothelial cells and Fibroblasts identified by CellAssign. But I noticed that CopyKAT without a reference called those cells as aneuploid. I was worried that if they are getting called as trophoblasts they are stem cell like and could actually be tumor cells. So then I removed those from the normal cell reference and just used the Fibroblasts and Endothelial cells classified by CellAssign. 
- SCPCS000493 - Here I am using the Osteoblasts and Fibroblasts identified by CellAssign. These were the only group of cells I was confident was normal given CellAssign classified almost everything else as a T cell and SingleR classified most everything as a mix of muscle/neuron cell types. 

I could make a shell script that runs the workflow using each of these samples, but didn't think that was totally necessary at this point? Curious what others think. 

Yes 

Results from running the annotation workflow which includes classifications based on marker genes, CellAssign, CopyKAT, and InferCNV along with rendered reports summarizing the findings. 

For easy reviewing, here are the zip files of the reports for each of the samples that I ran: 
[SCPCS000491.zip](https://github.com/user-attachments/files/15516268/SCPCS000491.zip)
[SCPCS000492.zip](https://github.com/user-attachments/files/15516269/SCPCS000492.zip)
[SCPCS000494.zip](https://github.com/user-attachments/files/15516271/SCPCS000494.zip)

- SCPCS000492 was the only sample that I ran that had variable expression of a subset of marker genes. The other two seemed to have pretty poor expression of most of the genes and there wasn't a clear group of cells that appeared to express these genes. This doesn't seem great, since most of these samples should be tumor cells so I would expect more consistent expression of these marker genes. I don't think we only want to rely on marker genes to call tumor cells. 
- CellAssign is generally pretty bad with these limited references. Either all cells are assigned as "other" or one of the normal cell types. There weren't very many instances of cells being assigned as tumor with any of the references we used. All three samples CellAssign looks pretty bad and I don't think I would use those results to help with assigning tumor cells. 
- There still doesn't seem to be a difference between using CopyKAT with or without a reference. 
- In SCPCS000492 specifically, cells classified as aneuploid by CopyKAT have a chr16 CNV (which is one of the chrs affected in Ewings) and cells we classify as tumor in InferCNV show a higher proportion of CNV in Chr16 compared to those classified as normal. Seeing consensus between both methods gives me some confidence that maybe these are in fact tumor cells.  
- The cells that have high CNV/ aneuploid using either CNV method are opposite of the cells that are identified to have high tumor marker gene expression for SCPCS000492. This is the most confusing part and makes it hard to identify which are actually tumor cells. I tend to trust the consensus of both CopyKAT and InferCNV over just our marker gene analysis? 
- In SCPCS000494, CopyKAT identified most cells as "not defined", but in cells that are defined as both aneuploid and diploid have a large increase in mean CNV detection for Chr8 as compared to reference. 
- In that same sample, CopyKAT identifies a very clear gain in Chr8 in all cells compared to the reference. However, we are using a cutoff of mean CNV proportion being greater than the mean of all CNV proportions and I think maybe we are being too strict. It looks like even "normal" cells have that gain in Chr8. 

Based on these findings, I think we are going to have to do a little more work before we can confidently identify tumor cells in these samples. For next steps I would like to start with one of these samples and create an exploratory notebook that attempts to identify a group of tumor cells based on the results from marker genes, CopyKAT, and InferCNV. Ultimately, I think we will want to use results from a combination of these methods to solidify the population of tumor cells. In particular, I would like to look at: 

- The consensus between the two CNV methods. How much do these methods actually overlap? If we take the intersection or the union of cells classified as tumor, do we see CNVs in the expected chromosomes. 
	- How does this intersection or union change as we change the cutoffs for classifying cells with InferCNV results. 
- Comparison between cells labeled as tumor by CNV methods and marker genes. Specifically, what is the expression of each marker gene in aneuploid vs diploid cells. Are there specific marker genes that show high expression in cells with high CNVs. 
- How does CNV presence compare with expression of the EWS-FLI1 signature as a whole and not just the marker genes that we pulled out. This was noted in both papers and I think is probably something we might want to incorporate. I would think it could help us validate whether or not the cells identified as tumor cells using InferCNV/CopyKAT are in fact expressing the expected gene signature. 

I plan on filing an issue(s) to track this, but if other people have ideas I am definitely open to them! 

### Author checklists

_Check all those that apply._


allyhawkins:
> > The cells that have high CNV/ aneuploid using either CNV method are opposite of the cells that are identified to have high tumor marker gene expression for SCPCS000492. This is the most confusing part and makes it hard to identify which are actually tumor cells. I tend to trust the consensus of both CopyKAT and InferCNV over just our marker gene analysis?
> 
> As I start to review this, I wonder if the direction of the aneuploidy you see here is as expected? I can imagine that either method might make the wrong "baseline" assumption for expression, which would result in either method thinking it was seeing loss rather than gain. That could explain the discrepancy between methods, and make the marker gene expression the more reliable indicator.

I was definitely worried about this and tried two different sets of "normal" cells as the reference and still came to the same conclusion. We are seeing CNVs in Chr8, Chr12, and/or Chr16 so I do think it's as expected.

jashapiro (commented):
I don't have any problems with the choices that you made here for the sample sheet.

I am generally also okay with not yet using a script to run all samples. If you were to make a wrapper script though, I would probably use R or Python, just because parsing the TSV will be much more straightforward and robust. That said, the "right" thing to do at that point would probably be to use a workflow to handle the whole thing, and I don't think now is the right time to do that.

I do think it makes sense to put the reports directly into the repo, just to make it easier for somebody coming to this module directly. Currently the files are not too big, but the images are probably the biggest contributor to size. You might see if setting chunk options to use JPEG helps size while preserving sufficient quality for these reports. I think you should be able to add the following early in the template docs setup to do it:

```
knitr::opts_chunk$set(
  dev = "jpeg"
)
```

jashapiro:
> I was definitely worried about this and tried two different sets of "normal" cells as the reference and still came to the same conclusion. We are seeing CNVs in Chr8, Chr12, and/or Chr16 so I do think it's as expected.

Yes, but the code as I last saw it was counting both amplifications and deletions. So we would still expect to see CNVs in the correct places, as we do, but they might be going in the wrong direction.

allyhawkins:
@jashapiro I went ahead and added the html files here. I also addressed a few minor spelling issues that were remaining (see #486). This should be ready for another look.

.github/components/dictionary.txt
```diff
@@ -30,6 +30,7 @@ CNA
 CNAs
 CNV
 CNVs
+cnvs
```
jashapiro, line 33: Was this a code thing? Wondering why not capitalize the text

analyses/cell-type-ewings/template_notebooks/04-infercnv.Rmd
```diff
@@ -206,7 +206,7 @@ ggplot(cnv_df, aes(x = UMAP1, y= UMAP2, color = has_cnv)) +
   scale_color_viridis_c()
 ```
 
-Those cells in grey (bottom left) are cells used as a reference. 
+Those cells in gray (bottom left) are cells used as a reference.
```
jashapiro, line 209: boo.

analyses/cell-type-ewings/template_notebooks/01-marker-genes.Rmd
```diff
@@ -33,6 +33,10 @@ Cells with a z-score for any marker gene > 0 are classified as tumor cells.
 ```{r}
 project_root <- here::here()
 renv::load(project_root)
+
+knitr::opts_chunk$set(
+  dev = "jpeg"
+)
```
jashapiro, line 39: Just curious how much space savings this gave, if you know?

jashapiro (approved):
LGTM, though I have a couple little spelling comments. (Should we add "grey" to the dictionary?)

analyses/cell-type-ewings/template_notebooks/01-marker-genes.Rmd
allyhawkins: I don't remember... but I don't think it was a huge difference.

allyhawkins:
Just noting that I updated `cnvs` in the text to `CNVs` and then removed it from the dictionary. I also added `grey` to the dictionary.

PR #474, opened 2024-05-31
Actions for building docker images, alternative model

jashapiro:
This PR is an alternative approach to that taken in #459.

Rather than have one workflow that builds and pushes all of the docker images, with separate workflows for each individual module to test the build only, this keeps both build only and build-push workflows together for each module.
To build all modules, we then have a separate workflow that calls all the others.

The logic in this version is substantially simpler, as each workflow determines when it is triggered, so we save the step of having to determine for ourselves which modules have been modified: if the trigger files are modified, the workflow gets triggered.

The one disadvantage is that while the matrix workflow could incorporate up to 256 individual docker images, the workflow job here can only do 20, so we would potentially need multiple workflows.

There is an alternative here though, which I just thought of: we could move the build-push job out to its own workflow, then have the run-all version use that as well, as a matrix job. Assuming that all workflows have the same steps, that ought to work. I'm going to stack that version onto this one.

### DIFF
--- a/.github/workflows/docker_all-modules.yml
+++ b/.github/workflows/docker_all-modules.yml
@@ -1,0 +1,33 @@
+# This is a workflow to all of the analysis module workflows
+#
+# This workflow will run with the following triggers
+# - Manual trigger
+# - Periodic running on a timer
+
+name: Run all analysis workflows
+on:
+  workflow_dispatch:
+    inputs:
+      push-ecr:
+        description: "Push to AWS ECR"
+        type: boolean
+        required: false
+
+jobs:
+  # Modules to build Docker images for
+  # Note: only 20 modules can be run per workflow
+  simulate-sce:
+    uses: ./.github/workflows/docker_simulate-sce.yml
+    with:
+      push-ecr: ${{ inputs.push-ecr }}
+
+  ## Add additional modules above this comment, and to the needs list below
+  check-jobs:
+    if: ${{ always() }}
+    needs:
+      - simulate-sce
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for failures or cancelled jobs
+        if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
+        run: echo "Job failed" && exit 1
--- a/.github/workflows/docker_all-modules.yml
+++ b/.github/workflows/docker_all-modules.yml
@@ -1,10 +1,9 @@
-# This is a workflow to all of the analysis module workflows
+# This is a workflow to build and optionally push all module Docker images
 #
-# This workflow will run with the following triggers
-# - Manual trigger
-# - Periodic running on a timer
+# This workflow will run only on manual triggers
 
-name: Run all analysis workflows
+
+name: Build all Docker images
 on:
   workflow_dispatch:
     inputs:
--- a/.github/workflows/docker_simulate-sce.yml
+++ b/.github/workflows/docker_simulate-sce.yml
@@ -23,15 +23,20 @@ on:
       - analyses/simulate-sce/conda-lock.yml
   workflow_dispatch:
     inputs:
-      push:
+      push-ecr:
         description: "Push to AWS ECR"
         type: boolean
         required: false
+  workflow_call:
+    inputs:
+      push-ecr:
+        description: "Push to AWS ECR"
+        type: boolean
 
 jobs:
   test-build:
     name: Test Build Docker Image
-    if: github.event_name == 'pull_request' || !inputs.push
+    if: github.event_name == 'pull_request' || !inputs.push-ecr
     runs-on: ubuntu-latest
 
     steps:
@@ -48,7 +53,7 @@ jobs:
 
   build-push:
     name: Build and Push Docker Image
-    if: github.event_name == 'push' || inputs.push
+    if: github.event_name == 'push' || inputs.push-ecr
     environment: prod
     runs-on: ubuntu-latest
 
--- a/.github/workflows/docker_simulate-sce.yml
+++ b/.github/workflows/docker_simulate-sce.yml
@@ -1,8 +1,6 @@
 name: Build docker image for simulate-sce
 env:
   MODULE: simulate-sce
-  MODULE_PATH: analyses/simulate-sce
-  REGISTRY_ALIAS: openscpca
 
 on:
   pull_request:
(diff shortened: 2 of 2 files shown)


.github/workflows/docker_all-modules.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  build-docker-modules:
+    name: Build Docker Images
+    uses: ./.github/workflows/build-push-docker-module.yml
+    strategy:
+      fail-fast: false
+      matrix:
+        module:
+          - simulate-sce
```
jaclyn-taroni: So we'd add to this list as we add modules, correct?

.github/workflows/docker_all-modules.yml
```diff
@@ -0,0 +1,36 @@
...
+# This workflow will run only on manual triggers
+
+name: Build all Docker images
+on:
+  workflow_dispatch:
+    inputs:
+      push-ecr:
+        description: "Push to AWS ECR"
+        type: boolean
+        required: false
```
jaclyn-taroni, line 12: Should this also be scheduled?

jaclyn-taroni: If it should be scheduled, we should probably figure out how we'll know about it if it fails!

.github/workflows/docker_simulate-sce.yml
```diff
@@ -23,59 +19,40 @@ on:
...
+      push-ecr:
+        description: "Push to AWS ECR"
+        type: boolean
 
 jobs:
-  build:
-    name: Build Docker Image
+  test-build:
+    name: Test Build Docker Image
+    if: github.event_name == 'pull_request' || (contains(github.event_name, 'workflow_') && !inputs.push-ecr)
```
jaclyn-taroni, line 42: Do I have the right idea: If you're manually calling it or reusing a workflow and push is false, skip straight to `build-push`

.github/workflows/docker_simulate-sce.yml
```diff
@@ -23,59 +19,40 @@ on:
...
+          push: false
           cache-from: type=gha
           cache-to: type=gha,mode=max
+
+  build-push:
+    name: Build and Push Docker Image
+    uses: ./.github/workflows/build-push-docker-module.yml
+    if: github.event_name == 'push' || inputs.push-ecr
+    with:
+      module: simulate-sce
```
jaclyn-taroni, line 62: This field would get populated at template time, yea?

.github/workflows/docker_all-modules.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  build-docker-modules:
+    name: Build Docker Images
+    strategy:
+      fail-fast: false
+      matrix:
+        # add all modules with active docker images here
+        module:
+          - hello-python
+          - simulate-sce
```
jaclyn-taroni, line 23: Here's where we'd keep adding modules, correct? (This was kind of answered by your reorg that wiped out a comment 😄 )

.github/workflows/build-push-docker-module.yml
```diff
@@ -0,0 +1,59 @@
...
+        uses: aws-actions/amazon-ecr-login@v2
+        with:
+          registry-type: public
+
+      - name: Create ECR repository if needed
+        id: create-ecr
+        run: |
+          aws ecr-public describe-repositories --repository-names ${{ inputs.module }} \
+          || aws ecr-public create-repository --repository-name ${{ inputs.module }}
+      - name: Set up Docker Buildx
```
jaclyn-taroni, line 39:
```suggestion
          || aws ecr-public create-repository --repository-name ${{ inputs.module }}
          
      - name: Set up Docker Buildx
```

.github/workflows/docker_simulate-sce.yml
jaclyn-taroni: Looks like yes from what you just pushed!

jaclyn-taroni (commented):
Returning some comments to make sure I'm understanding correctly. 

We should also figure out what kind of interim documentation we need before this goes in (probably stacked on this)

.github/workflows/docker_simulate-sce.yml
jashapiro: yup!

.github/workflows/docker_all-modules.yml
jashapiro: Yes, I added a comment here too because I knew it would be unclear. I'd add another comment at the top to tell people where to go.

.github/workflows/docker_simulate-sce.yml
jashapiro: If you are manual or reuse and `push-ecr` is false, we run this, if `push-ecr` is true, we skip to `build-push`. (I made the variable different because `push` is overloaded)

jashapiro:
I added filing an issue of failure and adjusted some of the scheduled job timings so that the docker build would run before the analysis module runs, and offset the spelling. 

I had a thought about having the docker build trigger analysis builds when it was done, but that would mean we would run into the max number of workflows with no way to override by adding more actions, so I abandoned that idea.

.github/cron-issue-templates/all-docker-issue-template.md
```diff
@@ -0,0 +1,7 @@
+At least one of the docker images that is part of the [Build all Docker images](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/workflows/docker_all-modules.yml) workflow has failed.
```
jaclyn-taroni, line 1:
```suggestion
At least one of the Docker images that is part of the [Build all Docker images](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/workflows/docker_all-modules.yml) workflow has failed.
```

.github/cron-issue-templates/all-docker-issue-template.md
```diff
@@ -0,0 +1,7 @@
...
+
+Check the latest output from the workflow at the following link to see which module failed.
+https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/workflows/docker_all-modules.yml
+
+Then assign an OpenScPCA admin to this issue and add to this issue a list of the modules with failing docker builds.
+Alternatively, file a new issue for each failing docker image.
```
jaclyn-taroni, line 7:
```suggestion
Alternatively, file a new issue for each failing Docker image.
```

.github/workflows/docker_all-modules.yml
```diff
@@ -0,0 +1,67 @@
+# This is a workflow to build and optionally push all module Docker images
+#
+# The action works by use a GitHub matrix to build all images in parallel.
+# The modules that will have their docker images built are specified in the matrix variable `module`.
```
jaclyn-taroni, line 4:
```suggestion
# The modules that will have their Docker images built are specified in the matrix variable `module`.
```

.github/workflows/build-push-docker-module.yml
```diff
@@ -0,0 +1,63 @@
+name: Build and push a docker image for a module
```
jaclyn-taroni, line 1:
```suggestion
name: Build and push a Docker image for a module
```

.github/workflows/docker_simulate-sce.yml
```diff
@@ -1,8 +1,9 @@
 name: Build docker image for simulate-sce
```
jaclyn-taroni, line 1:
```suggestion
name: Build Docker image for simulate-sce
```

jaclyn-taroni (approved):
Returning little copy things, but LGTM otherwise 👍🏻

PR #476, opened 2024-05-31
Use matrix for multiple modules

jashapiro:
Stacked on #474

This PR pulls out the docker build-push action to its own workflow and then calls that from both the individual docker build actions and the all-module action

The one concern I have with this one is whether the github variable info that determines the docker tags  in the build action will be passed on properly from the calling workflows. This should be easy enough to test in practice (as well as by reading the docs carefully), but I _think_ they will? If not, there are workarounds.

### DIFF
--- a/.github/workflows/build-push-docker-module.yml
+++ b/.github/workflows/build-push-docker-module.yml
@@ -1,0 +1,57 @@
+name: Build and push a docker image for a module
+
+on:
+  workflow_call:
+    inputs:
+      module:
+        required: true
+        type: string
+      push-ecr:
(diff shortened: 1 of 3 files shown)


jashapiro:
> The one concern I have with this one is whether the github variable info that determines the docker tags in the build action will be passed on properly from the calling workflows. This should be easy enough to test in practice (as well as by reading the docs carefully), but I _think_ they will? If not, there are workarounds.

I have now checked this (I was not as able to find the answer in docs as I might have hoped) and it did seem to work: when the ultimate trigger was a tag, the tag was passed to the subworkflow.

PR #485, opened 2024-06-03
Fix spelling errors for Ewing's modules

### LINKED ISSUE #482: Monthly spellcheck results
This issue tracks fixing typos identified by scheduled spellcheck workflow.

## Spelling errors

Spellcheck found **124 errors**.

[Click here to download the errors TSV](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/9325750165/artifacts/1557946303).

## Issue checklist

- [x] Assign an OpenScPCA admin
- [ ] Spell check errors have been fixed via a PR
- [ ] Spell check workflow has been manually run to confirm spelling errors are fixed

allyhawkins:
Closes #482 

Here I fixed all the spelling errors I found in the Ewing's module. For any of the ones that affected the exploratory analysis notebooks, I re-rendered them and am including them here. 

One other comment I have is that the results tsv file tells you the filename, but not the relative path to the file within the module (or even what module it is). I feel like that might be something we want to do otherwise these spell checks could get out of hand.


sjspielman:
> One other comment I have is that the results tsv file tells you the filename, but not the relative path to the file within the module (or even what module it is). I feel like that might be something we want to do otherwise these spell checks could get out of hand.

I have been thinking about this too - on it nextdoor in the spellcheck repo!

sjspielman:
@allyhawkins can you merge in `main` changes (now has my spell check fixes)? That way I can review this making use of my local copy of the spell check script. Thanks!

analyses/cell-type-ewings/exploratory_analysis/01-marker-gene-tumor-classification.Rmd
```diff
@@ -46,7 +46,7 @@ theme_set(
 repository_base <- rprojroot::find_root(rprojroot::is_git_root)
 
 # The current data directory, found within the repository base directory
-data_dir <- file.path(repository_base, "data", "current")
+data_dir <- file.path(repository_base, "data", "2024-05-01")
```
sjspielman, line 49: Not spelling related, but noting that if you're going to tie this module to a specific data release then you'll want to document this required version in the module's README.

sjspielman (commented):
Spelling looks good! I only had one non-spelling comment here about docs for the data release version, so you should either:
- change the dates (in a couple spots) back to `current` in this PR, and do a second PR where you nail down the data version
- document that change in the README in this PR (it's a small enough change that I think it's fine)

analyses/cell-type-ewings/exploratory_analysis/01-marker-gene-tumor-classification.Rmd
jashapiro: I think we do not want to do this in general. The idea is that modules should almost always be expected to work with the current data release. In the specific case of exploratory analysis, tagging to a release might make sense, but I would rather it _output_ the release info rather than take it as hard-coded input.

allyhawkins: So the reason I had to do this is because I have downloaded both actual data and test data. Because I downloaded the test data second then the `current` folder symlinks to the test data. This update is _only_ in the exploratory notebooks because I needed to be able to regenerate the notebooks regardless of any other changes that have been made to symlinks. Because of that, I actually do want to tie this to a release.

allyhawkins: Note that this is not for the actual workflow itself. That can still take any release.

jashapiro: In that case, I think a note in the exploratory notebook directory or in the notebook itself would be the way to go. I would not put this requirement in the main module readme.

allyhawkins:
> Just checking, was this change meant to be implemented in this branch? I may not have the context I need to review the change - [8242770](https://github.com/AlexsLemonade/OpenScPCA-analysis/commit/824277089da2f0f533f3eb4f203fa64d8e0106dc)

Yes, I'm trying to figure out why checks are failing.

allyhawkins:
> In that case, I think a note in the exploratory notebook directory or in the notebook itself would be the way to go. I would not put this requirement in the main module readme.

Noting that I added a small note about this in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/485/commits/d5db0f146416c1556f21b99aa1e100cc722eaec2. 

I also had to make a small change to pass GHA. Basically we want to skip anything in the notebooks that fail for the test data in the workflow. One thing that tends to fail is comparing cell type annotations using a confusion matrix for the CellAssign runs. There are times where we only get one cell type from CellAssign and not all cell types are found, so the confusion matrix can't be calculated. The change I added just checks that all expected cell types are there before doing the confusion matrix.

sjspielman (approved):
This should do it!

PR #499, opened 2024-06-04
Notebook comparing doublet results across methods

sjspielman:

#446

This PR explores the overlap for each method's doublet calls for each dataset, and also assesses performance of a "consensus caller."

I wrote a single notebook to process all datasets with three main analysis sections, in addition to a conclusions section at the end:

1. Upset plot comparing doublet calls
2. PCA colored by consensus calls
3. Confusion matrix and associated metric calculations

I also updated the overall module run script to render this notebook as the next step.

There are a few other changes here:
- Spelling
- Fixed a bug in `template-notebooks/02_explore-benchmark-results.Rmd` where I was using the wrong variable in some functions

Yep.

`researcher-654654257431-us-east-2`

There are no additional result files, only the rendered notebook which contains all results from this analysis. I directly committed this notebook to the directory where I saved it in the module. Is this ok, or should I export it to results?

Doublet calls are not much in agreement, so consensus calls are small sets. The consensus calls do not appear to be the most accurate, either.

`renv` environment needed to render this notebook

I suppose I could add more analysis or interpretation to the notebook, but as there really isn't "much of a there there" to these results, as it were, I wasn't sure what else might be useful and informative to include. Do you have any ideas?

-

### Author checklists

_Check all those that apply._


analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,363 @@
...
+
+1. Do methods tend to predict overlapping or distinct sets of doublets on the same dataset?
+2. Is the consensus doublet call across methods predictive of the true doublet status?
+Here, the "consensus doublets" are those droplets which all methods identify as doublets.
+
+There are several items to bear in mind when interpreting these results:
+
+- The ground truth calls themselves to which we are comparing consensus calls may not be entirely accurate, since they were also computationally identified generally with demultiplexing algorithms. 
+- `cxds`, as we are using it, does not have a specific threshold for calling droplets. 
+By contrast, both `scublet` and `scDblFinder` identify a threshold based on the given dataset they are processing.
```
allyhawkins, line 22:
```suggestion
By contrast, both `scrublet` and `scDblFinder` identify a threshold based on the given dataset they are processing. 
```

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,363 @@
...
+```{r read_data}
+# used in PCA plots
+confusion_colors <- c(
+  "tp" = "lightblue",
+  "tn" = "pink",
+  "fp" = "blue",
+  "fn" = "firebrick2"
+)
+
+# Read in and data for analysis
```
allyhawkins, line 125:
```suggestion
# Read in data for analysis
```

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,363 @@
...
+      
+      dataset_df <- scdbl_df |>
+        dplyr::left_join(
+          scrub_df, 
+          by = "barcodes"
+        ) |>
+        dplyr::left_join(
+          sce_df, 
+          by = "barcodes"
+        )
```
allyhawkins, line 171:
```suggestion
                      PC2 = PCA.2) |> 
        dplyr::left_join(
          scrub_df, 
          by = "barcodes"
        ) |>
        dplyr::left_join(
          scdbl_df, 
          by = "barcodes"
        ) 
```
Just simplifying this a little bit.

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,363 @@
...
+            call_type == "tn" ~ unname(confusion_colors["tn"]),
+            call_type == "fp" ~ unname(confusion_colors["fp"]),
+            call_type == "fn" ~ unname(confusion_colors["fn"])
+          )
+          )
+      
+      return(dataset_df)
+    }
+  )
+names(doublet_df_list) <- dataset_names
```
allyhawkins, line 202:
```suggestion
  ) |> 
  purrr::set_names(dataset_names)
```

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,363 @@
...
+
+## Upset plots
+
+This section contains upset plots that show overlap across doublet calls from each method, displayed for each dataset.
+
+```{r}
+pull_barcodes <- function(df, pred_var) {
+  # Helper function to pull out barcodes for doublet calls
+  df$barcodes[df[[pred_var]] == "doublet"]
+}
```
allyhawkins, line 216: I don't know that having a function here is totally necessary since it's just one line

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,363 @@
...
+
+```{r, fig.width = 12, fig.height = 6}
+# Make a legend for the confusion-colored PCA
+legend_plot <- data.frame(
+  x = factor(names(confusion_colors), levels = names(confusion_colors)), y = 1:4
+) |>
+ ggplot(aes(x = x, y = y, color = x)) + 
+  geom_point(size = 3) + 
+  scale_color_manual(name = "Metric", values = confusion_colors) 
+confusion_legend <- ggpubr::get_legend(legend_plot) |> ggpubr::as_ggplot()
```
allyhawkins, line 267: Why do you need to do this separately and not within the function for creating the confusion plot? Also why do you need to do this at all... the legend should automatically match the colors that you specify which you can do with `scale_color_manual` and a named list of values.

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,363 @@
...
+    }
+  ) |>
+  dplyr::bind_rows()
+```
+
+
+
+## Conclusions
+
+Overall, methods do not have substantial overlap with out another.
```
allyhawkins, line 342:
```suggestion
Overall, methods do not have substantial overlap with each other. 
```

allyhawkins (commented):
Generally I think the majority of the content you have here looks good. I left a few comments throughout the notebook, but also have some overall comments: 

- The score for cxds that you are using here seems very stringent to me. I think I would lower to 0.5 and be more lenient here. If you are then looking at the consensus between the three methods then you are going to be even more stringent, so I think it's okay to be less stringent on the individual method's threshold. 
- It looks like you are getting a fair amount of false negatives, which means we are potentially missing doublets by using the consensus. I think this might be improved by just using two methods rather than all three. I would try with just `scDblFinder` and `scrublet` and see if that decreases your false negative number. 
- One thing that I think might be helpful is actually coloring a PCA based on which method calls each cell as doublet... I would make everything that's a singlet grayed out and then doublets that are only in `scDblFinder` one color, only in `scrublet` one color, only in `cxds` one color and then overlapping ones different colors... You might break this up so that you have one PCA for `scDblFinder` that shows cells identified as doublets in that method vs. cells overlapping with other methods. It could be total chaos and not informative, but I think it's helpful to see where you are losing the false negatives. I'm curious if the false negatives are because those are doublets found in `scDblFinder` and `scrublet`, but not in `cxds`.

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
sjspielman: Hm.... I did this because not all the plots had all of the four colors, but using a _named list_ never crossed my mind........................... That is of course better!

sjspielman:
We're back! I updated code throughout notebook in response to reviews, including using a `0.5` threshold for `cxds` and adding more PCAs. While looking at the PCAs, it actually looked to me like it was _`cxds`_ that was capturing a lot of the consensus false negative droplets, and those points were missed by `scDblFinder` and `scrublet`. Therefore, for this first round returning back to you, I didn't do a re-analysis with just those two methods. Do you still think it's worth doing?

allyhawkins (commented):
Thanks for updating this. It does look like the most conservative of the doublet callers is `scrublet` with `scDblFinder` or `cxds` calling the most doublets. Although this appears different than previous results where `cxds` was the most conservative because we changed the threshold. I think it's worth noting that changing that threshold can change the results of what gets called pretty drastically. 

I do think you are creating a lot of false negatives by using the consensus between all three vs. just `scDblFinder` and `cxds`. So I don't think it hurts to look at what happens when you just take the consensus between those two. I would expect the kappa scores to improve for all samples but the hm-6k. I would probably add this to the existing analysis and plots here. 

The other comment I have is about labeling the points that do not have consistent calls across all methods rather than just labeling everything as singlet. I do think it would be helpful to know which points are called differently by each method. Specifically I want to know how often we are throwing out true doublets that are identified in maybe 2/3 of methods but are not identified in all three methods? And then how often are the ambiguous calls actually consistent with being a singlet.

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,369 @@
...
+Further, the consensus doublets called by all three methods have some, but not substantial, overlap with the ground truth. 
+
+For three out of four datasets, `scDblFinder` predicts a much larger number of doublets compared to other methods. 
+For `pdx-MULTI`, however, `cxds` predicts a much larger number of droplets.
+
+Based on the PCAs, it additionally looks like there are many more false negatives in the consensus prediction that `cxds` appears to capture but other methods do not.
+
+The table below summarizes performance of the "consensus caller". 
+Note that, in the [benchmarking paper these datasets were originally analyzed in](https://doi.org/10.1016/j.cels.2020.11.008), `hm-6k` was observed to be one of the "easiest" datasets to classify across methods.  
+Consistent with that observation, it has the highest `kappa` value here, although it is still fairly low - though not as low as the other methods, which are very close to 0.
```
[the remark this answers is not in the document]
sjspielman: Ahhh I was getting all confused at how these numbers could have changed, but I clearly had a gap in work between changing the threshold to `0.5` and re-checking these numbers! 🔍

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,369 @@
...
+      
+      # Add a consensus call column
+      dataset_df <- dataset_df |>
+        dplyr::rowwise() |>
+        dplyr::mutate(consensus_call = dplyr::if_else(
+          all(
+            c(scdbl_prediction, scrublet_prediction, cxds_prediction) == "doublet"
+          ),
+          "doublet", 
+          "singlet"
```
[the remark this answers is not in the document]
sjspielman: > the addition of an "ambiguous" class.

Yeah, this seems interesting to look at - I'll define anything as ambiguous which has any disagreement, and update a lot of the code to accommodate this. But, I want to keep the fn/fp/tn/tp code for the `caret` calculations, so I will tack this on separately.

sjspielman:
The next iteration has finally landed! I've incorporated the conceptual items brought up in review, and did some notebook rearrangement accordingly. Note that I do think this could be more modular since there is some repeated code between different types of consensus analyses (all 3 methods vs only 2 methods), but given where we anticipate this module headed overall, I wasn't sure that was really worth the effort.

Here's a rendered notebook:
[03_compare-benchmark-results.nb.html.zip](https://github.com/user-attachments/files/16019113/03_compare-benchmark-results.nb.html.zip)

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,522 @@
...
+            plot_pca_calls(
+              df, 
+              color_column = !!sym(color_column), 
+              plot_type = "calls",
+              title = plot_title
+            ) + theme(legend.position = "none")}
+        ) 
+      # plot for consensus_class, which needs three colors
+      # note that the list name isn't actually used here
+      pc_plot_list$consensus_class <- plot_pca_calls(
```
allyhawkins, line 290: Can we add back the tp/fp/tn/fn plot too? I know that means there will be 7 plots but I think that was helpful along with the ambiguous plot.

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,522 @@
...
+         ) |>
+        # Add Update `confusion_call`
+        dplyr::mutate(
+          confusion_call = dplyr::case_when(
+            consensus_call == "doublet" && ground_truth == "doublet" ~ "tp",
+            consensus_call == "singlet" && ground_truth == "singlet" ~ "tn",
+            consensus_call == "doublet" && ground_truth == "singlet" ~ "fp",
+            consensus_call == "singlet" && ground_truth == "doublet" ~ "fn"
+          )
+        )
```
allyhawkins, line 404: I think I would pull this out to a function that you have at the top of the notebook and include an argument for `consensus_columns` or something like that where you can specify the vector of columns to use for consensus.

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,522 @@
...
+        color_column = consensus_class,
+        plot_type = "class",
+        title = "Consensus class"
+      )
+
+      patchwork::wrap_plots(pc_plot_list) +
+        plot_annotation(
+            glue::glue("PCA for {dataset}"), 
+            theme = theme(plot.title = element_text(size = 16))) + 
+        plot_layout(guides = "collect")
```
allyhawkins, line 457: I think same thing, just pull this out to a function at the top of the notebook. It can be really simple and just take the `pc_color_columns` and the `df` as input.

allyhawkins (commented):
In terms of analysis I think this is good for now. We should discuss what we're going to do next from here, if anything at this point. Yes, they perform well on the easy sample, but they don't do a good job on the hard samples, which are more realistic. I think `scDblFinder` and `cxds` are 50/50 on being the most consistent with the ground truth, but when you look at them together you still miss a lot of calls. So I think we need to think about how we want to use that information. One thought I had is that this module is an optional analysis module that can be used to run doublet detection using three different methods, but that's it. So contributors can have it if that's something they feel is necessary for their analysis, but we don't go beyond that. 

I do think this could benefit from moving some of the code to functions and then adding back in the tp/fp/tn/fn colored PCA plot. But other than that I don't think we need any other changes at this stage.

sjspielman:
> One thought I had is that this module is an optional analysis module that can be used to run doublet detection using three different methods, but that's it. So contributors can have it if that's something they feel is necessary for their analysis, but we don't go beyond that.

This seems pretty reasonable to me actually, for the end-goal of the module to be a utility for folks to run these three methods on an SCE, which would include associated results and metadata.

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
sjspielman: Noting we discussed off github that I'm not going to port this into a function.

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
sjspielman: Noting we discussed off github that I'm not going to port this into a function.

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,522 @@
...
+4. Ground truth single/doublets
+5. Consensus _call_ droplets are either "doublet" or "singlet"
+6. Consensus _class_: droplets are either "doublet", "singlet", or "ambiguous"
+
+- For the first five panels, doublets are shown in black and singlets in light gray.
+- In the sixth panel showing the consensus class, doublets are shown in black, singlets in light gray, and ambiguous are blue.
+
+```{r, fig.width = 12, fig.height = 6}
+doublet_df_list |>
+  purrr::imap(
```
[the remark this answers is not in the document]
sjspielman: I wish, but sadly no. Patchwork won't print the plots unless I do `map`, so we're stuck with the extra printing...

sjspielman:
Plots restored!
[03_compare-benchmark-results.nb.html.zip](https://github.com/user-attachments/files/16073598/03_compare-benchmark-results.nb.html.zip)

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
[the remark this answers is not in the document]
allyhawkins: I've had the same issue before when `iwalk` doesn't work 😢

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
```diff
@@ -0,0 +1,575 @@
...
+
+
+
+### Conclusions
+
+Overall, methods do not have substantial overlap with each other. 
+They each tend to detect different sets of doublets, leading to fairly small sets of consensus doublets. 
+Further, the consensus doublets called by all three methods have some, but not substantial, overlap with the ground truth. 
+
+For three out of four datasets, `scDblFinder` predicts a much larger number of doublets compared to other methods.
```
allyhawkins, line 398:
```suggestion
For two out of four datasets, `scDblFinder` predicts a much larger number of doublets compared to other methods. 
```

allyhawkins (approved):
LGTM!

PR #500, opened 2024-06-04
Validate tumor cell annotations for SCPCS000490

### LINKED ISSUE #479: Annotate tumor cells in SCPCS000490

#292 

The first step of annotating cell types in the Ewing's samples is annotating tumor cells. As of now, we have utilized multiple methods to obtain annotations - marker gene analysis, CellAssign, CopyKAT, and InferCNV. Now that we have all this data on each sample we will want to create final tumor cell annotations for each sample. The goal is to confidently annotate a handful of samples and then use those as a reference to annotate the remaining samples. This issue is to track creating a TSV file annotating each cell as normal or tumor in SCPCS000490 using the combination of results from running different annotation methods. 

This PR will contain a notebook that reads in the results from marker gene analysis, CopyKAT, InferCNV, and gene set scoring (see #478) for SCPCS000490 and uses those results to come up with a final combined annotation of tumor vs. normal. 

- We will look at the consensus between tumor and normal cells identified by CopyKAT and InferCNV. We will validate these cells as tumor by looking at expected chromosome gains. Based on our initial analysis, this sample has a gain in Chr8, so we expect that tumor cells will show that in comparison to normal cells. 
- We will look at marker gene expression for each marker gene in cells that are identified as tumor vs. normal using CNV methods. 
- Additionally, we will look at gene expression of the EWS-FLI1 signature in tumor and normal cells to validate which cells should be called as tumor. 
- Final annotations will then be saved as a TSV file with two columns - barcode and tumor annotation (Tumor or Normal). 

_No response_

_No response_

_No response_

allyhawkins:

Closes #479 

Before finalizing our annotations of tumor cells, we want to explore the results from marker genes, InferCNV, and CopyKAT and validate these findings. Here I'm adding a notebook that is specific to SCPCS000490 to identify and validate the tumor cell annotations. 

I'm also adding a TSV file with the annotations for this sample. Part of this is making a folder to store these cell types and the notebooks used to validate the annotations. I chose to put these in the same `cell_type_annotations` folder rather than the exploratory notebooks since this notebook is specific to validating tumor cell annotations. I envision doing the same thing for a few more samples so will make a new folder for each sample. 

- The first thing I did is just read in the results from the workflow for annotating tumor cells. This workflow produces results from each method, but does not combine them, so the first thing I did was look at cells that were identified as tumor cells by multiple methods. This is the "easy" sample, so we see a pretty good consensus across methods. I chose to use any cells labeled as tumor by both InferCNV and CopyKAT as tumor. 
- The rest of the notebook is focused on validation: 
  - I look at marker gene expression and see that it is higher in tumor cells. 
  - I used the gene set scores that I calculated and also see that it's higher in tumor cells. 
  - I then looked for specific gains and losses using the InferCNV output. 

Although this is the easier of the samples, I did create some new plots that I think will be helpful in validating tumor cells in other samples. I might have gone overkill with the heatmaps, but I actually think it's helpful to see which cells have expression of each gene/ gene set/ CNV. This should help us identify any cells that we may be categorizing incorrectly in future samples. 

Another note is that I expect these notebooks to be tailored very specifically to each sample. This one is mostly validation rather than looking at different classification cutoffs since it was pretty obvious to me, but I envision future notebooks will have some more exploration of different cutoffs. 

Yes 

There is a TSV file with the final annotations that right now only includes `tumor_cell_classification`. My plan is to build on this as we annotate normal cells and tumor cell states. 

Here's a zip of the rendered notebook for easy review: 
[SCPCL000822_tumor-cell-validation.html.zip](https://github.com/user-attachments/files/15570507/SCPCL000822_tumor-cell-validation.html.zip)

Nothing special since it's just rendering a notebook using previously produced results. 


analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+suppressPackageStartupMessages({
+  # load required packages
+  library(SingleCellExperiment)
+  library(ggplot2)
+})
+
+# Set default ggplot theme
+theme_set(
+  theme_bw()
+)
```
jashapiro, line 35: Do we maybe also want to set this to use jpgs?

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+final_annotations_file <- file.path(final_annotations_dir, "SCPCL000822_tumor-classifications.tsv")
+```
+
+
+```{r}
+# read in all files 
+sce <- readr::read_rds(sce_file)
+
+# read in marker genes table 
+marker_genes_df <- readr::read_tsv(marker_genes) |>
```
jashapiro, line 79: A minor comment, but do we maybe want to throw in `show_col_types = FALSE`  or `col_types = cols(.default = "c"))` just to quiet the `readr` messages.

Or do it once with `options(readr.show_col_types = FALSE)` in setup?

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+
+# copykat output
+copykat_results_df <- readr::read_tsv(copykat_predictions_file) |> 
+  # only select no_ref classification and rename to be consistent with naming of other classifications
+  dplyr::select(barcodes, copykat_classification = no_ref) |> 
+  dplyr::mutate(
+    copykat_classification = dplyr::case_when(
+      copykat_classification == "aneuploid" ~ "Tumor",
+      copykat_classification == "diploid" ~ "Normal",
+      TRUE ~ copykat_classification
```
jashapiro, line 94: Newer syntax is a bit more readable
```suggestion
      .default = copykat_classification
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+  dplyr::select(barcodes, UMAP1, UMAP2, gene_symbol, ensembl_gene_id, gene_expression) |> 
+  # add in marker gene classifications
+  dplyr::left_join(marker_gene_results_df) |> 
+  # add in copykat results 
+  dplyr::left_join(copykat_results_df) |> 
+  # add in infercnv results and metadata
+  dplyr::left_join(infercnv_results_df) |> 
+  dplyr::left_join(infercnv_metadata_df) |> 
+  # gene set scores 
+  dplyr::left_join(geneset_scores_df)
```
jashapiro, line 152: Better to be explicit?
```suggestion
  dplyr::left_join(marker_gene_results_df, by = "barcodes") |> 
  # add in copykat results 
  dplyr::left_join(copykat_results_df, by = "barcodes") |> 
  # add in infercnv results and metadata
  dplyr::left_join(infercnv_results_df, by = "barcodes") |> 
  dplyr::left_join(infercnv_metadata_df, by = "barcodes") |> 
  # gene set scores 
  dplyr::left_join(geneset_scores_df, by = "barcodes")
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+  # add in marker gene expression to dataframe 
+  dplyr::left_join(marker_gene_exp, by = "barcodes") |> 
+  # combine all genes into a single column for easy faceting
+  tidyr::pivot_longer(
+    cols = starts_with("ENSG"),
+    names_to = "ensembl_gene_id",
+    values_to = "gene_expression"
+  ) |> 
+  # join with marker gene df to get gene symbols for plotting 
+  dplyr::left_join(marker_genes_df, by = c("ensembl_gene_id")) |>
```
jashapiro, line 141:
```suggestion
  dplyr::left_join(marker_genes_df, by = "ensembl_gene_id") |>
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+  dplyr::select(barcodes, UMAP1, UMAP2, marker_gene_classification, copykat_classification, cnv_proportion_classification) |> 
+  tidyr::pivot_longer(
+    cols = ends_with("classification"),
+    names_to = "method",
+    values_to = "classification"
+  )
+
+ggplot(tumor_umap_df, aes(x = UMAP1, UMAP2, color = classification)) +
+  geom_point(size = 0.5, alpha = 0.5) +
+  facet_wrap(vars(method))
```
jashapiro, line 172: I noted in this plot that the copykat classification does not have a reference class. Is this from the no-reference data?

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+  dplyr::left_join(infercnv_results_df) |> 
+  dplyr::left_join(infercnv_metadata_df) |> 
+  # gene set scores 
+  dplyr::left_join(geneset_scores_df)
+
+```
+
+## Identify tumor cell population 
+
+Below is a UMAP showing which cells are labeled as tumor cells using each of the various classification methods.
```
jashapiro, line 158: I think a reiteration of what the classification methods are would be useful here. I don't know where the cnv_proportion came from, which copyKAT method we are using here, etc. (From later context I know that the first is inferCNV, but it took me a while to get there)

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+    names_to = "method",
+    values_to = "classification"
+  )
+
+ggplot(tumor_umap_df, aes(x = UMAP1, UMAP2, color = classification)) +
+  geom_point(size = 0.5, alpha = 0.5) +
+  facet_wrap(vars(method))
+```
+
+This plot shows us that there is a pretty clear consensus that most of the cells in the bottom left group are tumor cells.
```
jashapiro, line 175: bottom right?

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+# cells that are classified as tumor by both are kept as tumor, all other are normal 
+coldata_df <- coldata_df |> 
+  dplyr::mutate(
+    combined_cnv_classification = dplyr::if_else(
+      cnv_proportion_classification == "Tumor" & copykat_classification == "Tumor",
+      "Tumor",
+      "Normal"
+    ),
+    # there are cells classified as NA by copykat, we want those to be normal 
+    combined_cnv_classification = dplyr::if_else(is.na(copykat_classification) & is.na(combined_cnv_classification), "Normal", combined_cnv_classification)
```
jashapiro, line 190: I'm a bit confused by this being done in two steps. Can you use `case_when()` to do all the cases in a single call?

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+  )
+```
+
+```{r}
+# look at which cells are tumor/normal
+# be sure to remove any duplicates 
+coldata_df |> 
+  dplyr::select(UMAP1, UMAP2, barcodes, combined_cnv_classification) |>
+  unique() |>
```
jashapiro, line 199: Faster, apparently. But a bigger question is why are there duplicates? (got it..., see other comments)
```suggestion
  dplyr::distinct() |> 
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+
+# get the gene expression counts for all marker genes
+marker_gene_exp <- logcounts(sce[marker_genes, ]) |>
+  as.matrix() |>
+  t() |>
+  as.data.frame() |>
+  tibble::rownames_to_column("barcodes")
+
+# pull out the UMAP coordinates and make a data frame to use for plotting
+coldata_df <- sce |>
```
jashapiro, line 126: Ah, this (the pivot after join) explains why there are duplicates below. I was confused by calling this `coldata_df` when it is not really that after faceting. I might give it a more explicit name that reflects that it has marker expression values as well.

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+    names_to = "method",
+    values_to = "classification"
+  )
+
+ggplot(marker_density_df, aes(x = sum_exp, color = classification)) +
+  geom_density() +
+  facet_wrap(vars(method))
+```
+
+This plot shows us that overall the tumor cells in each method have higher marker gene expression.
```
jashapiro, line 241: Probably worth saying that the `marker_gene_classification` plot here is circular, in that it just reflects the cutoff that was used for classification

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+  row_names_side = "left",
+  row_dend_side = "right",
+  row_names_gp = grid::gpar(fontsize = 10),
+  ## Column parameters
+  cluster_columns = TRUE,
+  show_column_names = FALSE,
+  bottom_annotation = annotation
+)
+```
+As expected, the tumor cells appear to have more marker genes that are expressed than normal cells.
```
jashapiro, line 287: formatting
```suggestion

As expected, the tumor cells appear to have more marker genes that are expressed than normal cells. 
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+  as.matrix()
+
+# create annotation df, make sure that dimensions match up by only getting unique barcodes 
+annotation_df <- coldata_df |> 
+  dplyr::select(barcodes, combined_cnv_classification) |> 
+  unique() 
+
+annotation <- ComplexHeatmap::columnAnnotation(
+  classification = annotation_df$combined_cnv_classification,
+  col = list(classification = c("Tumor" = "#00274C", "Normal" = "#FFCB05")))
```
jashapiro, line 263: I see you

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+```
+
+
+```{r}
+# plot heatmap of marker genes 
+ComplexHeatmap::Heatmap(
+  marker_gene_heatmap,
```
jashapiro, line 270: Just for clarity: this plot end up showing scaled expression somehow? The label is `matrix_29` and it looks like almost all values are at 1 or 0?

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+
+```{r, fig.height=10}
+geneset_plot_df |> 
+  tidyr::pivot_longer(
+    cols = ends_with("classification"),
+    names_to = "method",
+    values_to = "classification"
+  ) |> 
+  ggplot(aes(x = mean_score, color = classification)) +
+  geom_density(bounds = c(0,1), bw = 0.05) +
```
jashapiro, line 327: I don't think you want bound here, as the data you are plotting is not constrained to 0,1. 
```suggestion
  geom_density(bw = 0.05) +
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+
+### Gene set scores
+
+Next, we will look at the gene set scores for three different EWS-FLI1 gene sets. 
+The below plots will all use the `mean-GENESET` columns. 
+
+First we will just look at the gene set scores for each cell on the UMAP. 
+
+```{r}
+# plot tumor cells for each classification method
```
jashapiro, line 297: I don't know that this UMAP is needed, but it doesn't really hurt, I guess

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+geneset_heatmap <- geneset_plot_df |> 
+  dplyr::select(mean_score, geneset, barcodes) |> 
+  unique() |> 
+  tidyr::pivot_wider(values_from = mean_score, 
+                     names_from = barcodes) |> 
+  tibble::column_to_rownames("geneset") |> 
+  as.matrix()
+
+ComplexHeatmap::Heatmap(
+  geneset_heatmap,
```
jashapiro, line 349: check scaling again?

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+  ## Row parameters
+  cluster_rows = TRUE,
+  row_title = "",
+  row_title_gp = grid::gpar(fontsize = 12),
+  row_title_side = "left",
+  show_row_names = FALSE,
+  row_dend_side = "right",
+  row_names_gp = grid::gpar(fontsize = 10),
+  ## Column parameters
+  cluster_columns = FALSE,
```
jashapiro, line 420: I see you have this right, but the sorting is still not correct for some reason?

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+colnames(gain_cnv_df) <- stringr::word(colnames(gain_cnv_df), -1, sep = "_")
+
+# need to make a new annotation, since we want to annotate the rows here 
+annotation <- ComplexHeatmap::rowAnnotation(classification = annotation_df$combined_cnv_classification,
+                                            col = list(classification = c("Tumor" = "#00274C", "Normal" = "#FFCB05")))
+```
+
+
+```{r}
+ComplexHeatmap::Heatmap(
```
jashapiro, line 407: Just noting that you are getting repeated messages about rasterization: I would turn those off during setup: "Set `ht_opt$message = FALSE` to turn off this message."

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,492 @@
...
+)
+```
+
+We do not see the expected loss in Chr16, but this is not present in every patient. 
+Additionally, we had already noted the gain in Chr8 in the tumor cells, so there is still at least one expected CNV present. 
+Again, there is a clear separation of tumor and normal cells in terms of presence of CNVs. 
+
+## Save final tumor annotations
+
+After validating that the cells classified as tumor by both `CopyKAT` and `InferCNV` show the expected expression patterns and CNV patterns as tumor cells, we can confidently label those cells as tumor cells.
```
jashapiro, line 469: I think we might want to be a bit more careful about the final annotations here. I agree that the "Tumor" annotations seem reliable, but I am less confident in the "Normal", so I think we may want to retain some information about "unknown" or "unclassified" cells.

jashapiro:
> What would you call unclassified here? Are you saying anything that's called as a tumor cell with one method but not the other methods should fall into this category? I'm fine with having an unclassified group of cells, but I don't actually feel like any of these in this sample should be at this point. I think there's a pretty clear distinction between tumor and non-tumor cells when looking at all the data together.

"Unclassified" may be the wrong word, but yes, the cells where there is disagreement among methods. Part of what I want to see is where these cells are in the various distributions. Are they in the middle? 

Looking at the data as I see it, it seems like there could well be some missed tumor cells in the "Normal" group. The number would be small, but I am maybe not as confident as you are!

allyhawkins:
> > What would you call unclassified here? Are you saying anything that's called as a tumor cell with one method but not the other methods should fall into this category? I'm fine with having an unclassified group of cells, but I don't actually feel like any of these in this sample should be at this point. I think there's a pretty clear distinction between tumor and non-tumor cells when looking at all the data together.
> 
> "Unclassified" may be the wrong word, but yes, the cells where there is disagreement among methods. Part of what I want to see is where these cells are in the various distributions. Are they in the middle?
> 
> Looking at the data as I see it, it seems like there could well be some missed tumor cells in the "Normal" group. The number would be small, but I am maybe not as confident as you are!

I see what you're saying. I can update to have an unknown class if it was identified as tumor by both methods, add that into the plots and then based on the results we can decide if they stay unknown or fit with tumor or normal cells better.

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
allyhawkins: 〽️

allyhawkins:
@jashapiro I've made the following changes which I believe should address your comments: 

- I created a function for plotting the heatmaps, one for plotting gene expression and one for CNV. This should hopefully make the code around the heatmaps simpler. 
- There is no scaling on the heatmaps, but I had an error in how I set up the color scheme that makes it look like that. They should now have the correct scale and titles for the legends. 
- I fixed the ordering for the CNV heatmaps so the chromosomes are all in numerical order. 
- The major update here is the addition of the "Unknown" class throughout exploration. I called only tumor cells as those that are tumor in both CNV methods and normal as cells that were considered normal by both CNV methods. All other cells were classified as "Unknown". The plots then incorporated that updated annotation and you can see that most of those cells line up with the normal cells with a few tumor cells in there. They also do fall in the middle for the CNV gains/losses. 
- That being said, I then re-annotated the Unknown based on whether or not they had a gain in Chr8 specifically. Alternatively we could keep these as unknowns... but I do think there is some _mostly_ clear separation to be able to classify these. 

This should be ready for another look! 
[SCPCL000822_tumor-cell-validation.html.zip](https://github.com/user-attachments/files/15809151/SCPCL000822_tumor-cell-validation.html.zip)

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,600 @@
...
+
+
+- First we look at the cells that are annotated as tumor cells by each method. 
+Using this information, we call any cell that is annotated as a tumor cell by both `CopyKAT` and `InferCNV` as a tumor cell. 
+- We then validate that these are in fact tumor cells by doing the following: 
+  - Looking at marker gene expression. 
+  - Looking at expression of EWS-FLI1 target gene sets. 
+  - Confirming presence of known CNVs in Ewing sarcoma
+  
+## Define functions
```
jashapiro, line 23: I might move this to after setup.

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,600 @@
...
+
+geneset_scores_file <- file.path(results_dir, "SCPCL000822_gene-set-scores.tsv")
+
+# output files 
+final_annotations_dir <- file.path(module_base, "cell_type_annotations", "SCPCS000490")
+fs::dir_create(final_annotations_dir)
+final_annotations_file <- file.path(final_annotations_dir, "SCPCL000822_tumor-classifications.tsv")
+```
+
+
```
jashapiro, line 167: A subheader here.
```suggestion
### Process input files

```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,600 @@
...
+classification_df <- classification_df |> 
+  dplyr::mutate(
+    combined_cnv_classification = dplyr::case_when(
+      # only tumor if both call tumor 
+      cnv_proportion_classification == "Tumor" & copykat_classification == "Tumor" ~ "Tumor",
+      # only normal if both call normal and/or reference
+      cnv_proportion_classification == "Normal" & copykat_classification == "Normal" ~ "Normal",
+      cnv_proportion_classification == "Reference" & copykat_classification == "Normal" ~ "Normal",
+      # if normal/tumor mismatch set to unknown 
+      .default = "Unknown"
```
jashapiro, line 304: I don't really like "Unknown" here. Maybe something more descriptive?
I suggested "conflicted", but "ambiguous", "disagreement" or something like that?

You may also want to throw in a factor ordering step here?

```suggestion
      # if normal/tumor mismatch set to Conflicted
      .default = "Conflicted"
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,600 @@
...
+In particular, we expect those that have a gain in Chr8 are tumor cells, so let's just look at Chr8 gains and see if we can find a cut off to use to categorize our "Unknown" cells. 
+
+```{r}
+#density plot of chr8 gains
+ggplot(classification_df, aes(x = proportion_scaled_dupli_chr8, fill = combined_cnv_classification)) +
+  geom_density() +
+  facet_grid(rows = vars(combined_cnv_classification))
+```
+
+It looks like we have a pretty clear distinction between the `proportion_scaled_dupli_chr8` for normal and tumor cells and we could call any cell with a value > 0.25 as tumor cells and those that are less than 0.25 as normal cells.
```
jashapiro, line 539: This looks less clear to me, as I think there are some high values in the normal samples: they are just swamped by the many low values.

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,600 @@
...
+
+In both of these plots, the majority of "Unknown" cells cluster between tumor and normal or with the normal cells. 
+
+## Create final tumor annotations
+
+In looking at the plots, it looks like there might be some of the "Unknown" cells that should actually be classified as tumor cells. 
+In particular, we expect those that have a gain in Chr8 are tumor cells, so let's just look at Chr8 gains and see if we can find a cut off to use to categorize our "Unknown" cells. 
+
+```{r}
+#density plot of chr8 gains
```
jashapiro, line 533:
```suggestion
# density plot of chr8 gains
```

jashapiro (commented):
This looks pretty good to me, and I appreciate the cleanup!

> * That being said, I then re-annotated the Unknown based on whether or not they had a gain in Chr8 specifically. Alternatively we could keep these as unknowns... but I do think there is some _mostly_ clear separation to be able to classify these.

I think I would not add the extra reannotation here, at least not at this stage. There are some subtle bumps in the "Normal" class with higher chr8 scores, So I would not want to use that as a cutoff on their own. I could see adding a column with just the chr8 scores, but I would want to see this across a few samples before deciding that, as we might expect to find different CNVs in different samples.

I think the real question is what is the goal of the "final" annotations at this point? Is it better to leave more cells as ambiguous or undefined when we have disagreements, or to try to get as many cells as possible classified, even if we are wrong. My bias tends to be toward the former, but this case where we have relatively little disagreement between methods may not be the best test case. If only calling where the methods agree resulted in obviously bad results, or very few net calls, I could reconsider!

allyhawkins:
> I think I would not add the extra reannotation here, at least not at this stage. There are some subtle bumps in the "Normal" class with higher chr8 scores, So I would not want to use that as a cutoff on their own. I could see adding a column with just the chr8 scores, but I would want to see this across a few samples before deciding that, as we might expect to find different CNVs in different samples.

Just a general comment that I don't expect alterations in Chr8 in every sample. The genome studies of Ewings patients have shown that Chr8, Chr12, Chr16, and Chr1 are most prevalent but not present in every single patient. I was using Chr8 here because it is the most striking of those options and without knowing the actual mutation status of the patient at the DNA level and only using both CopyKAT and InferCNV output, I assume that most tumor cells will have Chr8. So classifying cells with Chr8 may be something that only happens with this sample. 

That being said, I would agree that we don't _need_ to reannotate here and I think it's okay to have some unknown/ ambiguous cells. So I kept the plot looking at Chr8, but removed the reannotation. 

> I think the real question is what is the goal of the "final" annotations at this point? Is it better to leave more cells as ambiguous or undefined when we have disagreements, or to try to get as many cells as possible classified, even if we are wrong. My bias tends to be toward the former, but this case where we have relatively little disagreement between methods may not be the best test case. If only calling where the methods agree resulted in obviously bad results, or very few net calls, I could reconsider!

The goal here is to annotate as many cells as tumor or normal that we can confidently annotate. I think it's okay to have cells that we don't annotate because we don't feel confident in those annotations. Here it's easy because most cells agree and we have a small population of cells that don't. In this case, I am fine with leaving those as "Ambiguous" (I updated the label) since we don't have a consensus. 

However, in looking at other samples this week, this is not going to be typical. In the other cases we don't have a lot of agreement between methods so I think we need to investigate and find out which results we trust more and label cells accordingly. In those scenarios I do think we are going to need to do some refining of classification outside of just taking the consensus among methods.

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,567 @@
...
+
+This plot shows us that there is a pretty clear consensus that most of the cells in the bottom right group are tumor cells. 
+From the original marker gene analysis we know that not all marker genes are expressed and that there is quite a bit of variance in expression. 
+Because of that and the fact that both CNV inference methods show similar annotations, we will classify any cells that are labeled as tumor by both `CopyKAT` and `InferCNV` as tumor cells. 
+
+```{r}
+# create a combined classification between inferCNV and copyKAT
+# cells that are classified as tumor by both are kept as tumor
+# classified by both as normal are normal 
+# in between are classified as unknown
```
jashapiro, line 294:
```suggestion
# in between are classified as ambiguous
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,567 @@
...
+classification_df <- classification_df |> 
+  dplyr::mutate(
+    combined_cnv_classification = dplyr::case_when(
+      # only tumor if both call tumor 
+      cnv_proportion_classification == "Tumor" & copykat_classification == "Tumor" ~ "Tumor",
+      # only normal if both call normal and/or reference
+      cnv_proportion_classification == "Normal" & copykat_classification == "Normal" ~ "Normal",
+      cnv_proportion_classification == "Reference" & copykat_classification == "Normal" ~ "Normal",
+      # if normal/tumor mismatch set to unknown
```
jashapiro, line 303:
```suggestion
      # if normal/tumor mismatch set to ambiguous 
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,567 @@
...
+# be sure to remove any duplicates 
+classification_df |> 
+  dplyr::select(UMAP1, UMAP2, barcodes, combined_cnv_classification) |>
+  dplyr::distinct() |> 
+  ggplot(aes(x = UMAP1, y = UMAP2, color = combined_cnv_classification)) +
+  geom_point(alpha = 0.5, size = 0.5)
+```
+
+The remainder of the notebook will focus on validating that these in fact are tumor cells and all other cells are classified as normal. 
+We will also look at where the unknown cells fall in relation to tumor vs. normal.
```
jashapiro, line 321:
```suggestion
We will also look at where the ambiguous cells fall in relation to tumor vs. normal. 
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,567 @@
...
+
+ggplot(marker_density_df, aes(x = sum_exp, color = classification)) +
+  geom_density() +
+  facet_wrap(vars(method))
+```
+
+This plot shows us that overall the tumor cells in each method have higher marker gene expression. 
+Although the `InferCNV` alone and combined CNV classifications appear to have the most distinct separation in marker gene expression. 
+
+Another note is that the "Unknown" class appears to have marker gene expression more similar to the normal cells.
```
jashapiro, line 371:
```suggestion
Another note is that the "Ambiguous" class appears to have marker gene expression more similar to the normal cells. 
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,567 @@
...
+```{r}
+# plot heatmap of marker genes 
+plot_gene_heatmap(marker_gene_heatmap, 
+                  row_title = "Marker gene symbol",
+                  legend_title = "Marker gene \nexpression",
+                  annotation = annotation)
+```
+
+As expected, the tumor cells appear to have more marker genes that are expressed than normal cells. 
+The "Unknown" cells appear to have a small group of cells that cluster with the tumor cells on the right hand side of the plot.
```
jashapiro, line 405:
```suggestion
The "Ambiguous" cells appear to have a small group of cells that cluster with the tumor cells on the right hand side of the plot. 
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,567 @@
...
+  ) |> 
+  ggplot(aes(x = mean_score, color = classification)) +
+  geom_density(bw = 0.05) +
+  facet_grid(rows = vars(method), 
+             cols = vars(geneset))
+```
+
+For the most part the gene set scores are higher in tumor cells regardless of classification. 
+
+We also see that the "Unknown" cells identified in the combined classification mostly mirror the gene set expression scores for normal cells with maybe a small group of cells having more similar gene set scores to tumor cells.
```
jashapiro, line 452:
```suggestion
We also see that the "Ambiguous" cells identified in the combined classification mostly mirror the gene set expression scores for normal cells with maybe a small group of cells having more similar gene set scores to tumor cells. 
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,567 @@
...
+plot_gene_heatmap(geneset_heatmap,
+                  annotation = annotation,
+                  legend_title = "Gene set \nscore")
+```
+
+For the most part we see that cells with higher gene set scores are classified as tumor cells. 
+One thing to note is that the `SILIGAN` gene set appears to have some expression in normal cells. 
+However, if you look at the spots where all three gene sets have higher scores, these apply to tumor cells. 
+
+Again, most of the "Unknown" tumor cells tend to align with normal cells, with a few grouping in the area where cells transition from tumor to normal.
```
jashapiro, line 478:
```suggestion
Again, most of the "Ambiguous" tumor cells tend to align with normal cells, with a few grouping in the area where cells transition from tumor to normal. 
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,567 @@
...
+                 cnv_col = "proportion_scaled_loss", 
+                 annotation = annotation, 
+                 legend_title = "Copy number \nloss")
+```
+
+We do not see the expected loss in Chr16, but this is not present in every patient. 
+Additionally, we had already noted the gain in Chr8 in the tumor cells, so there is still at least one expected CNV present. 
+Again, there is a clear separation of tumor and normal cells in terms of presence of CNVs. 
+
+In both of these plots, the majority of "Unknown" cells cluster between tumor and normal or with the normal cells.
```
jashapiro, line 527:
```suggestion
In both of these plots, the majority of "Ambiguous" cells cluster between tumor and normal or with the normal cells. 
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000490/SCPCL000822_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,567 @@
...
+
+We do not see the expected loss in Chr16, but this is not present in every patient. 
+Additionally, we had already noted the gain in Chr8 in the tumor cells, so there is still at least one expected CNV present. 
+Again, there is a clear separation of tumor and normal cells in terms of presence of CNVs. 
+
+In both of these plots, the majority of "Unknown" cells cluster between tumor and normal or with the normal cells. 
+
+## Create final tumor annotations
+
+In looking at the plots, it looks like there might be some of the "Unknown" cells that should actually be classified as tumor cells.
```
jashapiro, line 531:
```suggestion
In looking at the plots, it looks like there might be some of the "Ambiguous" cells that should actually be classified as tumor cells. 
```

jashapiro (approved):
LGTM. I just went through and changed some text to match new labels.

PR #503, opened 2024-06-05
Document the workflow results script

sjspielman:
Towards #428 (but not 100% actually sure if it closes it? Note also that I opened #502 for later).

This PR documents the results download script by adding a section, along with a little rearrangement, to the "getting access to data" doc. I tried to mostly mirror the way we describe the data download script, and I added a stub reminder to link in forthcoming `nf` docs. I also added a link to these docs in the "documenting your analysis" page as a reminder that all input data needs to be documented.

There's one spot with a clear request for reviewer phrasing support, but of course more generally let me know about phrasing/organization throughout! Are there spots I should add more details or context?

To facilitate the docs 🎢 ahead of us, I also modified the links and spell check GHAs to run on PRs to the reproducibility branch.

PR #505, opened 2024-06-05
Port project tools into getting started

### LINKED ISSUE #489: Move "additional project tools" to "getting started"
This issue tracks moving content in `software-tools/general-tools` into `getting-started`.

Its `index.md` page should also have an updated flow: We support UNIX/Linux and use git(hub), so you'll need to be comfortable with things like terminal, markdown, and git (link to git docs).

sjspielman:
Closes #489 

This PR moves the project tools pages into getting started. I added an index page that lists out the main project tools where I link out to the relevant docs. Am I missing any other ones to link? Note there are some STUB_LINKs in there too (

Note that _locally_, I ran the spellchecker and checked relative links, and they all look good! (Those GHAs won't run here since this isn't headed into `main`, and #503 has the changes needed to get these GHAs running on this target branch).


allyhawkins (approved):
LGTM

PR #510, opened 2024-06-05
GHAs should run on reproducibility branch

sjspielman:
This PR cherry-picks a commit from #503 to get it in sooner - we should run the link check and spell check actions on the reproducibility branch.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - main
       - reproducibility-docs
+      - feature/*
     paths:
       - docs/**
 
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -1,7 +1,9 @@
 name: Check docs links
 on:
   pull_request:
-    branches: main
+    branches:
+      - main
+      - reproducibility-docs
     paths:
       - docs/**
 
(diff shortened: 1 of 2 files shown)


.github/workflows/spellcheck.yml
```diff
@@ -1,7 +1,9 @@
 name: Check Spelling
 on:
   pull_request:
-    branches: main
+    branches:
+      - main
+      - reproducibility-docs
```
jaclyn-taroni, line 6:
```suggestion
    branches:
      - main
      - reproducibility-docs
      - feature/*
```

.github/workflows/check-links.yml
```diff
@@ -1,7 +1,9 @@
 name: Check docs links
 on:
   pull_request:
-    branches: main
+    branches:
+      - main
+      - reproducibility-docs
```
jaclyn-taroni, line 6:
```suggestion
    branches:
      - main
      - reproducibility-docs
      - feature/*
```

sjspielman:
Noting that based on the fact that this exact commit ran the actions in #503, I'm counting that as "tested." Let me know if you prefer that I test here with a `docs/` change, though!

jaclyn-taroni (commented):
Next time we do something like this, we should use a naming convention `feature/{description}` (this one would be `feature/reproducibility-docs`) so we can use wildcards here and in the branch protection rules (h/t David).

jaclyn-taroni (approved):
Meant to approve 😄

sjspielman:
> so we can use wildcards here and in the branch protection rules (h/t David).

nice!

PR #506, opened 2024-06-05
(1/5) Rearrange software docs: Using renv

sjspielman:
Towards #458 

This is the first of several ("pseudostacked"!) PRs to break up the current software page. Note the updated scheme for this plan in https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/458#issuecomment-2150560957

This PR adds the following:
- Initializes a reproducibility directory (just a stub index for now)
- Initializes a managing software directory with an `index.md`. This file is wholly new.
- Ports `renv` software content into a new file `using-renv.md` in the managing software directory. There are a few rephrasings here that I noticed, but no "big" changes besides the occasional updated wording.

----

Here's what you can expect for the additional PRs to close out this issue:
- A PR to add a `using-conda.md` doc
- A PR to add a `reporting-dependencies.md` doc
- [EDIT: this step added] A PR to add the primer page for #498 (since this will affect relative links 👇 )
- After all of the above (including this PR) go in, I'll file a PR which removes the original software dependencies doc, updates the `mkdocs.yml` file to match the current set of software files, and fixes all the relative links that break after the `mkdocs.yml` change 🎢  
  - I will likely also add some relative links to the `documenting-your-analysis.md` page that points to a lot of these changes

Issue #458, opened 2024-05-24
Docs request: Break up the software requirements page


sjspielman:
### What is the documentation improvement or update you wish to see?

Related to https://github.com/AlexsLemonade/OpenScPCA-admin/issues/194

--- 

We plan to change how we present software information by breaking up the `Software requirements` page into three different docs. While doing this, we need to:
- Ensure headers are consistently worded within a page
- Update the `index.md` page in `Determining requirements`
- Update any relative links throughout docs

Here are the new pages and their sections:

## Creating environments

- [Using renv](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/determining-requirements/determining-software-requirements/#using-renv)
- [Managing dependencies with conda](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/determining-requirements/determining-software-requirements/#managing-software-dependencies-with-conda)

## Adding new software packages

[The whole section from doc](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/determining-requirements/determining-software-requirements/#adding-software-to-the-environment-and-tracking-installed-software)

## Reporting Software Dependencies

- [`session.info()` in R](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/determining-requirements/determining-software-requirements/#using-sessioninfo)
- [`session_show.info()` in Python](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/determining-requirements/determining-software-requirements/#using-session_infoshow-in-python)
- We should also ensure there's a link on this page to [Documenting your analysis](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/analysis-modules/documenting-analysis/)

### Is there any additional context you would like to provide?

Noting that we should also consider adding some docs about making a brand new environment.yml file, as briefly discussed in #449, although probably without the use of `conda export`. This is likely to be a separate PR.

### If this request is for an existing documentation page, please provide the link here.

https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/determining-requirements/determining-software-requirements/

sjspielman:
After further discussion, these docs still going to get broken up, and then will _also_ be moved into a new docs section on reproducibility.

sjspielman:
Noting to send to @jaclyn-taroni for review.

sjspielman:
[...]

PR #511, opened 2024-06-06
Add software primer page

### LINKED ISSUE #498: Add primer on software requirements to analysis docs
With #458 and #494, software environment docs are going to be moved out of the `contributing-to-analyses` section. However, this information is still important to communicate somewhere where we introduce analyses. We should therefore add a "primer" page briefly introducing software environments into the `contributing-to-analyses/analysis-modules` section, alongside the compute docs which are moving up there as well (#488).

sjspielman:
Closes #498 
Related to #458 
See #506 for a little more context

---

This PR adds a short "primer" page on software dependencies to the analysis page. I added this to `mkdocs` as well to check formatting since I used buttons! I also did a bit of rearrangement of the conda doc that went in in #507 - two of the headers were the wrong level, and I updated the order so that "using an existing environment" comes after all instructions for creating/managing in the first place.

Noting I indeed would also like for @dvenprasad to have a look here for any feedback on the language as it relates to our goal of making it easier for contributors to find docs on adding dependencies.

Edit: I also want to note that a lot of the language here is very similar (..the same?) as that in this new managing software index file: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/reproducibility-docs/docs/ensuring-repro/managing-software/index.md. I think this repetition is probably good, do you?

### DIFF
--- a/docs/contributing-to-analyses/analysis-modules/documenting-analysis.md
+++ b/docs/contributing-to-analyses/analysis-modules/documenting-analysis.md
@@ -10,7 +10,7 @@ You should fill out the template for your analysis module and provide the follow
 - Usage (How to [run the analysis](running-a-module.md))
 - Input files
 - Output files
-- Software requirements
+- [Software and package dependencies](./module-dependencies.md)
 - Computational resources
 
 For some examples of analysis module `README.md` files, please see [our example R module](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/hello-R/README.md) and/or [our example Python module](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/hello-python/README.md).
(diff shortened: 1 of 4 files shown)


jaclyn-taroni:
> Edit: I also want to note that a lot of the language here is very similar (..the same?) as that in this new managing software index file: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/reproducibility-docs/docs/ensuring-repro/managing-software/index.md. I think this repetition is probably good, do you?

I have not looked at the files changed yet but my sense is yes, repetition is good as long as things do not conflict. So consider how likely it is that conflicting information would pop up as these are maintained over time!

jaclyn-taroni (approved):
This looks good to me, but please see my general comment about repeating ourselves: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/511#issuecomment-2155085131

sjspielman:
> So consider how likely it is that conflicting information would pop up as these are maintained over time!

Conflicting information would occur in this case if we were move away from using conda and renv in the first place....and if that happens we have bigger "problems"! So, this one seems like good repetition to me.

dvenprasad (approved):
LGTM! 🐌 🐌 🐌 

(You have actually hit the submit button for the comment to go through. I know, shocking! 🤦‍♀️ )

Issue #521, opened 2024-06-13
Add index page for reproducibility section


sjspielman:
This `index.md` file has been created but currently has no content. Eventually, it should contain an overview of how we ensure reproducibility throughout the OpenScPCA project. This page will include the reproducibility overview figure (#491). 

We should also include a section about how you should always be working from the `openscpca` conda environment (or your module's conda environment, as the case may be) when developing.

sjspielman:
Closed by #599

PR #522, opened 2024-06-13
Add tip for environment.yml creation

### LINKED ISSUE #449: Update conda docs: create an environment.yml file
While working with conda environments, I came across a situation that we did not document on the software page (note that we expect this page to change too, see https://github.com/AlexsLemonade/OpenScPCA-admin/issues/194).

Currently, we only document creating a conda lock file from an existing `environment.yml`. If someone wants to add a conda environment to a module _after the module has been created_, we currently direct contributors to run the environment creation command from the repo root. In this case, there is no existing `environment.yml` file in the module to run conda lock on. We therefore need to document exporting an environment to be able to create a conda lock file from it*.

**edit** - looks like we never actually introduce `export` even to update an existing environment, so there may be more to think about here.

sjspielman:
~Closes #449~

I was looking over our freshly rearranged docs to determine what we really need to do for #449. It turns out that the docs rearrangement really did make a difference here! The only thing I really thought to add was a tip to make sure you're creating the new environment from the `openscpca` environment. Note that I also opened #521 to make sure we document this more generally (we only currently say it during conda setup).

### DIFF
--- a/docs/ensuring-repro/managing-software/using-conda.md
+++ b/docs/ensuring-repro/managing-software/using-conda.md
@@ -28,7 +28,14 @@ You can activate the environment by running the following command:
 conda activate openscpca-{module_name}
 ```
 
-If no `environment.yml` file is present, you can create and activate a new environment for the module by running the following command from the module's root directory, replacing `{module_name}` with the name of the module you are working on:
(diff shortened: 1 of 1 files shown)


sjspielman:
womp, this helps, but does _not_ close #449!

sjspielman:
I figure we can actually do #449 later.

jashapiro (approved):
🛼

PR #531, opened 2024-06-18
Update doublet Dockerfile

### LINKED ISSUE #460: Update doublet-detection Dockerfile

_No response_

Currently, the `doublet-detection` analysis module's Dockerfile considers the module's `renv` environment only. In #454, the module is being updated to also include a conda environment. This issue tracks updating the Dockerfile to consider conda too.

Comparing `bioconductor` and `miniconda` base images to one another, I think it makes the most sense to keep `bionconductor` base and install `miniconda` itself into it, since `bioconductor` is based on `rstudio` is based on `r-ver`, but (edit: correct link to dockerfile) [`miniconda` just uses linux](https://github.com/ContinuumIO/docker-images/blob/main/miniconda3/debian/Dockerfile) and doesn't have too many installs. Tagging @jashapiro for any other base image suggestions!

An updated Dockerfile

N/A

N/A

_No response_

sjspielman:
Closes #460 

This PR updates the Dockerfile in the `doublet-detection` module to use both R and python. This Dockerfile _nearly_ builds, save some R package installation errors which may be problems on R's side (although both CRAN and bioconductor repos, so maybe not?). I'm going to keep playing around with this a bit, so opening as a draft.

Edit: building now! Let me know if I should go ahead and add a docker GHA here, too.


sjspielman:
https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/531/commits/810522664e7b4e6ca253b76fd51ddb080aba7b5e is probably not needed, so will likely end up reverting but leaving for now just on the off chance...

.github/workflows/docker_doublet-detection.yml
```diff
@@ -0,0 +1,63 @@
+name: Build Docker image for doublet-detection
+
+concurrency:
+  # only one run per branch at a time
+  group: "doublet-detection-${{ github.ref }}"
+  cancel-in-progress: true
+
+on:
+#  pull_request:
```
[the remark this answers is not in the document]
sjspielman: Not necessarily, other than waiting until the module matures a bit more?

.github/workflows/docker_doublet-detection.yml
```diff
@@ -0,0 +1,63 @@
...
+    branches:
+      - main
+    paths:
+      - analyses/doublet-detection/Dockerfile
+      - analyses/doublet-detection/.dockerignore
+      - analyses/doublet-detection/renv.lock
+      - analyses/doublet-detection/conda-lock.yml
+  push:
+    tags:
+      - "v*.*.*"
```
jashapiro, line 19: Tag pushes should be handled by the all-in-one workflow now.
```suggestion
    tags:
      - "v*.*.*"
```

.github/workflows/docker_doublet-detection.yml
```diff
@@ -0,0 +1,63 @@
...
+    inputs:
+      push-ecr:
+        description: "Push to AWS ECR"
+        type: boolean
+        required: false
+  workflow_call:
+    inputs:
+      push-ecr:
+        description: "Push to AWS ECR"
+        type: boolean
```
jashapiro, line 37: This workflow is never called from other workflows, so this should not be needed.
```suggestion
```

.github/workflows/docker_doublet-detection.yml
jashapiro: For some reason, I don't have the `push:` block in the template workflow, but I think it probably should be here.

jashapiro (approved):
Just a couple little updates to the workflow, but this should be good to go otherwise.

sjspielman:
It's worth noting I created this file starting off `docker_simulate-sce.yml` - https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/docker_simulate-sce.yml. Do you want me to make parallel changes there too?

jashapiro:
> It's worth noting I created this file starting off `docker_simulate-sce.yml` - https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/docker_simulate-sce.yml. Do you want me to make parallel changes there too?

Yes, and I think we probably need to add the `push` block to the template action as well. I can't think of why I might have removed it there.

sjspielman:
Looks like you already fixed the template file - it's just not in `main` yet! https://github.com/AlexsLemonade/OpenScPCA-analysis/blame/reproducibility-docs/templates/workflows/docker_module.yml

PR #532, opened 2024-06-18
Annotate tumor cells in SCPCS000492

### LINKED ISSUE #480: Annotate tumor cells in SCPCS000492

#292 

The first step of annotating cell types in the Ewing's samples is annotating tumor cells. As of now, we have utilized multiple methods to obtain annotations - marker gene analysis, CellAssign, CopyKAT, and InferCNV. Now that we have all this data on each sample we will want to create final tumor cell annotations for each sample. The goal is to confidently annotate a handful of samples and then use those as a reference to annotate the remaining samples. This issue is to track creating a TSV file annotating each cell as normal or tumor in SCPCS000492 using the combination of results from running different annotation methods. 

This PR will contain a notebook that reads in the results from marker gene analysis, CopyKAT, InferCNV, and gene set scoring (see #478) for SCPCS000492 and uses those results to come up with a final combined annotation of tumor vs. normal. 

- We will look at the consensus between tumor and normal cells identified by CopyKAT and InferCNV. We will validate these cells as tumor by looking at expected chromosome gains. 
- We will look at marker gene expression for each marker gene in cells that are identified as tumor vs. normal using CNV methods. 
- Additionally, we will look at gene expression of the EWS-FLI1 signature in tumor and normal cells to validate which cells should be called as tumor. 
- Final annotations will then be saved as a TSV file with two columns - barcode and tumor annotation (Tumor or Normal). 

_No response_

_No response_

_No response_

allyhawkins:

Closes #480 

Here I am adding an exploratory notebook that looks at the results from multiple tumor cell annotation methods (output from `annotate-tumor-cells.sh`) and aims to annotate the tumor cells in SCPCS000492. The output is a TSV file with annotations as either Tumor, Normal, or Ambiguous. 

This was not as straight forward as SCPCS000490 where we took the consensus between CNV methods and annotated those as tumor cells. Looking at the initial results from CopyKAT, InferCNV, and marker gene based classification, we don't see a lot of agreement in the cells that are annotated as tumor. Specifically, those with high marker gene expression tend to not necessarily be the ones with higher CNV proportions. Additionally, we don't see many cells with the expected CNVs in Ewing samples and there is not as clear a distinction between what would be considered normal and tumor cells. 

Because of this I ended up looking at the gene set scores from the EWS-FLI1 gene sets and using those to classify cells as tumor and normal cells. I was more wary of the CNV results. The gene set scores appeared to mostly line up with the marker gene expression, so ultimately I classified cells as tumor if they had above the mean gene set score for all three gene sets and were considered tumor cells by marker gene analysis. This is probably on the stricter side. 

I also ended up trying to use `SingleR` to classify these cells and used the annotations from SCPCS000490 as the reference. I then compared the annotations from `SingleR` to just setting cut offs for gene set scores and marker genes. It looks like the `SingleR` annotations actually capture cells that are considered tumor by both gene expression and CNV methods. I tend to trust those annotations more as they are not quite as arbitrary in the cut offs and using information from a sample that we feel much more confident in. 

Yes 

Just noting that there are no results currently on S3, _but_ the annotations file is ~ 250 KB so it doesn't pass pre commit. So maybe we should be saving those in S3 rather than trying to save them to the repo? 

The output is a TSV file with two classification columns - one obtained from SingleR and one from using marker gene and gene set classification. Both columns label cells as either tumor, normal, or ambiguous. 

Generally, I think using `SingleR` is the more reliable form of annotation here, but it is not without bias and relies on having a really good reference that we feel good about. Originally, I wanted to annotate multiple samples to use as a reference with `SingleR` for other samples, but I actually wonder if we the rest of samples don't have a clear consensus between CNV and marker gene methods then we should just be using `SingleR` with the "good" sample. 

Also, I tend to believe these samples will contain a high proportion of tumor cells with a low proportion of normal cells, so I expect the majority to be tumor cells which is what we are seeing with `SingleR`. 

Here's a full copy of the rendered notebook that includes commentary on the results: 

[SCPCL000824_tumor-cell-validation.html.zip](https://github.com/user-attachments/files/15889781/SCPCL000824_tumor-cell-validation.html.zip)

Should be able to render the notebook locally in R without any special requirements. 

Generally, I struggled a lot with this sample since there was no clear consensus between methods. I do tend to trust the gene expression based classification over CNV methods. If there are other plots that would be helpful to show please let me know. 

Just another note that I can't commit the output TSV file at this point. I could either add the TSV files to the results directory instead and sync to S3 or make an exception for TSV files in this directory in the pre commit config file. 


analyses/cell-type-ewings/cell_type_annotations/SCPCS000492/SCPCL000824_tumor-cell-validation.Rmd
```diff
@@ -0,0 +1,615 @@
...
+  restrict = rownames(sce)
+)
+
+# get singler results 
+singler_results <- SingleR::classifySingleR(
+  trained = singler_model,
+  test = sce,
+  fine.tune = TRUE
+)
+```
```
jashapiro, line 484: I think you can probably do this in the one-step mode. 

The one thing I would suggest is that you remove the `Ambiguous` classification from the SingleR reference. That is not a "real" cell class, so we don't want to assign any cells to it!

jashapiro: On the other hand, I am not sure I am comfortable with using SingleR with only 2 classes: it seems likely that it will overfit those two groups. Not sure the immediate solution here, though we could perhaps see what happens with multiple references: one with just tumor-normal, one with more cell types from normal tissue?

jashapiro (commented):
I did not do a careful review here, because I think this may well signal the need to go toward some different methods, particularly in the scoring based on gene sets. 

As I noted in one of the line comments, I think the major "problem" here is likely to be that this is a more pure Ewings sample than the first one. The result is that the distribution of gene set scores is not as clearly bimodal, and doing things like Z-scoring and taking the mean just isn't going to work well. 

The main way I would want to approach this for a general case would be to use some kind of mixture model to fit a mixture of two normal distributions, then classify the cells by which one they fell into better. That would add a lot of work, which could potentially pay off, but definitely is going to be a whole other set of PRs.

A quick alternative is to look more at the raw (well, normalized) marker gene scores or gene set scores: what were the cutoff values from the first dataset, and what happens if you apply that as the cutoff here? Do you get a more expected separation?

I guess what I would do first is plot the  the marker gene and gene set expression scores from `SCPCL000822` and `SCPCL000824` together, to see where the`SCPCL000824` scores fall relative to the clear bimodal scores from `SCPCL000822`. Once we have that, we may have a better sense of where to go next. Hopefully we won't have to go full mixture model, but we might...

jashapiro:
> What is the name of your results bucket on S3?
Just noting that there are no results currently on S3, but the annotations file is ~ 250 KB so it doesn't pass pre commit. So maybe we should be saving those in S3 rather than trying to save them to the repo?

I did have a couple thoughts about file size issues too, before I go. One is that you could just change this to output `.tsv.gz` and things would shrink a lot. Alternatively you could write out files in some other format, like [parquet or arrow](https://arrow.apache.org/docs/r/articles/read_write.html). The disadvantage of all of these is that you loose the ability to easily diff the files, but for intermediate files that might be okay, and having them in the repo for easier review/inspection by outside users may be worth the tradeoff. 

I think this is probably a larger issue that we want to have a more general discussion about: when do we want to keep outputs in the repo vs. on S3 & how much is it worth to use text files vs. other formats that may be much more efficient, while maintaining ease of use.

allyhawkins:
> A quick alternative is to look more at the raw (well, normalized) marker gene scores or gene set scores: what were the cutoff values from the first dataset, and what happens if you apply that as the cutoff here? Do you get a more expected separation?

> I guess what I would do first is plot the the marker gene and gene set expression scores from SCPCL000822 and SCPCL000824 together, to see where theSCPCL000824 scores fall relative to the clear bimodal scores from SCPCL000822. Once we have that, we may have a better sense of where to go next. Hopefully we won't have to go full mixture model, but we might...

I just want to note that for classifying cells using marker gene expression we are: 
- transforming the expression of each marker gene 
- summing the transformed values
- using a cutoff of sum > 0 to classify cells as tumor or normal 

This same approach is being done for both methods and they both show bimodal distributions of marker genes. 

**SCPCL000822**
Screenshot 2024-06-18 at 4 48 43 PM

**SCPCL000824**
Screenshot 2024-06-18 at 4 49 25 PM

As for the gene set scores, we never actually used the gene set scores to classify cells in SCPCL000822, we only used them to validate that the gene set scores were indeed higher in tumor cells than normal cells. You are correct that here the gene set scores are not a bimodal distribution like they were in SCPCL000822. 

image

So we didn't actually have any cutoffs where we could use the same value in both, but I did test taking the mean of each of the distributions for SCPCL000822 and using those as the cutoffs (I also just estimated cutoffs from the plot that would make sense which are pretty equivalent to the mean). When I do that the results don't really change in terms of which cells get called as tumor cells. 

This is an example plot from classifying using one of the gene sets: 
image

I also want to note that I did try using [`AUCell`](https://bioconductor.org/packages/release/bioc/vignettes/AUCell/inst/doc/AUCell.html) to classify these cells with each of the gene sets and because of the lack of bimodal distribution, that didn't work very well. Only a few cells were classified as tumor and there was no clear cutoff. 

Generally, samples look more like this one where most cells are tumor cells so there probably won't be a clear bimodal distribution. 

> On the other hand, I am not sure I am comfortable with using SingleR with only 2 classes: it seems likely that it will overfit those two groups. Not sure the immediate solution here, though we could perhaps see what happens with multiple references: one with just tumor-normal, one with more cell types from normal tissue?

I wonder if we could combine a celldex reference with the tumor cells from SCPCL000822 to create a custom reference? This way we would have more than the two classes. This might avoid the overfitting problem and allow for SingleR to hopefully better distinguish between potential normal cell types and then tumor cells.

jashapiro:
> This same approach is being done for both methods and they both show bimodal distributions of marker genes.

We expect individual genes to have bimodal distributions (zero counts), but in the first example the distribution of the sums was also bimodal:
Screenshot 2024-06-18 at 6 58 30 PM

I might want look at the distribution of the sums _without_ the transformations though: that was what I meant by comparing between the samples to find a cutoff. Zero was nice when it worked, but with a different balance of tumor/normal it is probably showing its limits, and we will have to go back to something more "raw."

> I also want to note that I did try using [AUCell](https://bioconductor.org/packages/release/bioc/vignettes/AUCell/inst/doc/AUCell.html) to classify these cells with each of the gene sets and because of the lack of bimodal distribution, that didn't work very well. Only a few cells were classified as tumor and there was no clear cutoff.

I'd definitely be interested to see what some of your results were with AUCell as well. (Note that your marker gene list is also a potential gene set to look at). Again, I'm most curious about whether you might be able to set AUC cutoffs from 822 that might be able to be applied to other samples, without having to find a bimodal distribution of AUC in every data set. I'm not sure how consistent to expect the cutoffs to be, but it would be worth a look.
 
> I wonder if we could combine a celldex reference with the tumor cells from SCPCL000822 to create a custom reference? This way we would have more than the two classes. This might avoid the overfitting problem and allow for SingleR to hopefully better distinguish between potential normal cell types and then tumor cells.

This is kind of what I was thinking, using the multiple reference functionality of SingleR. I kind of expect that you could keep the "Normal" label in the Tumor/normal reference from 822, as it should not fit better than any of the more specific references, and could serve as a bit of a negative control. Or maybe you could use the SingleR calls for 822 "normal" cells, but replace all of tumor calls with the "tumor" label. Maybe try both.

allyhawkins:
@jashapiro I did quite a bit of reorganization and made a bunch of changes based both on feedback here and in #537. 
Rather than seeking to create final annotations here, I am exploring different methods for classifying these cells as tumor cells that rely solely on gene expression. 

- I kept the section that explores the results from marker genes, CopyKAT, and InferCNV to keep the commentary about these methods not all coming to an agreement. 
- Then the rest of the notebook focuses on looking at gene expression methods including marker genes, [AUCell](https://www.bioconductor.org/packages/release/bioc/vignettes/AUCell/inst/doc/AUCell.html) and [UCell](https://carmonalab.github.io/UCell_demo/UCell_matrix_vignette.html#content). 
- I also am utilizing information from SCPCL000822 to classify SCPCL000824 because of the lack of bimodal distributions seen in SCPCL000824. This really does make me think that most of the cells are tumor cells. 

[SCPCL000824_tumor-gene-expression-exploration.html.zip](https://github.com/user-attachments/files/15991874/SCPCL000824_tumor-gene-expression-exploration.html.zip)

> I might want look at the distribution of the sums without the transformations though: that was what I meant by comparing between the samples to find a cutoff. Zero was nice when it worked, but with a different balance of tumor/normal it is probably showing its limits, and we will have to go back to something more "raw."

Part of the notebook looks at the sum of the normalized gene expression counts for each marker gene and compares the distributions across both samples. I also used the information from 822 to define a cutoff to use for 824 since the distributions cover the same range of values but do not show the same bimodal distribution. 

> I'd definitely be interested to see what some of your results were with AUCell as well. (Note that your marker gene list is also a potential gene set to look at). Again, I'm most curious about whether you might be able to set AUC cutoffs from 822 that might be able to be applied to other samples, without having to find a bimodal distribution of AUC in every data set. I'm not sure how consistent to expect the cutoffs to be, but it would be worth a look.

I added a section where I looked at AUCell with both gene sets and marker genes. I think it does a nice job in 822 (I confirmed by comparing the annotations to our validated annotations). Because of the lack of bimodal distribution in 824 it doesn't do a very good job, but I used the AUC cutoff from 822 to classify 824 and it looked much more like what I would expect given individual marker gene expression. 

I also explored using `UCell` as recommended in https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/537#discussioncomment-9840032. Here we get a gene set signature score, similar to the score we are calculating on our own. With UCell you don't see quite the same clear bimodal distribution in the marker genes, but you do in 2 of the gene sets. 

> This is kind of what I was thinking, using the multiple reference functionality of SingleR. I kind of expect that you could keep the "Normal" label in the Tumor/normal reference from 822, as it should not fit better than any of the more specific references, and could serve as a bit of a negative control. Or maybe you could use the SingleR calls for 822 "normal" cells, but replace all of tumor calls with the "tumor" label. Maybe try both.

This notebook was getting quite long, so I'm going to file an issue to create a separate notebook that looks at using `SingleR` to classify these cells as a follow up to this notebook. 

> I did have a couple thoughts about file size issues too, before I go. One is that you could just change this to output .tsv.gz and things would shrink a lot. Alternatively you could write out files in some other format, like [parquet or arrow](https://arrow.apache.org/docs/r/articles/read_write.html). The disadvantage of all of these is that you loose the ability to easily diff the files, but for intermediate files that might be okay, and having them in the repo for easier review/inspection by outside users may be worth the tradeoff.

For right now, I am just saving this as a `.gz` file. I think you are right about having a larger discussion re: files that should be in the repo vs. not in the repo.

analyses/cell-type-ewings/cell_type_annotations/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,670 @@
...
+  dplyr::mutate(auc_classification = dplyr::if_else(barcodes %in% new_tumor_cells, "Tumor", "Normal"))
+```
+
+```{r}
+# visualize which cells are classified as tumor 
+ggplot(classification_df, aes(x = UMAP1, y = UMAP2, color = auc_classification)) +
+  geom_point(alpha = 0.5, size = 0.5)
+```
+
+This looks similar to using the updated marker gene cutoff to classify cells where most cells in that bottom group are tumor cells.
```
jashapiro, line 477: Just for completeness, do you want to do a confusion matrix for this AUC vs. previous version?

analyses/cell-type-ewings/cell_type_annotations/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,670 @@
...
+
+Here we will look at using [`UCell`](https://carmonalab.github.io/UCell_demo/UCell_matrix_vignette.html#content) to calculate a gene set score for each cell and then attempt to classify cells based on the distribution of those gene set scores. 
+Again, we will look at both `SCPCL000822` and `SCPCL000824`. 
+
+```{r}
+# create list to use for ucell of gene sets and marker genes
+ucell_gene_sets <- c(genes_list, markers = list(marker_genes))
+
+# run ucell on both ref (822) and 824
+ref_ucell <- UCell::ScoreSignatures_UCell(counts(ref_sce), features = ucell_gene_sets)
```
jashapiro, line 489: There were a lot of genes not found here and a warning about imputing them to zero. Is that something we should worry about? Is it better to exclude the missing genes?

I think it is also worth saying something about what the scores are for UCell... they seem to be rank-based?

analyses/cell-type-ewings/cell_type_annotations/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,670 @@
...
+  )
+```
+
+```{r}
+ggplot(ucell_df, aes(x = signature_score, colour = sample)) +
+  geom_density() +
+  facet_grid(rows = vars(gene_list))
+```
+
+Here we see that the marker gene set maybe has a small bimodal distribution in `SCPCL000822`, but it is very small and not quite as striking as looking at the scores from `AUCell` or raw gene expression.
```
jashapiro, line 522: I think this is an artifact of scaling. I would make set free y scales for easier comparisons here. To my squinty eye it actually looks like you are getting better results with the markers than any of the other gene sets.

analyses/cell-type-ewings/cell_type_annotations/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,670 @@
...
+classification_df <- classification_df |> 
+  dplyr::mutate(
+    ucell_classification = dplyr::if_else(barcodes %in% new_ucell_tumor_cells, "Tumor", "Normal")
+  )
+
+ggplot(classification_df, aes(x = UMAP1, y = UMAP2, color = ucell_classification)) +
+  geom_point(alpha = 0.5, size = 0.5)
+```
+
+Here we see less cells classified as tumor cells than we saw with the `AUCell` classification.
```
jashapiro, line 580:
```suggestion
Here we see fewer cells classified as tumor cells than we saw with the `AUCell` classification. 
```

analyses/cell-type-ewings/cell_type_annotations/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,670 @@
...
+  dplyr::mutate(
+    ucell_classification = dplyr::if_else(barcodes %in% new_ucell_tumor_cells, "Tumor", "Normal")
+  )
+
+ggplot(classification_df, aes(x = UMAP1, y = UMAP2, color = ucell_classification)) +
+  geom_point(alpha = 0.5, size = 0.5)
+```
+
+Here we see less cells classified as tumor cells than we saw with the `AUCell` classification. 
+This also looks very similar to the original UMAP when we classified cells using marker gene expression.
```
jashapiro, line 581: Whatever UCell is doing, it's score do still seem to be much more sample-dependent than the AUCell scores, making transfering thresholds less appealing. The AUCell scores seem to be more consistent between samples, but it is hard to see precisely without pulling out the raw scores and plotting them together. It would be nice to have similar plots for that section if you could, because the default AUCell plots are kind of busy. Which might be fine, but without an explanation of what all the fits and lines are, it is hard to know for sure!

jashapiro (commented):
Thanks for these updates! I like where this is going now, and I feel much better about some parts of it. I think the overall flow of analyses is pretty good now, with most of my comments now being about expanding the text to make it a bit more clear what scores and measures each part is actually using.

Most of my comments are in the line questions, but a few that I am wondering about:

- What is the "gene set score" for the marker sets that you are using?
- Does AUCell use all of the gene sets you give it for its scoring, or just the marker genes? If the former, it might be getting an unfair advantage.
- Can you add a brief explanation of what UCell is doing to create its score? Maybe one for AUCell as well?
- For AUCell, how much work is it to extract the data for the histograms? If not too much, can you use that data to create density plots that are more similar to your other comparisons between 822 and 824? I _think_ they look to be on similar scales, with 824 being mostly within the upper mode of 822, but it is very hard to do that comparison directly.

allyhawkins:
I made a number of small adjustments to address your comments @jashapiro: 

- I added an explanation of the gene set scores, we did that with the [`calculate-gene-set-scores.R`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-ewings/scripts/calculate-gene-set-scores.R) script and it's just the mean expression of all genes in a gene set. In this notebook and previous notebooks I use the non-scaled version of this. 
- I added some more explanatory text around `AUCell` and `UCell`. This included an attempt to explain the graphs produced from `AUCell`. 
- In regards to your comment about `AUCell` using all lists, it's my understanding that it calculates a threshold for each gene set independently so we have them for all three EWS-FLI1 gene sets and the marker gene list. I then use the one from the marker gene list to calculate the threshold for the other sample. 
- Along with that I also added density plots for the AUC distribution like we have for the others and compare the distribution across both samples. They are very similar which is what I expected. 
- I also added a few confusion matrices comparing both AUCell and UCell annotations to the updated marker gene annotations for SCPCL000824. 

Another thing I did here was move some things based on our conversation of not committing tables earlier. I moved all the notebooks for validating tumor cells to `exploratory_analysis/annotation_notebooks` and then all TSV files are now in `results/annotation_tables`. 

Updated notebook for review: 
[SCPCL000824_tumor-gene-expression-exploration.html.zip](https://github.com/user-attachments/files/16059486/SCPCL000824_tumor-gene-expression-exploration.html.zip)

analyses/cell-type-ewings/cell_type_annotations/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,670 @@
...
+```
+
+Looking at this, we see that `SCPCL000822` has a bimodal distribution and shows a clear cutoff where cells are tumor vs. normal, but this is not the case for `SCPCL000824`. 
+Let's find the local minima in the bimodal distribution for `SCPCL000822` and then use that to classify tumor cells in `SCPCL000824`. 
+
+```{r}
+# create distribution 
+density_data <- density(combined_markers_df$sum_raw_exp)
+# find the local minima in the distribution
+exp_cutoff <- optimize(approxfun(density_data$x, density_data$y), interval = c(1,10))$minimum
```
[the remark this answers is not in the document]
allyhawkins: I think if we actually decide we want to use this to create our final annotations then I would agree with you. But I think for the purpose of the analysis here I think this is fine. I am leaning towards using `AUCell` over just the sum of marker genes anyways.

analyses/cell-type-ewings/cell_type_annotations/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,670 @@
...
+
+```{r}
+# get new auc and plot thresholds 
+new_auc <- AUCell::AUCell_run(counts(sce), collection)
+new_assignment <- AUCell::AUCell_exploreThresholds(new_auc, assign = TRUE)
+```
+
+
+Here we do not see any bimodal distributions and that most cells are considered to not be expressing the specified gene sets, which doesn't fit with our previous findings. 
+I expect that without a bimodal distribution this model does not do a good job of calculating a threshold.
```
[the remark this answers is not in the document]
allyhawkins: From this [vignette](https://scenic.aertslab.org/scenic_paper/tutorials/AUCell.html#determine-the-cells-with-the-given-gene-signatures-or-active-gene-sets): 

> To ease the exploration of the distributions, the function AUCell.exploreThresholds() automatically plots all the histograms and calculates several possible thresholds for each gene-set (returned in $aucThr). The distributions are plotted as dotted lines over the histogram and the corresponding thresholds as vertical bars in the matching color. The thicker vertical line indicates the threshold selected by default ($aucThr$selected): the highest value to reduce the false positives.

I tried to explain this in the actual notebook as best as I could.

analyses/cell-type-ewings/cell_type_annotations/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
allyhawkins: Just noting that I did filter the gene sets to only those found in our index and I believe that removed the warnings.

analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,749 @@
...
+                      names_to = "barcodes", 
+                      values_to = "auc") |> 
+  #dplyr::filter(gene_set == "markers") |> 
+  dplyr::mutate(sample = "SCPCL000824")
+
+all_auc_df <- dplyr::bind_rows(list(ref_auc_df, new_auc_df))
+
+ggplot(all_auc_df, aes(x = auc, color = sample)) +
+  geom_density() +
+  facet_grid(rows = vars(gene_set))
```
jashapiro, line 493: Hopefully making the distributions easier to see
```suggestion
  facet_grid(rows = vars(gene_list),
             scales = "free_y")
```

analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,749 @@
...
+```
+
+Looking at the AUC across both samples for all gene sets, we see that they are in the same range of values and tend to mirror each other. 
+The main difference here is the lack of bimodal distribution in `SCPCL000824`.
```
jashapiro, line 497: I wouldn't necessarily say they mirror each other... The peak for 824 looks to be in between the two peaks for 822 in many cases, with the marker gene list being the exception where the peaks do line up. Which is a good argument that the marker gene list is a good set to use.

analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,749 @@
...
+
+ggplot(all_auc_df, aes(x = auc, color = sample)) +
+  geom_density() +
+  facet_grid(rows = vars(gene_set))
+```
+
+Looking at the AUC across both samples for all gene sets, we see that they are in the same range of values and tend to mirror each other. 
+The main difference here is the lack of bimodal distribution in `SCPCL000824`. 
+
+Let's use the threshold identified by `AUCell` for `SCPCL000822` to identify tumor cells in `SCPCL000824`.
```
jashapiro, line 499: Probably an unneccesary clarification given the next sentence.
```suggestion
Let's use the marker gene threshold identified by `AUCell` for `SCPCL000822` to identify tumor cells in `SCPCL000824`.  
```

analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,749 @@
...
+ggplot(all_auc_df, aes(x = auc, color = sample)) +
+  geom_density() +
+  facet_grid(rows = vars(gene_set))
+```
+
+Looking at the AUC across both samples for all gene sets, we see that they are in the same range of values and tend to mirror each other. 
+The main difference here is the lack of bimodal distribution in `SCPCL000824`. 
+
+Let's use the threshold identified by `AUCell` for `SCPCL000822` to identify tumor cells in `SCPCL000824`.  
+Here, we label any cells that have passed the threshold set for the marker gene sets as tumor cells.
```
jashapiro, line 500:
```suggestion
Here, we label any cells that have passed the threshold set for the marker gene set as tumor cells. 
```

analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,749 @@
...
+
+```{r}
+# visualize which cells are classified as tumor 
+ggplot(classification_df, aes(x = UMAP1, y = UMAP2, color = auc_classification)) +
+  geom_point(alpha = 0.5, size = 0.5)
+```
+
+This looks similar to using the updated marker gene cutoff to classify cells where most cells in that bottom group are tumor cells.  
+As a reminder, the updated marker gene classification is the classification of tumor cells based on the marker gene expression cutoff determined in `SCPCL000822`. 
+We can confirm this by calculating the confusion matrix.
```
jashapiro, line 524: Just noting that in the confusion matrix it is hard to tell which is which... You might rename one of them for clarity?

The reason I was interested was that I kind of wanted to know which one was more "conservative" as there does seem to be a pretty substantial difference. I'm not sure what I would do with that info though.

> Update: This is more visible in the later plots: AUCell seems to be the one that is a bit more conservative in this case, calling more cells as normal.

analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,749 @@
...
+caret::confusionMatrix(
+  table(
+    caret_df$tumor_cell_classification, 
+    caret_df$ucell_classification 
+  )
+)
+```
+
+It looks like we get pretty good agreement between using `UCell` with the `RIGGI` gene set to the original classification. 
+Now we will use the cutoff from `SCPCL000822` to classify tumor cells in `SCPCL000824`.
```
jashapiro, line 629:
```suggestion
Now we will use the marker gene set cutoff from `SCPCL000822` to classify tumor cells in `SCPCL000824`. 
```

jashapiro (commented):
This looks to be in good shape now, with most of my questions answered. I had a few more smaller clarification suggestions, but it should be pretty close.

analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
allyhawkins: I'm not entirely sure what you mean by this comment since the `gene_set` column exists in the data frame but `gene_list` does not? Each row already is a gene set?

jashapiro: A bad copy paste! I just wanted the `free_y`

analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000492/SCPCL000824_tumor-gene-expression-exploration.Rmd
allyhawkins: > Just noting that in the confusion matrix it is hard to tell which is which... You might rename one of them for clarity?

I don't believe we can do this since the classes are supposed to have the same names to be able to compare. I get an error about classes not being in the same order if I change any of the names of the classes. But I can add in clarification in the text.

allyhawkins:
@jashapiro I incorporated your text suggestions and made some minor additional edits based on your comments: 

- Added a link to the explanation in the `AUCell` vignette, although I agree it is not very clear there either. 
- I added some text around the confusion matrices to say which class is which. I also noticed the numbers were way too large for some of them and it appeared we had created an extra large classification data frame along the way so I fixed that. 
- I updated the explanation around the `AUCell` density plots and also used `scales=free_y`. While I was there I adjusted the names that are displayed for the marker gene sets to make the facet labels easier to see. 

Updated notebook:
[SCPCL000824_tumor-gene-expression-exploration.html.zip](https://github.com/user-attachments/files/16071281/SCPCL000824_tumor-gene-expression-exploration.html.zip)

jashapiro (approved):
Thanks for the clarifications! Looks good to me.

PR #536, opened 2024-06-18
Add token permissions to Docker build/push workflow

jaclyn-taroni:
We need to fetch a token for this workflow ([ref](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers#adding-permissions-settings)). I expect that would fix the error we saw in: https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/9569918033/workflow

### DIFF
--- a/.github/workflows/build-push-docker-module.yml
+++ b/.github/workflows/build-push-docker-module.yml
@@ -13,6 +13,10 @@ on:
         description: "Push to AWS ECR"
         type: boolean
 
+permissions:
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
Ah yes, that seems likely to be the thing I had missed.

jaclyn-taroni:
I didn't spot it in my review either!

PR #538, opened 2024-06-24
Docs: using docker images

### LINKED ISSUE #496: Add documentation for using docker images
This issue tracks adding a new documentation page (`using-images.md` or so) that describes how to use existing Docker images, including how to get access to pushed images and how to use obtain/use it to run a given module. This page should live in the _relocated_ `docker/` directory in the reproducibility section (#494).

We might also describe why this approach might be preferred and/or complementary to "just" using `renv` and/or `conda`.

sjspielman:
Closes #496

This PR adds documentation for using Docker images. To start, I stayed fairly thin on details for how to use Docker in the first place, under the assumption that contributors who actually want to pull images and run in containers have a sense of what they're doing already. I included a note to this end, and I also added a couple more Docker resources (links from [training materials](https://github.com/AlexsLemonade/training-specific-template/blob/7d68771d0c07ebc74507ccde1d934ed8a361edec/additional-resources/using-environments-post-workshop.md?plain=1#L218)). 

Let me know if there are any particular details that I glossed over which you think I should mention and/or emphasize more. For example, should I say more about `docker run` flags or volumes?

One item that this PR does _not_ cover but we should probably open an issue for is running docker images in LSfR, which may be necessary for certain modules depending on their compute needs. Do you agree we should circle back to this?

### DIFF
--- a/docs/ensuring-repro/docker/using-images.md
+++ b/docs/ensuring-repro/docker/using-images.md
@@ -20,7 +20,7 @@ Images are named `openscpca/{module-name}`.
 
 To obtain a local copy of a Docker image, follow these steps:
 
-1. Open the Docker Desktop application and ensure sure it is running.
+1. Open the Docker Desktop application and ensure it is running.
 2. From a [terminal](../../getting-started/project-tools/using-the-terminal.md), pull the image using this command, where `{module-name}` is replaced with the given Docker image you wish to pull:
 
       ```sh
--- a/docs/ensuring-repro/docker/using-images.md
+++ b/docs/ensuring-repro/docker/using-images.md
@@ -15,15 +15,13 @@ The OpenScPCA Docker register is publicly available at <https://gallery.ecr.aws/
 
 ## Obtaining Docker images
 
-<!-- TODO: Instructions for using images on LSfR? -->
-
 You can see all available OpenScPCA Docker images at this URL: <https://gallery.ecr.aws/openscpca/>.
 Images are named `openscpca/{module-name}`.
 
 To obtain a local copy of a Docker image, follow these steps:
 
 1. Open the Docker Desktop application and ensure sure it is running.
-1. From a [terminal](../../getting-started/project-tools/using-the-terminal.md), pull the image using this command, where `{module-name}` is replaced with the given Docker image you wish to pull:
+2. From a [terminal](../../getting-started/project-tools/using-the-terminal.md), pull the image using this command, where `{module-name}` is replaced with the given Docker image you wish to pull:
 
       ```sh
       docker pull public.ecr.aws/openscpca/{module-name}:latest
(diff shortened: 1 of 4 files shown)


docs/ensuring-repro/docker/using-images.md
```diff
@@ -0,0 +1,45 @@
...
+      - Follow these instructions to update your Docker Desktop settings for [Mac](https://docs.docker.com/desktop/settings/mac#resources) and [Windows](https://docs.docker.com/desktop/settings/windows/#resources) computers. You will need to click "Apply and Restart" to ensure these changes go through.
+
+4. You can now use the image to run code from the analysis module using [the `docker run` command](https://docs.docker.com/reference/cli/docker/container/run/) from a terminal.
+      - The analysis module's `README.md` file should contain documentation for what command(s) you need to issue to [run the module](../../contributing-to-analyses/analysis-modules/running-a-module.md).
+      - You will also need to [mount a volume](https://docs.docker.com/storage/volumes/) as part of your command to ensure that the Docker container has access to all files needed, including any data files.
+      - Note that containers for [images based on `bioconductor/r-ver`](./docker-images.md#r-based-images) can also be run in the browser as interactive RStudio Server sessions, [as described in the Bioconductor documentation](https://www.bioconductor.org/help/docker/).
```
allyhawkins, line 43: I'm wondering if it might make sense to show a minimal example of what running the image with mounting a volume looks like? I know we are assuming previous experience, but I do think it could be helpful to provide an example in one place. Something like: 

```
docker run -it --mount type=bind,source="path to local files",target="path to container files" <module-name>:latest
```
If you have a different preferred way of running a container then we could use that too, but I think an example could be helpful.

allyhawkins (commented):
Generally I think this looks good. I'm on board with the level of detail, but I do think at least some minimal example of the `docker run` command would be helpful with notes that link you to the Docker docs so contributors can customize to their needs. 

Also, I agree that using Docker on LSfR is something that can be added later. I may actually add that in the LSfR section rather than the Docker section and link to it from these pages.

sjspielman:
This should now be ready for another look! In addition to other responses to review, I've expanded the discussion of using `docker run` to cover more ground, and confirmed that the instructions I've written work for the pushed `simulate-sce` image.

docs/ensuring-repro/docker/using-images.md
```diff
@@ -0,0 +1,77 @@
...
+
+          ```sh
+          docker pull --platform linux/x86_64 public.ecr.aws/openscpca/{module-name}:latest
+          ```
+
+## Running Docker images
+
+### Update your resource settings
+
+Before running any Docker image, we recommend updating your Docker Desktop resource settings to ensure Docker has access to sufficient compute resources to run the module. For more information on module compute requirements, please refer to the [module's `README.md` file](../../contributing-to-analyses/analysis-modules/compute-requirements.md#readme-files).
```
allyhawkins, line 44:
```suggestion
Before running any Docker image, we recommend updating your Docker Desktop resource settings to ensure Docker has access to sufficient compute resources to run the module. 
For more information on module compute requirements, please refer to the [module's `README.md` file](../../contributing-to-analyses/analysis-modules/compute-requirements.md#readme-files).
```

docs/ensuring-repro/docker/using-images.md
```diff
@@ -0,0 +1,77 @@
...
+
+Follow these instructions to update your Docker Desktop settings for [Mac](https://docs.docker.com/desktop/settings/mac#resources) and [Windows](https://docs.docker.com/desktop/settings/windows/#resources) computers. You will need to click "Apply and Restart" to ensure these changes go through.
```
allyhawkins, line 46:
```suggestion
Follow these instructions to update your Docker Desktop settings for [Mac](https://docs.docker.com/desktop/settings/mac#resources) and [Windows](https://docs.docker.com/desktop/settings/windows/#resources) computers. 
You will need to click "Apply and Restart" to ensure these changes go through.
```

allyhawkins (approved):
LGTM!

PR #541, opened 2024-06-25
Force rebuild/push of doublet Dockerfile

sjspielman:
Last week, the GHA failed to push the `doublet-detection` Dockerfile to the registry due to a credentials issue. We think this _may_ have been fixed, so this PR is really just meant to see if it's fixed! With any luck, this image gets pushed to our registry upon merge. I modified the Dockerfile enough to trigger the GHA, but not enough to actually change the Dockerfile in a meaningful way.


allyhawkins (approved):
🚀 🤞

PR #543, opened 2024-06-26
Add update-symlink argument to data download script

### LINKED ISSUE #530: Update data download script: symlink treatment
Related to #523 

We'd like to update the `download-data.py` script to have an option where users can swap symlinks between test and release without having to re-download all the data.  

To implement this, we discussed an argument `--update-symlinks <test/current/a given release date>` to toggle the symlink. This argument should check to see if the data of interest exists locally and give some kind of informative message about the swap, but should not attempt to download fresh. We should write a small function to do the symlink toggling.

sjspielman:
Closes #530

This PR adds the `--update-symlink` argument to the `data-download.py` script, and it's working as expected on my end 👍 . I added a function `update_symlink()` which is called in two spots:
  - After downloading a data release (where the code originally was before I ported it into a function)
    - I also added a sentence here to print out how the symlink _would_ change if `--dryrun` is used 
  - Before any creds get used (to list releases), I include an `if args.update_symlink:` to update the symlink as desired. This is only done if the specified directory exists, and errors out otherwise with a message that you should download the data instead. I also use `sys.exit(0)` to exit out after changing the symlink, since we don't want to be doing any downloads with this flag.

Let me know if there are other behaviors we'd like to see for this flag. Also, I'm not convinced my help string for this argument is as clear as it could be, what do you think?

### DIFF
--- a/download-data.py
+++ b/download-data.py
@@ -152,19 +152,16 @@ def add_parent_dirs(patterns: List[str], dirs: List[str]) -> List[str]:
 
 
 def update_symlink(
+    data_dir: pathlib.Path,
     release: str
 ) -> None:
     """
     Update symlink when switching test and release data
     """
-    ### Update current link to point to new or test data if required ###
-    # only do this if we are using test data or the specified release is "current" or "latest", not for specific dates
-    if update_current and not dryrun:
-        # update the current symlink
-        current_symlink = data_dir / "current"
-        current_symlink.unlink(missing_ok=True)
-        current_symlink.symlink_to(release)
-        print(f"Updated 'current' symlink to point to '{release}'.")
+    current_symlink = data_dir / "current"
+    current_symlink.unlink(missing_ok=True)
+    current_symlink.symlink_to(release)
+    print(f"Updated 'current' symlink to point to '{release}'.")
 
 
 
@@ -263,7 +260,7 @@ def download_release_data(
     # only do this if we are using test data or the specified release is "current" or "latest", not for specific dates
     if update_current:
         if not dryrun:
-            update_symlink(release)
+            update_symlink(download_dir, release)
         else:
             print(f"\nIf this were not a dry run, the 'current' symlink would be updated to point to '{release}'.")
 
@@ -328,7 +325,7 @@ def main() -> None:
         "--test-data",
         action="store_true",
         help="Download test data from the test bucket and direct the `current` symlink to the test data directory."
-        " To switch back, rerun this script with the `--release current` option.",
+        " To switch back, rerun this script with either the `--release current` option, or the `--update-symlink release` option.",
     )
     parser.add_argument(
         "--metadata-only",
@@ -363,7 +360,12 @@ def main() -> None:
         default="",
         help="The AWS profile to use for the download. Uses the current default profile if undefined.",
     )
-
+    parser.add_argument(
+        "--update-symlink",
+        type=str,
+        default="",
+        help="Update the 'current' symlink to point to requested release without re-downloading data if it is already present."
+    )
     args = parser.parse_args()
 
     ### Validate the arguments ###
--- a/download-data.py
+++ b/download-data.py
@@ -171,7 +171,7 @@ def update_symlink(
     else:
         current_symlink = data_dir / "current"
         current_symlink.unlink(missing_ok=True)
-        current_symlink.symlink_to(target_path)
+        current_symlink.symlink_to(target)
         print(f"Updated 'current' symlink to point to '{target_path}'.")
 
 
@@ -374,7 +374,7 @@ def main() -> None:
         "--update-symlink",
         type=str,
         default="",
-        help="The release version to update the 'current' symlink to direct to. Release directory must be present. No data will be downloaded."
+        help="Update the 'current' symlink to point to the given release version. The release directory must be present. No data will be downloaded."
     )
     args = parser.parse_args()
 
@@ -456,10 +456,13 @@ def main() -> None:
 
     ### Update symlink if requested, without downloading data, if the target directory exists
     if args.update_symlink:
+        if args.update_symlink.casefold() in {"latest", "release"}:
+            # find the latest available release
+            ...
         if not args.dryrun:
             update_symlink(args.data_dir, args.update_symlink)
         else:
-            print(f"\nThe 'current' symlink would be updated to point to '{args.release}'.")
+            print(f"\nThe 'current' symlink would be updated to point to '{args.update_symlink}'.")
         sys.exit(0)
 
     ### List the available releases or modules ###
--- a/download-data.py
+++ b/download-data.py
@@ -472,7 +472,13 @@ def main() -> None:
 
         if target == "current":
             # find the most recent local release to use as target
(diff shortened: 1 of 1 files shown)


download-data.py
```diff
@@ -244,12 +258,12 @@ def download_release_data(
...
-        # update the current symlink
-        current_symlink = data_dir / "current"
-        current_symlink.unlink(missing_ok=True)
-        current_symlink.symlink_to(release)
-        print(f"Updated 'current' symlink to point to '{release}'.")
+    if update_current:
+        if not dryrun:
+            update_symlink(download_dir, release)
+        else:
+            print(f"\nIf this were not a dry run, the 'current' symlink would be updated to point to '{release}'.")
```
jashapiro, line 265: I think we can be more compact here. 
```suggestion
            print(f"\nThe 'current' symlink would be updated to point to '{release}'.")
```

download-data.py
```diff
@@ -346,7 +360,12 @@ def main() -> None:
         default="",
         help="The AWS profile to use for the download. Uses the current default profile if undefined.",
     )
-
+    parser.add_argument(
+        "--update-symlink",
+        type=str,
+        default="",
+        help="The release version to update the 'current' symlink to direct to. Release directory must be present. Data will not be re-downloaded if it is already present."
```
jashapiro, line 367: This seemed to imply that data would be downloaded if not present.
```suggestion
        help="The release version to update the 'current' symlink to direct to. Release directory must be present. No data will be downloaded."
```

jashapiro (commented):
This looks mostly good, with a few wording suggestions and a suggestion to move the presence check to inside the function.

sjspielman:
I moved the check into the function, flipped the logic order, and also added some code to also print what would happen under the `--update_symlink <arg> --dryrun` flag combo.

download-data.py
```diff
@@ -346,7 +370,12 @@ def main() -> None:
         default="",
         help="The AWS profile to use for the download. Uses the current default profile if undefined.",
     )
-
+    parser.add_argument(
+        "--update-symlink",
+        type=str,
+        default="",
+        help="The release version to update the 'current' symlink to direct to. Release directory must be present. No data will be downloaded."
```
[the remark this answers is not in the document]
jashapiro: Do we want to support "latest" or a generic "release" value? I think we should! Let's not make people remember all of their release versions, and add a bit of logic to find the latest release.

download-data.py
```diff
@@ -424,6 +453,15 @@ def main() -> None:
             file=sys.stderr,
         )
 
+
+    ### Update symlink if requested, without downloading data, if the target directory exists
+    if args.update_symlink:
+        if not args.dryrun:
```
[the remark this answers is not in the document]
jashapiro: Note that I also just learned about `.casefold()` which is a more robust alternative to `.lower()` for unicode: we could replace a number of calls to `.lower()` if we want to...

sjspielman:
I've updated the code to reflect the behavior described in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/543#pullrequestreview-2146165879, and updated a few `.lower()` spots to `.casefold()`.
Any particular feedback on shorter ways to list directories, let me know. In listing local directories, I made an assumption (too strong?) that there is nothing in `data` besides release folders, but I could also list out the expected potential `regex`s for data folder names and find which are present?

Note that I test with these flags:
```
# works as expected
./download-data.py --update-symlink  
./download-data.py --update-symlink  --release 2024-05-01
./download-data.py --update-symlink  --release 2024-03-11 
./download-data.py --update-symlink  --release test
./download-data.py --update-symlink  --test-data

# fails as expected
./download-data.py --update-symlink --release 2024-10-11
```

download-data.py
```diff
@@ -424,6 +461,37 @@ def main() -> None:
...
+        if not target in local_releases:
+            if args.dryrun:
+                print(f"\nThe '{target}' data release is not locally available. The symlink would not be updated.")
+            else:
+                print(f"\nThe '{target}' data release is not locally available. Could not update symlink.")
+            sys.exit(1)
+        else:
+            if target == "current":
+                # find the most recent local release
+                dated_releases = [x for x in local_releases if re.match("\d{4}-\d{2}-\d{2}", x)]
```
jashapiro, line 485: move the `local_releases` here, because we don't need to look at all except when "current" is used. You can probably eliminate the variable completely by combining the two comprehensions:
```suggestion
                dated_releases = [x.name for x in args.data_dir.iterdir() if x.is_dir and re.match("\d{4}-\d{2}-\d{2}$", x.name)]
```

download-data.py
```diff
@@ -424,6 +461,37 @@ def main() -> None:
             file=sys.stderr,
         )
 
+
+    ### Update symlink if requested, without downloading data, if the target directory exists
+    if args.update_symlink:
+
+        # find all non-hidden directories in data_dir, representing the local releases
+        local_releases = [x.stem for x in args.data_dir.iterdir() if re.match("^[^.]", x.name) and x.is_dir()]
```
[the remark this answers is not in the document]
jashapiro: update: see below to remove/combine this.

jashapiro (commented):
I think we can simplify a but further! You are doing a number of duplicated checks, so I think we can rely on the checks in the function and eliminate a lot of extra nesting and more complex manipulation.

sjspielman:
:bowtie: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/543/commits/716f981048c98e2dc28b0baa78c813c643090c93

download-data.py
```diff
@@ -424,6 +461,26 @@ def main() -> None:
...
+
+        if args.test_data:
+            target = "test"
+        else:
+            target = args.release
+
+        if target == "current":
+            # find the most recent local release to use as target
+            dated_releases = {x.name for x in args.data_dir.iterdir() if x.is_dir and re.match("\d{4}-\d{2}-\d{2}$", x.name)}
+            most_recent = sorted(dated_releases)[-1]
```
jashapiro, line 476: Since we are not using `in` or any other set operation, no need for a set here. (But since we only use it once, we can use a generator)
```suggestion
            dated_releases = (x.name for x in args.data_dir.iterdir() if x.is_dir and re.match("\d{4}-\d{2}-\d{2}$", x.name)
            most_recent = max(dated_releases)
```

download-data.py
```diff
@@ -424,6 +461,26 @@ def main() -> None:
...
+    if args.update_symlink:
+
+        if args.test_data:
+            target = "test"
+        else:
+            target = args.release
+
+        if target == "current":
+            # find the most recent local release to use as target
+            dated_releases = (x.name for x in args.data_dir.iterdir() if x.is_dir and re.match("\d{4}-\d{2}-\d{2}$", x.name)
```
jashapiro, line 475: I realized we probably want to check the date here too. (I didn't mean to send the last comment yet!)
```suggestion
            dated_releases = (
              x.name
              for x in args.data_dir.iterdir()
              if x.is_dir()
              and re.match("\d{4}-\d{2}-\d{2}$", x.name)
              and x.name <= datetime.date.today().isoformat()
            )
```

jashapiro (approved):
The last comment wasn't fully complete and had some errors, but I think what I have now should work...

With that change and local testing, this should be good to go.

sjspielman:
Noting that I ran the script locally with a variety of arguments, and it works as expected

PR #545, opened 2024-06-26
Bump mkdocs version and fix some link anchors

sjspielman:
I recently updated my local version of `material-mkdocs`, and found that in the current version I have locally, serving the docs found several bad anchor links which had not been flagged previously.

This PR fixes those anchor links and also bumps the `mkdocs` versions in the `docs/requirements.txt` file that we use when running the GHA to check links (though maybe I was a bit too prescriptive with the patch version here...). To test if the GHA now flags bad anchor links, I added an intentionally bad one so hopefully the links action fails 🤞

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -29,7 +29,7 @@ jobs:
 
           # Capture all the bad links (don't let grep exit -1 if no match)
           grep "relative link" build-log.txt > bad_links.txt || true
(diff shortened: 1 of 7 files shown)


sjspielman:
This PR has gotten a bit out of hand with all my commit reversions to test whether the `mkdocs-material` bump is indeed necessary (it is!), so I'm going to close this and open a fresh PR with a cleaner history!

PR #533, opened 2024-06-18
Add Docker GHA docs

sjspielman:
Closes #518 
Closes #514 

Here, I'm adding docs for the docker building GHA as well as the index file (which is pretty brief) for this section. I made a few wording tweaks the module GHA docs as well for consistency, and also there was one part that was simply wrong/bad....

Originally, I had planned to also mention something in these docs about the order of operations for updating modules and docker images. But, it seemed to me that that information made better sense in #497 (where we already plan to write about it anyways), and it didn't quite fit into the overall logic of this page, which I ended up writing to largely mirror the module GHA page. This page has somewhat fewer details because, well, I think this GHA has fewer details to communicate! Please let me know which missing details should make the cut.

PR #546, opened 2024-06-26
Bump mkdocs and update links action

sjspielman:
I recently updated my local version of `material-mkdocs`, and found that this updated version flags bad _anchor links_ which had not been flagged previously. After playing around with this a bit in the now-closed #545, I realized that in addition to bumping `mkdocs-material`, I also needed to update the `check-links.yml` GHA itself to also grab anchor link warning messages. This GHA indeed failed as expected: https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/9683208495/job/26718098940

This PR takes care of all these items:
- Fixes the bad anchor links we had missed
- Bumps the `mkdocs-material` and `mkdocs` versions to versions that I know actually tell you about bad anchor links
  - Let me know if I was a bit too prescriptive with the `mkdocs-material` patch version here? 
- Updates the GHA to catch bad anchor links as well as bad relative links

Edit: One additional question I have is whether we actually don't want the links action to be _required_ to pass before merging into `main`. During docs development, we may actually expect some `#STUB_LINKS`.  We don't necessarily want that to hold up merging into a feature branch like `reproducibility-docs`, but we definitely don't want bad links going into `main`. I don't have the permissions to make this change though, so if we want to make it we can tag in someone else who can add that protection rule.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -29,7 +29,7 @@ jobs:
 
           # Capture all the bad links (don't let grep exit -1 if no match)
           grep "relative link" build-log.txt > bad_links.txt || true
-          grep "no such anchor" build-log.txt > bad_links.txt || true
+          grep "no such anchor" build-log.txt >> bad_links.txt || true
 
           # Save the number of bad links to github output
           echo "n_bad_links=$(wc -l < bad_links.txt)" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -29,6 +29,7 @@ jobs:
 
           # Capture all the bad links (don't let grep exit -1 if no match)
           grep "relative link" build-log.txt > bad_links.txt || true
+          grep "no such anchor" build-log.txt > bad_links.txt || true
 
           # Save the number of bad links to github output
           echo "n_bad_links=$(wc -l < bad_links.txt)" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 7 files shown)


allyhawkins:
> Edit: One additional question I have is whether we actually don't want the links action to be required to pass before merging into main. During docs development, we may actually expect some #STUB_LINKS. We don't necessarily want that to hold up merging into a feature branch like reproducibility-docs, but we definitely don't want bad links going into main. I don't have the permissions to make this change though, so if we want to make it we can tag in someone else who can add that protection rule.

I would agree that we don't want things that are broken being merged into `main`, so I would keep the check there. I think if you want to make the check run but don't require it to pass on feature branches then we do need to tag in someone to update branch protection. I assume it's turned on for `reproducibility-docs` given merging being blocked here, so maybe we adjust to run the CI but it doesn't have to pass CI until it is merged into `main`?

sjspielman:
> I assume it's turned on for reproducibility-docs given merging being blocked here, so maybe we adjust to run the CI but it doesn't have to pass CI until it is merged into main?

Yes, this is what I was thinking!

sjspielman:
Noting that the failure in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/546/commits/7192b2396482be684014558363fb3724ec48f279 is for an _expected_ stub link which can be updated once #533 goes in.

allyhawkins (approved):
This looks good to me, but we just need to update branch protection rules or wait for #533

allyhawkins:
> This looks good to me, but we just need to update branch protection rules or wait for #533

Actually it looks like checks aren't required to pass so we are good to go.

sjspielman:
> Actually it looks like checks aren't required to pass so we are good to go.

Ah, indeed! I was just on autopilot assuming they were 😬

PR #548, opened 2024-06-27
Set Renv not to activate in docker images

### LINKED ISSUE #542: Packages not found in hello-r docker image
While working on #538, I went to test out running the `hello-python` and `hello-R` docker images, and had difficulty running the analysis in the the `hello-R` image. The image built fine, but there were issues with `renv` - 

I built the image and launched a container with:
```
docker buildx build --platform linux/amd64 --tag hello-r:latest .

docker run --rm -it --platform linux/amd64 --mount type=bind,source=/Users/sjspielman/ALSF/open-scpca/OpenScPCA-analysis,target=/OpenScPCA-analysis hello-r:latest bash
```

Then in the container:
```
cd OpenScPCA-analysis/analyses/hello-R

bash run_hello.R
```

Which led to:

```
- One or more packages recorded in the lockfile are not installed.
- Use `renv::status()` for more details.
Error in loadNamespace(x) : there is no package called ‘rmarkdown’
Calls: loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted
```

---

Packages were indeed installed from the `renv.lock` file as expected, but because of this line in the Dockerfile, there is no `renv` cache that knows about it. 
https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/8ec8fa0dde4670bc440fefa58cf8d3e75aa6b984/analyses/hello-R/Dockerfile#L12-L13
As a consequence, when this script tries to load up the `renv` environment since we still have `.Rprofile` activating it, we end up with errors that packages aren't installed.

Off the top of my head, here's one not-awful idea and one bad idea, and one idea whose quality I'm unsure of, to address this...
- Not-awful: We could add a line to the Dockerfile to remove the module's `.Rprofile` file so that `renv` doesn't get activated. 
- Bad: We could add `Rscript -e 'renv::restore()'` into `run_hello-R.sh`. This seems like a not great idea, since we'd have to maintain something like this in every R module.
- Unsure: We could just not disable the cache

jashapiro:
Closes #542

Here I added a bit of simple logic to avoid running `renv/activate.R` when we are in a Docker image.  To do this I added an ENV variable named `OPENSCPCA_DOCKER` to all docker images, and then only source the activate script if the value is not set to `TRUE`. 

The original issue isn't so much that related to caching in the docker images as that the files are being installed in the system library, so we don't want renv to be activated and looking for its own versions of the packages. If we were to do that, we would almost certainly run into path problems with files stored or not in the host filesystem.

I tested this in the hello-R image/script, but applied the same changes everywhere, including in the `create-analysis-module.py` script to cover future modules. I also included the ENV variable in the template, and in the hello-python module Dockerfile, even though it may not be needed there. Seemed a good place for just a bit of extra consistency.

### DIFF
--- a/analyses/cell-type-ewings/.Rprofile
+++ b/analyses/cell-type-ewings/.Rprofile
@@ -1,1 +1,4 @@
-source("renv/activate.R")
+# Don't activate renv in an OpenScPCA docker image
+if(Sys.getenv('OPENSCPCA_DOCKER') != 'TRUE'){
+  source('renv/activate.R')
+}
--- a/analyses/cell-type-ewings/Dockerfile
+++ b/analyses/cell-type-ewings/Dockerfile
@@ -8,6 +8,9 @@ LABEL org.opencontainers.image.description="Docker image for the OpenScPCA analy
 LABEL org.opencontainers.image.authors="OpenScPCA scpca@ccdatalab.org"
 LABEL org.opencontainers.image.source="https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-ewings"
 
+# Set an environment variable to allow checking if we are in an OpenScPCA container
+ENV OPENSCPCA_DOCKER TRUE
+
 # Disable the renv cache to install packages directly into the R library
 ENV RENV_CONFIG_CACHE_ENABLED FALSE
 # set a name for the conda environment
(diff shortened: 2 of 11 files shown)


sjspielman (approved):
LGTM! Noting I successfully tested this in the following ways:
- built the `hello-R` docker image and ran the module from the container
- confirmed the new `.Rprofile` code works as expected with the `OPENSCPCA_DOCKER` environment variable
- confirmed that `create-analysis-module.py --use-renv` creates the appropriate `.Rprofile`

allyhawkins (approved):
LGTM! I also tested by building the hello-R image and running the analysis module from inside the image.

PR #549, opened 2024-06-27
Add docker workflows for hello modules

jashapiro:
While we have the "hello" modules in the `docker_all-modules.yml` action, they don't currently have their own actions to test and build Docker changes in those modules themselves, so I thought we should make sure we have that as well. 

I meant to put this as part of #542, but it slipped my mind until after merging!


allyhawkins (approved):
LGTM

PR #551, opened 2024-06-28
Add Docker workflow for ewings cell type module

allyhawkins:
Here I'm adding the Docker GHA for building and pushing the image for `cell-type-ewings`. I just copied the same file we have for the other modules to build and push the individual files and replaced the module name. I also added this to the `docker_all-modules.yml` file, but let me know if I should remove it.


jashapiro (approved):
There is currently a problem with GHA (https://www.githubstatus.com/), so you may need to rerun actions before merging, but this looks good to me!

Issue #529, opened 2024-06-17
Update step order in module GHAs


sjspielman:
> We may want to reverse the order of these two, when I think about it. Mostly because environment setup tends to be the harder part so we would rather fail with that first.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/528#discussion_r1643084639_
            
---

This comment was specifically about swapping the order of downloading test data and setting up the environment, so that the environment is set up first. 

When making this change, we should update the following:
- module GHA docs
- existing module GHAs
- template module GHA

sjspielman:
Closed by #562

PR #556, opened 2024-07-02
Add module GHA template for conda/renv module

### LINKED ISSUE #526: Add template module GHA for renv+conda modules
We should consider adding a GHA template file for modules that use _both_ `renv` and conda. ~This template could then be used when contributors specify both `renv` and conda/python flags to `create-analysis-module.py`.~

Edit: For this issue, we'll focus only on creating the template GHA file. In the future, we can decide specifically what to do with it, namely how much we want to actively surface the existence and usage of this template to users, and including whether we should add it to the analysis creation script.

sjspielman:
Closed #526, and does what it says! I derived this by working backwards to generalize my `doublet-detection` workflow.
Let me know if I should also file an issue for _potential_ future steps we may take to use this template.


templates/workflows/run_conda-renv-module.yml
```diff
@@ -0,0 +1,69 @@
...
+# - Manual trigger
+# - On pull requests where code in the module has changed
+# - As a reusable workflow called from a separate workflow which periodically runs all modules
+#
+# At initialization, only the manual trigger is active
+
+name: Run {{ openscpca_module }} analysis module
+env:
+  MODULE_PATH: analyses/{{ openscpca_module }}
+  AWS_DEFAULT_REGION: us-east-1
```
jashapiro, line 13: I don't know that it matters, but all of the OpenScPCA data is in `us-east-2` so we should probably use that. Which means this should probably be updated in the other templates too. 
```suggestion
  AWS_DEFAULT_REGION: us-east-2
```

jashapiro (approved):
LGTM, with some very minor edits.

templates/workflows/run_conda-renv-module.yml
sjspielman: noting I tacked this onto https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/529

sjspielman: Just kidding, I just went ahead and did it here: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/556/commits/e0aaf61c1afa2b0722d8e46d5419df109cb3f321

PR #557, opened 2024-07-02
Add docker on LSfR instructions

### LINKED ISSUE #539: Add documentation for using docker on LSfR
We should add a page with instructions for how to use docker containers on LSfR. We know that Docker is installed already on LSfR, but we need to investigate its usage further (e.g., is it a GUI or only CLI?)

sjspielman:
Closes #539 

This PR adds instructions for using docker (CLI only) on LSfR instances. I added a small section to describe usage (which is basically the same as locally, since we present `docker pull` and `docker run` already), and I linked to this section from the AWS index page. Let me know if there are any other details/caveats/considerations I should add here.

Note that, per https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/539#issuecomment-2203620553, we do _not_ expect `sudo` to be needed here.

### DIFF
--- a/docs/aws/index.md
+++ b/docs/aws/index.md
@@ -33,3 +33,4 @@ Please refer to these pages to learn more about setting up and using LSfR:
 - [Developing on virtual computers](./lsfr/starting-development-on-lsfr.md)
     - This page also provides instructions for setting up virtual computers after they've been created.
 - [Working with snapshots](./lsfr/working-with-snapshots.md)
+- [Using Docker images on virtual computers](../ensuring-repro/docker/using-images.md#using-docker-images-on-virtual-computers)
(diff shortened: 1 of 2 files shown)


allyhawkins (commented):
I don't have any additional comments here, but I do think we should maybe wait to ensure everything works as expected based on https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/539#issuecomment-2205974687 before approving and merging.

sjspielman:
@allyhawkins, `docker` is working as expected! See https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/539#issuecomment-2206149951

allyhawkins (approved):
LGTM

PR #561, opened 2024-07-03
560 Error in downloading test data using download-data.py

### LINKED ISSUE #560: Error in downloading test data using download-data.py
### Describe the bug

In attempting to download test data in the GHA for running the `cell-type-ewings` module, I get an error that it can't update the symlink because no data is downloaded. 

Screenshot 2024-07-03 at 8 56 45 AM

### Steps to reproduce

This is the command that is being run `./download-data.py --test-data --format "sce,anndata" --samples SCPCS000490`. 

Based on the description of arguments in the script I should still be able to download test data this way, but that doesn't appear to be the case.

### Consequences of the bug

GHA is currently failing and I cannot download test data for the first time. 

### Proposed solution

_No response_

jashapiro:
The error described in #560 appeared to be just in the symlink updating, not the download itself. I believe I fixed it here. with the change on line 269 in the new file

All other changes are automatic `ruff` formatting.

closes #560

### DIFF
--- a/download-data.py
+++ b/download-data.py
@@ -43,7 +43,9 @@ def get_releases(bucket: str, profile: str, test_data: bool) -> List[str]:
     # get only date-based versions and remove the trailing slash
     date_re = re.compile(r"PRE\s+(20\d{2}-[01]\d-[0123]\d)/$")
     return [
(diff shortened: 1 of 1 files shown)


sjspielman (approved):
Thanks for catching this! Code in this branch works for me to download test data

PR #562, opened 2024-07-03
Download data after environment setup in module GHAs

### LINKED ISSUE #529: Update step order in module GHAs

sjspielman:
Closes #529 

This PR updates workflows and associated docs (which currently live only in the `reproducibility-docs` branch, hence that is the target branch here) to download data after environment setup. I also updated a couple small spots in workflows, both existing workflows and templates, based on the reviews in #556. 

Any other spots to update that I missed here? 

(Incidentally, I noticed we also don't have a GHA to run the `simulate-sce` module. Did we want one?)

### DIFF
--- a/.github/workflows/run_doublet-detection.yml
+++ b/.github/workflows/run_doublet-detection.yml
@@ -55,7 +55,6 @@ jobs:
             lzma-dev \
             libglpk-dev
 
-
       - name: Set up pandoc
         uses: r-lib/actions/setup-pandoc@v2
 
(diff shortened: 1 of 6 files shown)


jashapiro (approved):
LGTM

Issue #565, opened 2024-07-05
Create a template report for evaluating tumor cell annotations


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#292 
Step 2 to address #563 

### Describe the goals of the changes to the analysis module.

Once we have annotated tumor cells in a sample, we want to have some sort of output that we can use to evaluate those annotations. We should make a template report that will be run after the script created in #564 that will output some basic plots to validate our findings. 

### What will your pull request contain?

A template report that ultimately will be part of a workflow to run #564 and generate this report for each sample in SCPCP000015. 

I think we will want to include: 

- A UMAP showing cells as tumor or normal 
- If using `AUCell` as I anticipate we will, a density plot showing the distribution of the AUC in the query sample and in the ref sample (SCPCL000822). 
- Density plots showing marker gene expression and gene set score distributions in tumor and other cells. 
- Heatmaps showing marker gene expression and gene set scores in individual cells. Here we expect to see tumor and normal cells clustering together with distinct expression/ score profiles. 

### Will you require additional software beyond what is already in the analysis module?

No

### Will you require different computational resources beyond what the analysis module already uses?

No

PR #569, opened 2024-07-08
Script for running AUCell on an individual SCE object

### LINKED ISSUE #564: Identify a method and write a script to annotate tumor cells in SCPCP000015

#292 
Step 1 to address #563 

Here we will identify a method that we feel comfortable with using to annotate tumor cells in all samples in SCPCP000015. Then we will write a script that runs that method on a single SCE object and outputs annotations. 

Based on findings in #532 and #558 I think we have two options: 

1. Use [`AUCell`](https://www.bioconductor.org/packages/release/bioc/html/AUCell.html) with the pre-defined threshold calculated for `SCPCL000822` and the marker gene list as the gene set. 
2. Use `SingleR` with the combination of the `SCPCL000822` reference and `BlueprintEncodeData` reference. 
3. Use both of these methods and find the consensus between them. Any cells labeled as tumor in one and not the other would be labeled as ambiguous. 

I personally am leaning towards using `AUCell`, because it utilizes a pre-defined set of tumor marker genes that we expect to be present in all tumor cells. `SingleR` uses tumor cells from one sample to define tumor cells in the rest of the samples which is probably fine for the most part, but given the heterogeneity of Ewing sarcoma, I am nervous we might miss some tumor cells. Although we are using a threshold from a single sample to define the auc cutoff, so both methods are somewhat biased. 

A script to run the identified method on a single SCE object. This will ultimately be part of a workflow that runs the method on all samples in SCPCP000015. 

No

No

_No response_

allyhawkins:
Closes #564 

The goal of this PR is to add a script that can be used to run `AUCell` on a single SCE object with a specified AUC threshold. The script added here will be followed up with a template notebook (see #565) and combined into a single workflow. The idea is that we can then use this workflow to run `AUCell` on every sample in SCPCP000015. 

I chose to use `AUCell` over just the sum of marker genes or `UCell` because the distributions seemed to be more consistent across samples making it easier to use a threshold from one heterogeneous sample to define tumor cells in a more homogenous sample. I also think this is a good place to start and if we see lots of wonky results then we can update the method used in the script but at least this gets set up to start trying something. 

I wanted to be able to use an AUC threshold that was determined in a sample that had a nice bimodal distribution, like SCPCL000822. Because of this, I am actually adding two scripts. 

- The first script is to identify an AUC threshold to use. This takes a reference SCE and a list of marker genes, runs `AUCell`, pulls out the thresholds, and then saves that value to a file. 
- The second script takes in a SCE file, marker gene list, and an optional AUC threshold to use for classification. `AUCell` is run using the marker gene list and if the AUC threshold is provided, then that threshold is used to classify cells. The results are then saved as a TSV file with `barcodes`, `auc`, and `auc_classification`. 

In designing these scripts, I was thinking that we would have a workflow that takes the following steps: 
1. Identifies a threshold for AUC given a reference SCE file, in our case this will be `SCPCL000822`. The AUC value will be saved in the scratch directory to be used by the next script. Optionally, we could just have the script return the value without saving a file? 
2. Then we run the `01-run-aucell.R` on all samples, with the `--auc_threshold` from step 1.

Yes 

This shouldn't need anything special and can just be run locally in the Rproj. 

How do we feel about having two separate scripts here? I thought about trying to combine them and adding flags to help specify if you should save annotations or output a threshold, but in the workflow we are going to run the first script once and the second script multiple times. That's what led me to using two scripts, but I could really see it both ways. 

In doing this I noticed that the main README is really focused on the original workflow that we created to annotate tumor cells with `CellAssign` and CNV methods. I think I want to do some rearranging of docs there, and maybe some reorganization of where scripts live, but I didn't want to include that in this PR. I'm going to file an issue to do that separately. I just wanted to point out that I do plan on making further documentation changes as we get this workflow set up. 


analyses/cell-type-ewings/scripts/00-identify-ref-aucell.R
```diff
@@ -0,0 +1,79 @@
...
+
+# This script is used to identify a reference threshold to use for classifying cells with AUCell
+# The reference SCE is run through AUCell using the marker gene list 
+# The auc threshold output by AUCell for this sample is returned as the output
+
+project_root <- here::here()
+renv::load(project_root)
+
+library(optparse)
+library(SingleCellExperiment)
```
jashapiro, line 11: Minor, but I would might loading `SingleCellExperiment` until after the option parsing for faster failures (and maybe wrap the loading with `suppressPackageStartupMessages()`)

analyses/cell-type-ewings/scripts/00-identify-ref-aucell.R
```diff
@@ -0,0 +1,79 @@
...
+  auc_results, 
+  assign = TRUE, 
+  plotHist = FALSE
+  )
+
+# pull out threshold used for assigning cells from ref sample 
+ref_threshold <- auc_assignments$markers$aucThr$selected
+
+# save threshold to output file 
+writeLines(ref_threshold, opt$output_file)
```
jashapiro, line 78: Just checking that this is a single number? Is there anything we should do with regard to formatting?

analyses/cell-type-ewings/scripts/01-run-aucell.R
```diff
@@ -0,0 +1,104 @@
...
+    opt_str = c("--marker_genes_file"),
+    type = "character",
+    default = file.path(project_root, "references", "tumor-marker-genes.tsv"),
+    help = "Path to file containing all marker genes used to run AUCell. 
+      Must contain `cell_type` and `ensembl_gene_id` columns.
+      Any marker genes where `cell_type` is 'Tumor' will be used."
+  ),
+  make_option(
+    opt_str = c("--auc_threshold"),
+    type = "double",
```
jashapiro, line 30: Just noting that you are reading this in not from a file, but as a raw value. Is the assumption that you will use something like `--auc_threshold $(cat threshold_file.txt)` in a script to feed in the value?

analyses/cell-type-ewings/scripts/00-identify-ref-aucell.R
```diff
@@ -0,0 +1,79 @@
...
+    type = "character",
+    default = file.path(project_root, "references", "tumor-marker-genes.tsv"),
+    help = "Path to file containing all marker genes used to run AUCell. 
+      Must contain `cell_type` and `ensembl_gene_id` columns.
+      Any marker genes where `cell_type` is 'Tumor' will be used."
+  ),
+  make_option(
+    opt_str = c("--output_file"),
+    type = "character",
+    help = "Path to file where AUC threshold will be stored."
```
jashapiro, line 31: Maybe something about the format of this file?

analyses/cell-type-ewings/scripts/01-run-aucell.R
```diff
@@ -0,0 +1,104 @@
...
+
+# This script is used to classify tumor and normal cells using AUCell with a marker gene list 
+# The threshold for defining cells as tumor cells should be determined separately by running 
+# `00-identify-ref-aucell.R`
+
+project_root <- here::here()
+renv::load(project_root)
+
+library(optparse)
+library(SingleCellExperiment)
```
jashapiro, line 11: Same comment as earlier about failing before loading this big package.

analyses/cell-type-ewings/scripts/README.md
```diff
@@ -143,6 +145,44 @@ Rscript run-singler.R \
...
+
+The output will be a TSV file containing the AUC value determined by `AUCell` and the classification (tumor or normal).
+
+By default, tumor marker genes are identifed from the genes listed in `references/tumor-marker-genes.tsv`.
+To use a different file, you can use the `--marker_genes_file` option.
+
+To run this script use the following command:
+
+```sh
+Rscript 00-identify-ref-aucell.R \
```
jashapiro, line 180:
```suggestion
Rscript 01-run-aucell.R \
```

analyses/cell-type-ewings/scripts/README.md
```diff
@@ -143,6 +145,44 @@ Rscript run-singler.R \
...
+  --sce_file <path to reference sce file> \
+  --auc_threshold <AUC value required to be classified as a tumor cell> \
```
jashapiro, line 182: I think you want to describe how to get the value from the file to here?

jashapiro (commented):
This looks good overall, but I had a few questions that most likely can be addressed with some additional comments. Mostly this is about how these two scripts will functionally work together such that the output of one would be fed into the next in a workflow, since the output of the first is a file but the input is listed as a number.

I suggested a process substitution option, which would be fine, but maybe you had something else in mind, like reading it into a shell variable. That would be fine too, but it makes me think that maybe the default output for the first script could just be to `STDOUT`? Alternatively, maybe just make the input value a file as well (and then maybe just store as rds to save converting back and forth)?

allyhawkins:
@jashapiro I was thinking about this more and testing it with a mini shell script (mainly to get results to use for working on #565) and I decided to merge this into one script. I am now just using the `01-run-aucell.R` script and have added a flag there to return the AUC value, `--return_auc`. If this is true then the AUC value will be printed to `stdout`. 

The reason I did this is because I actually need the AUC results from `01-run-aucell.R` for the reference also, so it didn't make sense to have to run AUCell twice on the reference, especially when most of the code is the same. 

For some additional context, this is kind of what I'm thinking as far as implementation: 

1. First run for reference sample and save auc value to bash variable 
2. Loop through all samples in project folder 
  - If sample is the reference sample then skip it
  - All other samples are run through `01-run-aucell.R` using the specified auc threshold output from step 1

Let me know what you think.

jashapiro (approved):
This looks good to me. I think I like having it consolidated down to one script.

There is a part of me that wonders whether this should all be one script that can take multiple SCEs for testing (so you never have to write out the AUC and don't have to reload the SCE library) but I doubt the complexity is worth the extra speed. I never really asked how long the AUC function takes though... how fast or slow is this?

allyhawkins:
> There is a part of me that wonders whether this should all be one script that can take multiple SCEs for testing (so you never have to write out the AUC and don't have to reload the SCE library) but I doubt the complexity is worth the extra speed. I never really asked how long the AUC function takes though... how fast or slow is this?

I would say it takes about 1 minute to run the script on a single sample. I think I want to leave it as reading in one at a time because I want to save the annotations and results separately to read into a notebook downstream.

PR #574, opened 2024-07-09
Add clarification about workflows in forks

### LINKED ISSUE #540: Expand docker GHA documentation: What happens after merge
We should add information to the Docker build GHA explaining what they can expect after the workflow is activated and merged into `main`. Specifically, the workflow will also attempt to run in contributors' forks once they sync, and it will fail (which is good!) due to credentialing. We should explain that this failure is expected and not a concern.

sjspielman:
Closes #540 

This PR adds a small section to the workflow docs explaining that, once activated workflows get merged, they will also run in your fork, and don't worry about when they fail! 

In addition to wanting feedback about clarity as usual, I have some questions for reviewers:

- Is this the right spot in the docs for this information? Should it be linked from anywhere?
- I am not at all sold on the header I used. Do we even need a header? If we do, should it be different?

### DIFF
--- a/docs/ensuring-repro/workflows/index.md
+++ b/docs/ensuring-repro/workflows/index.md
@@ -8,3 +8,19 @@ This section highlights two of these workflows:
 2. [A Docker-building workflow](./build-docker-gha.md) which builds and pushes module-specific Docker images to a public registry
 
 Generally, the Data Lab maintains and writes these GHA workflow files, but you are welcome to contribute as well if you are interested.
+
+## Workflows in your fork
+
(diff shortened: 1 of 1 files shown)


jashapiro:
I wonder if we should just fix this instead. I think we can add `if: github.repository_owner == 'AlexsLemonade'` to each job and that would prevent them from running in forks.

sjspielman:
> I wonder if we should just fix this instead. 

I think this is definitely the better choice if possible! I can open a PR with one workflow updated to see if it works as expected 🤞

allyhawkins:
> I wonder if we should just fix this instead. I think we can add `if: github.repository_owner == 'AlexsLemonade'` to each job and that would prevent them from running in forks.

I like this option, so I'm not going to review this right now.

PR #576, opened 2024-07-09
Don't run hello-R test GHA in forks

sjspielman:
Based on https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/574#issuecomment-2218200567, this PR updates the `hello-R` workflow so we can whether it runs in forks on merge/sync. I made a small change to ensure the GHA is triggered upon merge. If this works, I'll make these changes more broadly, including to the GHA templates.

Note that we could also try `github.repository = "AlexsLemonade/OpenScPCA-analysis` (or other options, scroll through https://docs.github.com/en/actions/learn-github-actions/contexts#github-context for more), but it should be equivalent in practice, so `github.repository_owner == "AlexsLemonade"` seems fine.

### DIFF
--- a/.github/workflows/run_hello-R.yml
+++ b/.github/workflows/run_hello-R.yml
@@ -24,7 +24,7 @@ on:
       - analyses/hello-R/**
 jobs:
   run-module:
-    if: github.repository_owner == "AlexsLemonade"
+    if: github.repository_owner == 'AlexsLemonade'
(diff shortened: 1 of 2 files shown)


jashapiro (approved):
👍🏼

sjspielman:
After merge, I made sure GHAs were turned on in my fork before syncing my fork back up with `main`. The action did not run!

Issue #467, opened 2024-05-28
Test benchmark data for the doublet GHA 


sjspielman:
> If you have a small dataset that you can download and run instead, that would be the preferred solution: you can set a ENV variable to select that option when running your wrapper script. 

I had been wondering about this too. I think this is a separate PR, to not automatically download the full benchmark zip, but include an option to use smaller `zip` file with dummy data. I'd also have to change the dataset array here, too. 

_Originally posted by @sjspielman in https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/462#issuecomment-2135796690_
            
----

This issue tracks (maybe!) adding tests benchmark data to improve the doublet detection workflow. That said, it's worth noting that, in the end, we may not care too too much about benchmarking results vs the doublet results on ScPCA data.

sjspielman:
This issue was originally created based on the idea of downloading a small dataset analogous to the data we're using for benchmarking, and run on that dataset instead. Since the test data format is in a quite specific format, we'll need to create such a dataset for download if we want to pursue this route. Since there isn't an existing downloadable dataset for this, we'd have to make one and store it in the repo, which might might having an actual download step a little overkill, even if technically more in line with the original script's behavior. Either way, I was thinking I'd just subset (and permute?) one of the benchmark datasets we already use for testing and add it to the `doublet-detection` module (I can't find any particular [licensing restrictions on the Zenodo page](https://zenodo.org/records/4562782) suggesting that would be a problem?). Then, maybe we just directly use this dataset in the GHA rather than downloading it, since the repo will have already been checked out. We can still `wget` it too. Any thoughts @jashapiro ?

PR #579, opened 2024-07-09
Detect doublets in ScPCA data

### LINKED ISSUE #544: Use SingleR to classify cells in SCPCL000824
> I wonder if we could combine a celldex reference with the tumor cells from SCPCL000822 to create a custom reference? This way we would have more than the two classes. This might avoid the overfitting problem and allow for SingleR to hopefully better distinguish between potential normal cell types and then tumor cells.

This is kind of what I was thinking, using the multiple reference functionality of SingleR. I kind of expect that you could keep the "Normal" label in the Tumor/normal reference from 822, as it should not fit better than any of the more specific references, and could serve as a bit of a negative control. Or maybe you could use the SingleR calls for 822 "normal" cells, but replace all of tumor calls with the "tumor" label. Maybe try both.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/532#issuecomment-2177237451_
        
This issue is to track looking at using different types of references to classify tumor cells in SCPCL000824. We should then compare to the annotations from marker gene based methods output in #532.

### LINKED ISSUE #554: Doublets: Write script to process ScPCA samples

#364 

Our next step in the `doublet-detection` module will be run doublet detection on ScPCA data. Specifically, we have discussed running `scDblFinder` on ScPCA data with the following approach:
- An R script to run scDblFinder on a single SCE. This should export a table of results for that SCE.
- A shell script to run that R script over all ScPCA libraries.

Two scripts: the R script to run `scDblFinder`, and the shell script to call it.

The overall run module shell script will also be modified so that you can indicate whether the module shoud be run in "benchmark" or "process" (?) mode, where the former will run benchmarking scripts/notebooks, and the latter will run doublet detection on ScPCA libraries.

There will be no changes in dependencies.

There will be no changes in compute - it can still be run on a laptop.

I expect to start this ~next sprint, which starts July 15th. So, we can expect a PR in probably 3ish weeks.~ towards the end of this sprint. We can expect a PR in 2ish weeks.

sjspielman:

Closes #554

This PR adds updates the `doublet-detection` module to run on ScPCA data, and I made some other module changes to enable this.

- Since we already had the script `scripts/01a_run-scdblfinder.R`, I decided to just use that script to run ScPCA data rather than write a new one that would be almost entirely identical. I figured there was no need to reinvent the wheel, but let me know if we think there should be any changes to script names or numbering now!
  - I updated existing script input arguments (also for the `scrublet` script, for consistency) to take an input file name, instead of a dataset name. This change allowed me to use the same `scripts/01a_run-scdblfinder.R` to process ScPCA data
  - I also updated comments in this script regarding the "multiple samples" argument, to make it clear that it doesn't apply to multiplexed libraries (https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/554#issuecomment-2203510839)
- I renamed the module bash script to `run_doublet-detection-benchmarking.sh`, with a few small associated changes to accommodate update script arguments. This script is now just used for the benchmarking side of the module. There were enough code changes here that GitHub is showing it as a new file, not a diff.
- I added a new script `run_doublet-detection-scpca.sh` to run the module on ScPCA data.
  - This script takes 1 argument, the project ID, and outputs a TSV for each processed SCE in the same directory structure as the OpenScPCA data, as: `{project}/{sample}/{library}_processed_scdblfinder.tsv`.
- I activated the module GHA and added _only_ `run_doublet-detection-scpca.sh` to it, to run on the `SCPCP000001` test data. 
  - Note that the benchmarking code will get added in #467. I commented out the conda environment code to save a little build time for now, since the ScPCA code doesn't need the conda environment.
- `README` files updated all around; let me know if I missed a spot!

I re-ran the benchmarking script to ensure that all works, and the output file names reflect the code changes.
I also ran `./run_doublet-detection-scpca.sh SCPCP000001`. 

Yes; the benchmarking script is currently not run in the GHA, but that should be addressed as part of #467.

Everything is here: `s3://researcher-654654257431-us-east-2/doublet-detection/`

- Benchmarking results, re-generated with slightly updated file names reflecting code changes I made here
- Doublet result TSVs for SCPCP000001

Well, we've got TSVs!

The `renv` and `conda` environments are included, and this can be run on a laptop.

N/A

(edit) There are some items I did not yet do. Is it time to do any of these? (CC @jashapiro)
- I did not add this module to `run_all-modules.yml`. 
- I still am setting up GHA environment with renv and conda (albeit the latter is commented out), but the Docker image has been pushed and is available for use.


analyses/doublet-detection/README.md
```diff
@@ -47,24 +59,39 @@ Specifically, these datasets are used (descriptions from [Xi and Li (2021)](http
   - `PBMCs` from a patient with systemic lupus erythematosus
   - Droplets were annotated with `demuxlet`
 
+### ScPCA input files
+
+The ScPCA portion of this module uses `processed` SCE files from most recent OpenScPCA data release, which can be obtained with:
```
allyhawkins, line 64:
```suggestion
The ScPCA portion of this module uses `processed` SCE files from the most recent OpenScPCA data release, which can be obtained with:
```

analyses/doublet-detection/run_doublet-detection-scpca.sh
```diff
@@ -0,0 +1,34 @@
...
+PROJECT_ID=$1
+
+# Define directories
+DATA_DIR="../../data/current"
+RESULTS_DIR="results/scpca-results/${PROJECT_ID}"
+mkdir -p ${RESULTS_DIR}
+
+# Detect doublets on each processed SCE file in each sample directory
+for SAMPLE_DIR in ${DATA_DIR}/${PROJECT_ID}/SCPCS*; do
+    SAMPLE_BASE=$(basename $SAMPLE_DIR)
```
allyhawkins, line 22:
```suggestion
    SAMPLE_ID=$(basename $SAMPLE_DIR)
```
This just returns the sample ID correct? Let's be explicit in how we name that then.

analyses/doublet-detection/run_doublet-detection-scpca.sh
```diff
@@ -0,0 +1,34 @@
...
+    echo "Processing ${SAMPLE_BASE}..."
```
allyhawkins, line 23:
```suggestion
    echo "Processing ${SAMPLE_ID}..."
```

analyses/doublet-detection/run_doublet-detection-scpca.sh
```diff
@@ -0,0 +1,34 @@
...
+
+    SAMPLE_RESULTS_DIR=${RESULTS_DIR}/${SAMPLE_BASE}
```
allyhawkins, line 25:
```suggestion
    SAMPLE_RESULTS_DIR=${RESULTS_DIR}/${SAMPLE_ID}
```

analyses/doublet-detection/run_doublet-detection-scpca.sh
```diff
@@ -0,0 +1,34 @@
...
+    mkdir -p ${SAMPLE_RESULTS_DIR}
+
+    for SCE_FILE in $(ls ${SAMPLE_DIR}/*processed.rds); do
+        Rscript scripts/01a_run-scdblfinder.R \
+            --data_dir ${SAMPLE_DIR} \
+            --input_sce_file $(basename ${SCE_FILE}) \
```
allyhawkins, line 31: Why do we need both? Can we just use `--input_sce_file`?

analyses/doublet-detection/run_doublet-detection-benchmark.sh
```diff
@@ -0,0 +1,59 @@
...
+
+    # formatted SCE and AnnData files will be saved here
+    DATASET_DIR=${DATA_DIR}/$dataset
+    mkdir -p $DATASET_DIR
+
+    # Read raw downloaded data and export SCE, AnnData files
+    ./scripts/00_format-benchmark-data.R --dataset ${dataset} --input_dir ${DATA_DIR}/raw --output_dir ${DATASET_DIR}
+
+    # Infer doublets with scDblFinder
+    ./scripts/01a_run-scdblfinder.R --input_sce_file ${dataset}.rds --data_dir ${DATASET_DIR} --results_dir ${RESULTS_DIR}
```
allyhawkins, line 45: I think I would define the SCE and AnnData files before running the scripts and then just have the `--input_sce_file` be an argument to the script and provide the full path. 

```suggestion
    SCE_FILE=${DATASET_DIR}/${dataset}.rds
    ./scripts/01a_run-scdblfinder.R --input_sce_file ${dataset}.rds --results_dir ${RESULTS_DIR}
```

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
```diff
@@ -71,14 +74,14 @@ run_scdblfinder <- function(sce,
 
 option_list <- list(
   make_option(
-    "--dataset_name",
+    "--data_dir",
```
allyhawkins, line 77: Based on my other comments, I would just remove this argument all together.

analyses/doublet-detection/scripts/01b_run-scrublet.py
```diff
@@ -66,8 +66,8 @@ def main() -> None:
     args.results_dir.mkdir(parents = True, exist_ok = True)
 
     # Run scrublet and export the results
-    input_anndata = args.data_dir / (args.dataset_name + "_anndata.h5ad")
-    result_tsv = args.results_dir / (args.dataset_name + "_scrublet.tsv")
+    input_anndata = args.data_dir / (args.input_anndata_file)
```
allyhawkins, line 69: Same thing, I would just define this path outside of the script and take the full path as the input here.

allyhawkins (commented):
This mostly looks good to me, my biggest comment that I made throughout is about removing the `--data_dir` argument and defining the paths to any input files in the shell script rather than in the scripts that run scDblFinder and Scrublet. That's going to make it a lot easier to port over to Nextflow.

.github/workflows/run_doublet-detection.yml
```diff
@@ -65,15 +62,21 @@ jobs:
...
-          conda-lock install --name test ${MODULE_PATH}/conda-lock.yml
+      - name: Download test data
+        run: ./download-data.py --test-data --format SCE --project SCPCP000001
 
       - name: Run doublet-detection module
         run: |
           cd $MODULE_PATH
-          bash run_doublet-detection.sh
+
+          bash run_doublet-detection-scpca.sh SCPCP000001
```
jashapiro, line 82: For best testing, we probably want to run on all projects. Probably something like:

```
projects=$(for p in data/current/SCPCP*; do basename $p; done)

cd $MODULE_PATH
for project in projects; do
  bash bash run_doublet-detection-scpca.sh $project
done
```

Of course, you will also want to download all of the test data as well.


In response to your other question, you may well want to use your Docker image to speed up test runs: see https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/run_cell-type-ewings.yml for an example where @allyhawkins just tested setting this up: you will need to install `aws` unless that is already part of your Docker image. (Also note that the `defaults:` section can probably be deleted; I meant to tell @allyhawkins that during review, but forgot amid the other things we were trying to get running)

sjspielman:
It looks like the GHA is failing because of a similar situation to https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/542.

But, according to https://gallery.ecr.aws/openscpca/doublet-detection, the image was last pushed 8 days ago, and [`blame` says](https://github.com/AlexsLemonade/OpenScPCA-analysis/blame/main/analyses/doublet-detection/Dockerfile) that `ENV OPENSCPCA_DOCKER TRUE` was added to the Dockerfile 2 weeks ago. The module's `.Rprofile` file does check this environment variable as well, so I'm not seeing why it's still trying to load `renv`. 

Anyone else see the issue here?

analyses/doublet-detection/run_doublet-detection-scpca.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Detect doublets on each processed SCE file in each sample directory
+for SAMPLE_DIR in ${DATA_DIR}/${PROJECT_ID}/SCPCS*; do
+    SAMPLE_ID=$(basename $SAMPLE_DIR)
+    echo "Processing ${SAMPLE_ID}..."
+
+    SAMPLE_RESULTS_DIR=${RESULTS_DIR}/${SAMPLE_ID}
+    mkdir -p ${SAMPLE_RESULTS_DIR}
+
+    for SCE_FILE in $(ls ${SAMPLE_DIR}/*processed.rds); do
```
jashapiro, line 28: It is better not to use `ls` in for loops; just the glob is sufficient
```suggestion
    for SCE_FILE in ${SAMPLE_DIR}/*_processed.rds; do
```

sjspielman:
`conda` did not like removing the `defaults:` section as suggested _might_ be ok in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/579#discussion_r1673971056, so I'm going to revert https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/579/commits/1b1a5b97b7e80ced9c39333c427ba8020508251a

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
```diff
@@ -7,18 +7,20 @@
 project_root <- rprojroot::find_root(rprojroot::is_renv_project)
 renv::load(project_root)
```
jashapiro, line 8: You need to remove `renv::load()` commands from your scripts to prevent loading when not wanted.

jashapiro (commented):
It loolks like you still have some `renv::load()` commands floating around. Those should be removed so the `.Rprofile` skipping is resepected.

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
sjspielman: omg 🤦🏻‍♀️

sjspielman:
The workflow failed on of a test SCE that only had 4 cells (droplets), and it doesn't surprise me that `scDblFinder` wouldn't like that. So, I should probably add a check for the number of cells (maybe 10? I'll poke around there) before running `scDblFinder`, but what do we think is the best way to handle an output TSV for such datasets? I feel like it would seem suspicious for a TSV to just be _missing_, but we can always include a README with the results that says "no TSVs for SCEs with fewer than X cells," and we'd have to include that README as part of the release results.

allyhawkins:
> The workflow failed on* a test SCE that only had 4 cells (droplets), and it doesn't surprise me that `scDblFinder` wouldn't like that. So, I should probably add a check for the number of cells (maybe 10? I'll poke around there) before running `scDblFinder`, but what do we think is the best way to handle an output TSV for such datasets? I feel like it would seem suspicious for a TSV to just be _missing_, but we can always include a README with the results that says "no TSVs for SCEs with fewer than X cells," and we'd have to include that README as part of the release results.

Can we do something like we do for cell typing in ScPCA where we still output the TSV but add in the columns and say "Not run" or something more informative like that?

jashapiro:
> Can we do something like we do for cell typing in ScPCA where we still output the TSV but add in the columns and say "Not run" or something more informative like that?

I like this, but I would probably just use `NA` in this case.

sjspielman:
> Can we do something like we do for cell typing in ScPCA where we still output the TSV but add in the columns and say "Not run" or something more informative like that?

Yeah, I also thought of that. I wondered if `NA` might be better for though for more flexibility down the line? I'm thinking, for example, if someone wrote code to loop over a project and add each tsv to each library's colData, the numeric columns would end up different across SCEs - they'd be character in anything with `"Not run"` and numeric otherwise, which might cause some weird script behavior for someone.

allyhawkins:
@sjspielman can you let me know when you've finished making any changes regarding the failing libraries and I'll review then?

sjspielman:
@allyhawkins this should be ready for another work. It's going to take a little while for the GHA to reach the problematic SCE, but the updated code works for me locally on small SCEs so hopefully that'll be good!

analyses/doublet-detection/README.md
```diff
@@ -22,17 +22,32 @@ conda-lock install --name openscpca-doublet-detection conda-lock.yml
...
+There are two portions of this module, which can be run as follows:
+
+- Benchmarking analysis: Compare several doublet detection methods on publicly available (non-ScPCA) datasets with annotated doublets.
+This analysis runs several doublet detection methods and assesses and compares their performance.
+  ```sh
+  ./run_doublet-detection-benchmarking.sh
+  ```
+
+- ScPCA analysis: Detect doublets in ScPCA data using `scDblFinder` for a given project id (`SCPCPXXXXXX`).
+This analysis exports a TSV file for each library in a given ScPCA project with `scDblFinder` inferences.
```
allyhawkins, line 34: Can you expand on what columns we should expect in the TSV?

analyses/doublet-detection/README.md
```diff
@@ -22,17 +22,32 @@ conda-lock install --name openscpca-doublet-detection conda-lock.yml
...
+SCEs with fewer than 10 droplets will not be run through `scDblFinder`, and their associated result TSVs
```
allyhawkins, line 35: I think this can be removed?

analyses/doublet-detection/README.md
allyhawkins: This might belong in the output files section, but either way you should describe the contents of the TSV file.

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
```diff
@@ -67,18 +67,53 @@ run_scdblfinder <- function(sce,
...
+      "ratio.k20",
+      "ratio.k25",
+      "lsizes",
+      "nfeatures",
+      "nAbove2",
+      "src",
+      "cxds_score",
+      "include.in.training",
+      "score",
+      "class"
```
allyhawkins, line 95: Ah there are a lot of columns here! Do we want to output everything? I think we could probably just output `score` and `class`? Those appear to be the most relevant (see https://github.com/plger/scDblFinder?tab=readme-ov-file#basic-usage) and I don't think we need these remaining stats...

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
```diff
@@ -105,35 +140,53 @@ option_list <- list(
...
+if (ncells < cell_threshold) {
+  warning(
+    glue::glue(
+      "The provided SingleCellExperiment object only has {ncells} droplets, which is fewer than {cell_threshold} so `scDblFinder` will not be run.
+       An output TSV file will still be produced, but it will be populated with `NA` values."
+    )
+  )
+  export_na_tsv(
+    colnames(sce),
+    output_tsv_file
```
allyhawkins, line 180: I think you could remove your function all together and replace with: 

```suggestion
  data.frame(
    barcodes = colnames(sce),
    score = NA, 
    class = NA
  ) |> 
  readr::write_tsv(output_tsv_file)
```

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
allyhawkins: We will need to update the function for running `scDblFinder` to only return those columns.

allyhawkins (commented):
My biggest remaining comment here is that we probably only want to return the `class` and `score` columns. I am pretty sure that's what we discussed when we talked about implementing this. There are a lot of extra columns that you are exporting that I think don't need to be there. That should allow you to simplify this more.

sjspielman:
> My biggest remaining comment here is that we probably only want to return the class and score columns. 

I realized we still need to keep the `cxds_score` column for the benchmarking part, so I added a `--benchmark` flag to the `scDblFinder` script to handle this.

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
```diff
@@ -91,10 +90,16 @@ option_list <- list(
     default = 4,
     help = "Number of cores to use during scDblFinder inference. Only used when there are multiple samples in the SCE."
   ),
+  make_option(
+    "--benchmark",
+    action = "store_true",
+    default = FALSE,
+    help = "Whether the script is being run as part of the doublet benchmarking analysis. Different columns will be output in this mode."
```
allyhawkins, line 97:
```suggestion
    help = "Whether the script is being run as part of the doublet benchmarking analysis. If this flag is invoked, the `cxds_score` column will be returned along with the `score` and `class` columns that are always returned. " 
```

allyhawkins (approved):
LGTM, just one documentation comment

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
```diff
@@ -91,10 +90,16 @@ option_list <- list(
     default = 4,
     help = "Number of cores to use during scDblFinder inference. Only used when there are multiple samples in the SCE."
   ),
+  make_option(
+    "--benchmark",
+    action = "store_true",
+    default = FALSE,
+    help = "Whether the script is being run as part of the doublet benchmarking analysis. If this flag is invoked, the `cxds_score` column will be returned along with the `score` and `class` columns that are always returned. "
```
sjspielman, line 97:
```suggestion
    help = "Whether the script is being run as part of the doublet benchmarking analysis. If this flag is invoked, the `cxds_score` column will be included in the output TSV along with the `barcodes`, `score`, and `class` columns that are always included." 
```

PR #580, opened 2024-07-09
Only run GHAs on forks

### LINKED ISSUE #577: Only enable GHAs in AlexsLemonade repos
This issue supplants #540 and tracks updating GHAs to run only in `AlexsLemonade` repos so that they don't run in contributors' forks, where they are expected to fail. 

All existing module running and docker building workflows, as well as the associated template workflow files, should be updated to include this `if` for all jobs:
https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/d22aaa9202454e2fa331e0cbb10a0f5e88101c47/.github/workflows/run_hello-R.yml#L25-L28

sjspielman:
Closes #577 

This PR updates GHAs to add a conditional statement so they won't run on forks.
Opening this as a draft, since there will be much testing and probably some syntax I need to fix, and then add a few more conditionals once I check some of said syntax.

I did not add the conditional to `create_periodic_release.yml` since this was only a manual trigger, but I can still do it if we want?  I also did not touch any of the jobs with an `always()` conditional, since it seemed to me that those jobs all had (unless I missed one!) `needs:` statements with jobs that I _did_ add the repo conditional to.

### DIFF
--- a/.github/workflows/file_periodic_release_issue.yml
+++ b/.github/workflows/file_periodic_release_issue.yml
@@ -6,7 +6,9 @@ on:
 
 jobs:
   post_issue:
+    if: github.repository_owner == 'AlexsLemonade'
     runs-on: ubuntu-latest
+
 
     steps:
       - name: Checkout repo
--- a/.github/workflows/run_cell-type-ewings.yml
+++ b/.github/workflows/run_cell-type-ewings.yml
@@ -24,7 +24,9 @@ on:
       - analyses/cell-type-ewings/**
 jobs:
   run-module:
+    if: github.repository_owner == 'AlexsLemonade'
     runs-on: ubuntu-latest
+
     defaults:
       run:
         shell: bash -el {0}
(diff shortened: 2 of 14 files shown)


.github/workflows/file_periodic_release_issue.yml
```diff
@@ -6,8 +6,10 @@ on:
 
 jobs:
   post_issue:
+    if: github.repository_owner == 'AlexsLemonade'
     runs-on: ubuntu-latest
 
+
```
jashapiro, line 12:
```suggestion
```

.github/workflows/run_cell-type-ewings.yml
```diff
@@ -24,7 +24,9 @@ on:
       - analyses/cell-type-ewings/**
 jobs:
   run-module:
+    if: github.repository_owner == 'AlexsLemonade'
     runs-on: ubuntu-latest
+
```
jashapiro, line 29:
```suggestion
```

jashapiro (approved):
LGTM, though I pulled out a couple extra blank lines that crept in.

PR #590, opened 2024-07-11
Remove AWS install from ewings action

jashapiro:
I realized while reviewing another PR that the Ewings module already does have `aws` installed, just in the conda environment. So here I am eliminating the install step and substituting activation.

I may do some additional testing, because the Docker image _should_ auto-activate, but github may be overriding that somehow.

### DIFF
--- a/.github/workflows/run_cell-type-ewings.yml
+++ b/.github/workflows/run_cell-type-ewings.yml
@@ -33,12 +33,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Download test data
-        run: |
-          conda activate openscpca-cell-type-ewings
-          ./download-data.py --test-data --format "sce,anndata" --samples SCPCS000490
+        run: ./download-data.py --test-data --format "sce,anndata" --samples SCPCS000490
 
       - name: Run analysis
         run: |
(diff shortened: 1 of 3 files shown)


jashapiro:
Converted back to a draft as #591 may affect the implementation here, at least in part.

.github/workflows/run_cell-type-ewings.yml
```diff
@@ -27,22 +27,15 @@ jobs:
...
-      - name: Install awscli
-        run: |
-          curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
-          unzip awscliv2.zip
-          ./aws/install
-
       - name: Download test data
-        run: ./download-data.py --test-data --format "sce,anndata" --samples SCPCS000490
+        run: |
+          conda activate openscpca-cell-type-ewings
```
allyhawkins, line 37: Do we still need this and the one in the next step if it's activated in the image?

jashapiro: Hopefully not. Will be testing once the docker image finishes building

allyhawkins (approved):
LGTM

PR #591, opened 2024-07-11
Add conda activation to ewings dockerfile

jashapiro:
When I filed https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/590, I didn't think to check that the conda environment was actually being activated as expected, but it wasn't! So here I am adding that step to the Dockerfile, which should simplify usage with GHA


allyhawkins (approved):
LGTM, thanks for catching this.

PR #597, opened 2024-07-11
Handle some stub links

### LINKED ISSUE #504: Replace stub links
As part of the #487 epic, we'll end up introducing some more `STUB_LINK`s to docs that are still in process. This issue tracks catching them all before merging the reproducibility docs branch into `main`.

sjspielman:
Closes #504 

We're getting to the end of the reproducibility docs epic, so I went ahead and checked out the remaining stub links. Turns out we had already gotten all the reproducibility ones along the way, but there were a few pre-existing ones that I figured could be addressed now. I filled in one stub link, and removed several stub link comments which are almost certainly not needed. Let me know if you think these decisions were reasonable!


allyhawkins (approved):
LGTM

Issue #567, opened 2024-07-05
Run workflow for annotating tumor cells and evaluate results 


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#292 
Step 4 of #563 

### Describe the goals of the changes to the analysis module.

Once we have created the workflow in #566 we will want to actually run that workflow and then take a look at the output reports that are generated. When looking at these reports we should identify any samples where: 

- the AUC distribution does not match up between the query and reference, meaning maybe we shouldn't be using this method to define tumor cells 
- The expected separation of gene set scores and marker gene expression between tumor and normal cells is not observed 

### What will your pull request contain?

This won't be a PR necessarily, but we should comment on any samples that do not fit our expectations and then come up with a plan on how to proceed with those samples before closing this. 

### Will you require additional software beyond what is already in the analysis module?

No 

### Will you require different computational resources beyond what the analysis module already uses?

No

allyhawkins:
I ran through `aucell-annotation.sh` for the whole project and am including a zip file of the rendered reports here: 

[aucell_annotation.zip](https://github.com/user-attachments/files/16195758/aucell_annotation.zip)

A few things that I noticed that I think are worth discussion:

- It looks like `AUCell` does a pretty good job with a library like SCPCL000824, the one we originally tested with, since the AUC values tend to be mostly within the range of the AUC values assigned to tumor cells in SCPCL000822. However, this is generally not the case. For almost all the other samples, the AUC values are either within the non-tumor cell range or centered like the examples below. Because of this, there are large numbers of cells getting called as normal cells. I generally have a hard time believing that most of these cells are normal.

![image](https://github.com/user-attachments/assets/0bb9a10a-f9bf-458d-a675-847d16000201)

![image](https://github.com/user-attachments/assets/492165d7-9841-438e-8422-40921176addd)


- I think we should be operating under the assumption that the majority of these cells are going to be tumor cells, so what we really want to do is find those cells that are outliers and likely to be normal cells. To me, any cells that don't have any marker gene expression should be called normal cells, which is happening with `AUCell`. However, there are a number of cells that do have a decent amount of marker gene expression that are getting called as normal, and I don't agree with that. In looking at some of the validation plots, most of them look like the one below. We see that a lot of cells that have marker gene expression are getting called as normal: 

![image](https://github.com/user-attachments/assets/70f8873b-1b4a-4d0e-9e15-18f1de2fc484)

[...]

PR #602, opened 2024-07-12
Add marker gene classification to AUCell report

allyhawkins:

Related to findings in #567 

Here I am making a small modification to the existing report rendered in the `AUCell` workflow to also classify cells based on marker gene expression. Any cell that expresses at least one marker gene is loosely classified as a "tumor" cell as we do expect the majority of the cells in these samples to be tumor cells. 

The purpose is to compare marker gene expression alone results with `AUCell` results, so I have updated all the plots to show both classifications. 

- Before creating any plots, I classify cells based on expression of any marker genes. If a cell expresses at least 1 gene then it's a tumor cell for the purpose of this exploration. We had previously started with this way back when we started this analysis. 
- All plots in the report have been updated to show both the AUCell classification and marker gene classification. 

Yes 

`s3://researcher-211125375652-us-east-2/cell-type-ewings/results/aucell_annotation` 
Note that I will update the results bucket with the new reports once this is approved. 

Here's a zip of the updated reports: 
[updated_aucell_reports.zip](https://github.com/user-attachments/files/16197059/updated_aucell_reports.zip)

An updated AUCell report and a new TSV file with the classifications from marker gene analysis. 

With this approach, we are classifying a lot more cells as tumors, which is what I would expect. This is from ScPCL000825: 

image

In this case, many of the "normal" cells from `AUCell` did have marker gene expression and there didn't seem to be clear separation between tumor and normal cells. Note that this plot is slightly circular for the marker gene classification since marker gene expression was directly used to classify, but I think it's helpful in showing the point that AUCell is under-calling tumor cells here. 

image

Also, with AUCell, I was seeing a lot of overlap in the distributions of gene set scores between tumor and normal cells. Adding this classification shifts the gene set scores so there is more of a separation between tumor and normal cells, which is what we would expect to see. 

image
image

I just want to note that I don't necessarily want to use this to call our final tumor cells, but I do think it's helpful in showing where samples may not be appropriate for use with `AUCell`. There are some samples that work really well and we see the expected separation in marker gene and gene set scores: SCPCL000822, SCPCL000824 and maybe SCPCL000823 and SCPCL001114. But the rest of them seem to have lower distributions of marker gene expression causing most cells to be classified as normal. 

I think next we need to use SingleR with multiple samples as the reference for those samples where `AUCell` did not perform well. 


analyses/cell-type-ewings/template_notebooks/auc-workflow/01-auc-results.Rmd
```diff
@@ -110,13 +111,24 @@ if(sum(markers_df$gene_expression) == 0){
 ```
 
 ```{r}
-# add marker gene classification based on expression of at least 1 marker gene  
+# get a list of tumor cells based on those that have number of marker genes expressed 
+# greater than the threshold 
+tumor_cells <- markers_df |> 
+  dplyr::mutate(gene_detected = dplyr:if_else(gene_expression > 0 , 1, 0)) |>
```
jashapiro, line 117: Simplify but keep it explicitly integer. 
```suggestion
  dplyr::mutate(gene_detected = as.integer(gene_expression > 0)) |> 
```

allyhawkins:
Just noting that I have re-run this and updated the results in my S3 bucket which can be found in `s3://researcher-211125375652-us-east-2/cell-type-ewings/results/aucell_annotation`

PR #612, opened 2024-07-15
Validate annotations for SCPCS000752

### LINKED ISSUE #606: Annotate tumor cells in SCPCS000752

#292 

In order to annotate tumor cells in the Ewing samples, we want to have 2-3 samples with reliable annotations and use them as a reference along with a normal `celldex` reference to annotate samples using `SingleR`. We already have annotations for SCPCS000490 and SCPCS000492 that we feel confident in, so we should manually annotate another sample to build that reference. 

This issue will track annotating SCPCS000752 so that we can eventually use it with `SingleR`. This sample was chosen based on the results from AUCell, as there does seem to be high marker gene expression and some clear separation between cells based on marker gene expression. It looks like AUCell is doing a decent job, but I believe we might be able to refine those assignments a little bit better by exploring other methods that we have used. 

A notebook that looks at assigning tumor cells using three different methods:

- Marker gene expression using the cutoff expression we determined from SCPCL000822. 
- AUCell using the cutoff we determined from SCPCL000822 (note that we already have these results) 
- UCell using the cutoff we determined from SCPCL000822 

We will then look at the consensus between all methods and validate our findings using the density plots and heatmaps we have been using throughout the module. 

No 

No 

_No response_

allyhawkins:

Closes #606

Here I am adding a notebook to explore and validate the tumor cell annotations for SCPCS000752/SCPCL001114. Ultimately, I would like to use this sample and the two previous samples we have been working with to build a `SingleR` reference to use for the remaining samples. 

This notebook includes: 

- Classifying cells based on using the marker gene cutoff obtained from `SCPCL000822`. This is the same approach I took in the notebook for `SCPCL000824` where I compared distributions of marker gene expression and then used the local minima as a cutoff for defining tumor cells. I think this approach is a bit more robust then just calling cells tumor if they have expression of at least one gene. 
- I compared the calls using the reference marker gene approach to the results from running `aucell-annotation.sh`. This includes looking at both the `AUCell` calls and marker gene calls (expression of 1 gene). 
- I also looked at using `UCell` here, but noticed that again we don't have a bimodal distribution and that the scores lie in between the scores for normal and tumor cells in `SCPCL000822`. So I didn't actually do any classification there since it didn't seem to make sense. 
- I did also take the consensus between the reference marker gene cutoff and `AUCell`. 
- The last section looks at marker gene expression and gene set scores across all methods. 

I will also note that I followed a very similar outline to what we have in `SCPCL000824_tumor-gene-expression.Rmd`. 

Yes

Rendered report for reference: 
[SCPCL001114_tumor-gene-expression-exploration.html.zip](https://github.com/user-attachments/files/16237886/SCPCL001114_tumor-gene-expression-exploration.html.zip)

An important piece of context here is that this is from a PDX and I do expect that most cells are going to be tumor cells. Although I do expect there to be some normal cells, I think they should be a minority. 

Generally, `AUCell` is the most conservative when calling tumor cells and using the single gene cutoff is the least conservative. Using the reference-based marker gene cutoff is somewhere in the middle. 

I think it's important to note that any cells called tumor by `AUCell` are also tumor in the marker gene methods. The biggest difference is that more cells are being called normal by `AUCell`. If you take the consensus of marker genes and `AUCell` we see that the "ambiguous" cells lie in the middle of tumor and normal when looking at marker gene expression and gene set scores. 

When we build our reference for `SingleR`, I would like to use either the reference based marker gene classification or the consensus of that with `AUCell`. I think with `AUCell` alone you are losing some tumor cells and with using a single marker gene as a cutoff you are over-simplifying and calling too many cells as tumor. 

I ran this locally on my laptop. 

While here I also fixed the spelling errors found in #609. Let me know if you disagree with any of my decisions on words to add to the dictionary. 


analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000752/SCPCL001114_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,522 @@
...
+caret::confusionMatrix(
+    table(
+      classification_df$ref_marker_gene_classification, 
+      classification_df$auc_classification)
+  ) 
+```
+
+It looks like using `AUCell` results in more cells being called normal than with the marker gene cutoff determined by `SCPCL000822`. 
+
+### Expression of at least one marker gene
```
jashapiro, line 266: What happens if you try at least 2 marker genes?

analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000752/SCPCL001114_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,522 @@
...
+classification_df <- classification_df |> 
+  dplyr::mutate(
+    marker_gene_classification = forcats::fct_relevel(marker_gene_classification, "Tumor")
+  )
+
+# compare using a confusion matrix
+caret::confusionMatrix(
+    table(
+      classification_df$marker_gene_classification, 
+      classification_df$ref_marker_gene_classification)
```
jashapiro, line 284: Can you swap the order of these for consistency with the previous plot? I was very confused trying to compare these two for a while.
```suggestion
      classification_df$ref_marker_gene_classification,
      classification_df$marker_gene_classification)
```

jashapiro (commented):
Overall, I think this looks good. At least for where it is... 

I think if we are going to use the sum of the marker genes, we should be more rigorous in fitting a model and finding the crossover point, but that is fine-tuning, so I am happy to keep pushing that off a bit. 

The one thing I do wonder is about whether 2 marker genes (or generally n marker genes) is a better cutoff than 1. It makes sense that 1 would be overly sensitive, but do we do better with 2 or lose too much?

Finally, just a note to be careful about consistent ordering!

allyhawkins:
> What happens if you try at least 2 marker genes?

I added a section for this, and it's slightly more conservative than using 1 gene, but it's still calling the majority as tumor cells. Which could be true, but I don't think we have enough evidence to prove this. I'm still hesitant to rely on a strict cutoff here. 

Here's the updated report: 
[SCPCL001114_tumor-gene-expression-exploration.html.zip](https://github.com/user-attachments/files/16256525/SCPCL001114_tumor-gene-expression-exploration.html.zip)

> Again, I like this generally, but I want something more rigorous if we are going to be looking for a break in a bimodal distribution. The curve minimum is about right, and works here, but it isn't going to be as good as actually fitting a mixture of distributions and then using that to find the cutoff. On the other hand, doing this "correctly" is not going to be easy...

I totally agree with this! I actually did start looking into doing this when I was working on the notebook, but there's a lot that would need to be worked out so I didn't want to go there just yet. Another thing to note is that this is one of few samples where the distributions tend to overlap. In many of the other samples the marker gene sum is much lower, so I don't feel great about using a cutoff in all the samples. This is why I think we need to move to something like SingleR where you are looking at more than just the pre-defined marker gene list. 

I also don't feel great about using the cutoff alone to call tumor cells and think for this sample and the previous sample we annotated (SCPCL000824) we should actually use the consensus between using the minimum from 822 as the cutoff for marker genes and AUCell. I think for now I would rather label cells as "ambiguous" rather than tumor if they are getting classified differently with those methods.

analyses/cell-type-ewings/exploratory_analysis/annotation_notebooks/SCPCS000752/SCPCL001114_tumor-gene-expression-exploration.Rmd
```diff
@@ -0,0 +1,576 @@
...
+classification_df |> 
+  tidyr::pivot_longer(
+    cols = ends_with("classification"),
+    names_to = "method",
+    values_to = "classification"
+  ) |> 
+  dplyr::mutate(
+    method = forcats::fct_recode(method, !!!method_levels)) |> 
+  ggplot(aes(x = UMAP1, y = UMAP2, color = classification)) +
+  geom_point(size = 0.1, alpha = 0.5) +
```
jashapiro, line 436: Two things here: I am not a fan of piping into ggplot and I woudl rather see you create a new df first.
But the more important thing is that I think it is worth throwing a `forcats::fct_reorder("Consensus", after = Inf)` to move the consensus to the end here (and everywhere else you do this transformation). It feels strange to have it be somewhere in the middle

jashapiro (approved):
LGTM, with one suggestion for factor ordering.

PR #615, opened 2024-07-16
Provide output TSV to doublets

### LINKED ISSUE #613: Doublets script should take output file name, not path
From https://github.com/AlexsLemonade/OpenScPCA-nf/pull/72#discussion_r1679904578:

> Okay, you convinced me. You should change this script to take an output file name, rather than a directory. Then your wrapper shell scripts can change this more easily and the whole thing becomes more flexible.

---

This issue tracks updating the `scDblFinder` script as described.

sjspielman:
Closes #613

This PR updates the doublet scripts (both R and python; I did the latter for module consistency) to accept an input arg `output_tsv_file` instead of `results_dir`. This will make porting the analysis to `OpenScPCA-nf` more straightforward. Note that I'm requesting @jashapiro for review in case there's anything else that might jump out at you that might benefit from a change before heading over to `nf`. The rest looks ok to me, but let me know what you think (and if you have thoughts about my bash `sed` line too!).

### DIFF
--- a/analyses/doublet-detection/run_doublet-detection-scpca.sh
+++ b/analyses/doublet-detection/run_doublet-detection-scpca.sh
@@ -4,7 +4,7 @@
 # Usage: ./run_doublet-detection-scpca.sh {scpca project id}
 
 set -euo pipefail
-
+CORES=4
 
 # Ensure script is being run from its directory
 MODULE_DIR=$(dirname "${BASH_SOURCE[0]}")
@@ -29,6 +29,7 @@ for SAMPLE_DIR in ${DATA_DIR}/${PROJECT_ID}/SCPCS*; do
         TSV_FILE=$(basename "${SCE_FILE%.rds}_scdblfinder.tsv")
         Rscript scripts/01a_run-scdblfinder.R \
             --input_sce_file ${SCE_FILE} \
-            --output_tsv_file ${SAMPLE_RESULTS_DIR}/${TSV_FILE}
+            --output_tsv_file ${SAMPLE_RESULTS_DIR}/${TSV_FILE} \
+            --cores $CORES
     done
 done
--- a/analyses/doublet-detection/run_doublet-detection-scpca.sh
+++ b/analyses/doublet-detection/run_doublet-detection-scpca.sh
@@ -26,7 +26,7 @@ for SAMPLE_DIR in ${DATA_DIR}/${PROJECT_ID}/SCPCS*; do
     mkdir -p ${SAMPLE_RESULTS_DIR}
 
     for SCE_FILE in ${SAMPLE_DIR}/*_processed.rds; do
-        TSV_FILE=`sed 's/.rds/_scdblfinder.tsv/' <<<"$(basename $SCE_FILE)"`
+        TSV_FILE=$(basename "${SCE_FILE%.rds}_scdblfinder.tsv")
         Rscript scripts/01a_run-scdblfinder.R \
             --input_sce_file ${SCE_FILE} \
             --output_tsv_file ${SAMPLE_RESULTS_DIR}/${TSV_FILE}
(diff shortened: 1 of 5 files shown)


analyses/doublet-detection/scripts/01a_run-scdblfinder.R
```diff
@@ -80,9 +80,9 @@ option_list <- list(
     help = "Path to input SCE file in RDS format."
   ),
   make_option(
-    "--results_dir",
+    "--output_tsv_file",
```
jashapiro, line 83: I tend to favor slightly shorter option names. So either `output_file` or `output_tsv`... But this is very minor.
```suggestion
    "--output_file",
```

analyses/doublet-detection/scripts/01a_run-scdblfinder.R
```diff
@@ -114,20 +114,11 @@ opts <- parse_args(OptionParser(option_list = option_list))
 if (!file.exists(opts$input_sce_file)) {
     glue::glue("Could not find input SCE file, expected at: `{opts$input_sce_file}`.")
 }
-if (is.null(opts$results_dir)) {
-  stop("Must provide an output path for results with --results_dir.")
+if (!stringr::str_ends(opts$output_tsv_file, ".tsv")) {
```
jashapiro, line 117: You didn't set a default for this arg, so `NULL` is still possible which will break this. I might give the arg a default value of "scdblfinder_results.tsv" and leave this as is. (I have come to dislike NULL defaults for reasons like this... I get tired of having to include `is.null()` checks everywhere)

analyses/doublet-detection/scripts/01b_run-scrublet.py
```diff
@@ -58,23 +58,27 @@ def main() -> None:
...
         )
         sys.exit(1)
-    result_tsv = args.results_dir / (input_anndata_file.name.replace(".h5ad", "_scrublet.tsv"))
+
+    if not args.output_tsv_file.name.endswith(".tsv"):
+        print(
+            "The output TSV file must end in `.tsv`.",
+            file=sys.stderr
+        )
+        sys.exit(1)
```
jashapiro, line 74: Minor, but if there are a series of argument checks, I tend to prefer to set a `arg_error=False` at the start of the checks and then use `arg_error=True` (or set to zero and increment, depending on mood) rather than immediately exiting. Then at the end of the checks: 
```
if arg_error:
  sys.exit(1)
```

I'd rather just print all the errors at once if I can.

analyses/doublet-detection/scripts/01b_run-scrublet.py
```diff
@@ -58,23 +58,27 @@ def main() -> None:
...
     # Define and check files, directories
-    args.results_dir.mkdir(parents = True, exist_ok = True)
-
-    input_anndata_file = Path(args.input_anndata_file)
-    if not input_anndata_file.exists():
+    if not args.input_anndata_file.exists():
         print(
             "The input AnnData file could not be found at:",
             args.input_anndata_file,
             file=sys.stderr
```
jashapiro, line 65: fstrings ftw
```suggestion
            f"The input AnnData file could not be found at: {args.input_anndata_file}.",
            file=sys.stderr
```

analyses/doublet-detection/run_doublet-detection-scpca.sh
```diff
@@ -26,8 +26,9 @@ for SAMPLE_DIR in ${DATA_DIR}/${PROJECT_ID}/SCPCS*; do
     mkdir -p ${SAMPLE_RESULTS_DIR}
 
     for SCE_FILE in ${SAMPLE_DIR}/*_processed.rds; do
+        TSV_FILE=`sed 's/.rds/_scdblfinder.tsv/' <<<"$(basename $SCE_FILE)"`
```
jashapiro, line 29: I know this is less obvious, but extension removal is a builtin bash substitution thing, and in this case it is a bit safer than what you did.

Also, better not to mix backticks and `$()`: you can nest `$()` and it is the preferred method for process substitution.
```suggestion
        TSV_FILE=$(basename "${SCE_FILE%.rds}_scdblfinder.tsv")
```

jashapiro (approved):
Looks pretty good, but I had a few little cleanup suggestions. 

The only one that is critical is the fix for a missing output file argument in the R script.

PR #622, opened 2024-07-17
Create merged tumor cell only reference for SingleR

### LINKED ISSUE #607: Create a reference for Ewing samples to use with `SingleR`

#292 

We want to be able to run `SingleR` on the remaining Ewing samples that have been difficult to annotate based on marker gene expression alone. Before we can do this we need to create a trained model from our reference samples. We will use the tumor cell assignments from SCPCL000822, SCPCL000824, and SCPCL001114 along with a normal cell reference from `celldex`. 

The first thing we will need to do is add in the tumor cell annotations to the SCE object. 
Then we will finalize the annotations stored in the individual reference SCE objects. 

- Tumor cells will be labeled as "Tumor"
- Any non tumor cells that are labeled with the same cell type in using both SingleR and CellAssign will get that assignment (using the original scpca-nf annotations). 
- All other cells will be re-annotated using `BlueprintEncodeData` from `celldex`

Then the `SingleR` model will be created using the three samples and Blueprint and saved to an RDS object to use later to annotate all other samples. 

No

No 

_No response_

allyhawkins:
Closes #607 

The goal here is to create a merged `SingleCellExperiment` object that contains only the tumor cells we are confident in for all libraries in SCPCP000015. This object will then be used as input to the next script which will run `SingleR` on a single library. 

- For every library in the project, tumor cells are obtained by taking the results from running `AUCell`. `AUCell` is very conservative in calling tumor cells, so we feel confident that any cells that are called tumor are in fact a tumor cell. 
- The processed SCE objects for each library are read in and subset to contain only the tumor cells. 
- I then added a new column to the `colData` of each object, `ref_tumor_label`. This is going to be the label we use when running `SingleR` and it contains the "tumor" label and the library ID. This means that every group of tumor cells from each library will be treated as a different cell type when labeling cells. We could also harmonize the labels so that they are all just tumor cells and be agnostic to the original library, but I'm not sure we really want to do that at this point. 
- The output of the script is a merged SCE object with only tumor cells and the new `ref_tumor_label` column as well as a `library_id` column. That way when we run through `SingleR` we can remove any cells from the library that is the same as the library being annotated. 

Yes 

`s3://researcher-211125375652-us-east-2/cell-type-ewings/scratch`

A single RDS file with a merged object containing all tumor cells identified by `AUCell` for SCPCP000015. 

I ran this script locally without any issue.

I did have to deal with a case where we don't have any tumor cells in the object. Ultimately we just don't want to include that library in the final merged object since there are no tumor cells to include, but if there's a more efficient way to handle that in the code I wrote let me know! 

Also the way I have this setup right now is to use the [reference-specific labels](https://bioconductor.org/books/release/SingleRBook/using-multiple-references.html#using-reference-specific-labels) method for using multiple references. However, we could easily [harmonize the labels and follow that approach](https://bioconductor.org/books/release/SingleRBook/using-multiple-references.html#using-harmonized-labels). I thought this was a good place to start, but let me know if you think otherwise. 


analyses/cell-type-ewings/scripts/singler-workflow/01-generate-tumor-ref.R
```diff
@@ -0,0 +1,118 @@
...
+# list of tumor cell sce 
+sce_list <- purrr::map2(sce_files, annotation_files, 
+                        \(sce_file, annotation_file){
+                          subset_tumor_cells(sce_file, annotation_file)}) 
+
+# remove any objects that dont' have tumor cells
+sce_present <- sce_list |> 
+  purrr::map_lgl(is.null)
+
+remaining_sce_list <- sce_list[!sce_present]
```
jashapiro, line 106: Making the variable match contents, adding `which` for safety (though I don't think you can get NA with `is.null`), and fixing a little typo
```suggestion
# remove any objects that don't have tumor cells
sce_absent <- sce_list |> 
  purrr::map_lgl(is.null)

remaining_sce_list <- sce_list[which(!sce_absent)]
```

analyses/cell-type-ewings/scripts/singler-workflow/01-generate-tumor-ref.R
```diff
@@ -0,0 +1,118 @@
...
+
+# make sure annotation files are in the same order as sce files 
+annotation_files <- annotation_files[order(annotation_files, sce_library_ids)]
+
+# Create merged tumor cell ref -------------------------------------------------
+
+# list of tumor cell sce 
+sce_list <- purrr::map2(sce_files, annotation_files, 
+                        \(sce_file, annotation_file){
+                          subset_tumor_cells(sce_file, annotation_file)})
```
jashapiro, line 100: Since the function takes those two arguments in that order, I don't think you need to define an inline function
```suggestion
sce_list <- purrr::map2(sce_files, annotation_files, subset_tumor_cells) 
```

jashapiro (approved):
This looks fine to me, with only a couple little suggestions.

PR #624, opened 2024-07-18
Add test benchmark data for doublets module

sjspielman:
Towards #467 

This PR adds a very small data file which can be used for benchmark testing to the `doublet-detection` module. My idea was to add a file to the repo which can be `wget`'ed (...`wgotten`?) in `run_doublet-detection-benchmark.sh` instead of the full zenodo zip, and this downloaded file can be used in the GHA to actually test the benchmark code. After this PR, I'll update the shell script and GHA to use this file in a testing mode.

My concern here is I might be playing this a little fast and loose by assuming that the raw data has already been downloaded, and further I'm not intending to run this script as part of the module pipeline at all. What are your feelings about these choices? Should I make this script a more reproducible part of the module, or is it ok that it's just a quick script that I ran on the fly to create this file and is effectively just being recorded here for posterity? 

Also, would we like to see a `benchmark-test-data/README.md` file or is the documentation I added to the module's top-level `README.md` sufficient?


analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
sjspielman, line 1: fyi this diff is just a typo fix for #609, so recommended to turn off whitespace!

analyses/doublet-detection/benchmark-test-data/generate-benchmark-test-data.R
```diff
@@ -0,0 +1,22 @@
...
+
+# This script generates a test data file for use when testing the benchmarking script `../run_doublet-detection-benchmark.sh`.
+# The test data file is a subsetted version of the `hm-6k` dataset used for benchmarking, as described in `../README.md`.
+# This script assumes that ../scratch/benchmark-datasets/raw contains the .rds files from <https://doi.org/10.5281/zenodo.4562782>'s `real_datasets.zip` file.
+
+
+library(Matrix)
+
+input_file <- file.path("..", "scratch", "benchmark-datasets", "raw", "hm-6k.rds")
+output_file <- file.path("hm-6k_subset.rds")
```
jashapiro, line 11: These paths should be set relative to something concrete and not assume that this script will be run from its current location.

analyses/doublet-detection/exploratory-notebooks/03_compare-benchmark-results.Rmd
jashapiro: Why are your editors/processors not agreeing on whitespace?

jashapiro (commented):
Just a paths comment

sjspielman:
> Why are your editors/processors not agreeing on whitespace?

I think I'm going to blame an RStudio update changing my settings here such that RStudio didn't capture these, whether that's a valid thing to blame or not! Def need to fix those settings.......

analyses/doublet-detection/benchmark-test-data/generate-benchmark-test-data.R
sjspielman: I added `projroot::find_root(rprojroot::is_renv_project)` here - https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/624/commits/59ec24970992aa440e6b1505c31b300c59cb032b

jashapiro (approved):
lgtm

PR #625, opened 2024-07-18
Remove doublets dockerfile entrypoint

sjspielman:
As discussed, I've removed the entrypoint to see the consequences when running with this image in nextflow. Let's see how it goes!

PR #626, opened 2024-07-18
Fix concurrency groups

jashapiro:
I noticed in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/625 that actions were being cancelled because both docker and "run" actions were being triggered. We want to be able to run both! So here I made the concurrency groups more specific, including in the templates.


jashapiro:
I also added a little fix for some complaints that docker was making about syntax. It wants `=` for env setting now.

allyhawkins (approved):
Assuming this passes, LGTM

sjspielman (approved):
everything looks in place to me!

PR #627, opened 2024-07-18
Script to run SingleR on all samples from SCPCP000015

### LINKED ISSUE #621: Script to run SingleR using combined tumor cell reference

#292 

As outlined in https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/607#issuecomment-2233988203, we plan on creating a reference with confident tumor cells across all samples in SCPCP000015. We then will use that combined reference and `BlueprintEncodeData` from `celldex` to annotate the remaining cells in each library. 

This PR will have a single script that takes as input a single SCE file and the combined tumor cell reference. 

- The tumor cells corresponding to the input library will be removed from the combined tumor cell reference. 
- All tumor cells will be removed from the input SCE file. 
- SingleR will be run with the remaining tumor cell reference and `BlueprintEncodeData` on all remaining cells (non high confident tumor) 
- The final annotations for all cells in the input SCE file will be saved to a TSV. 

No 

No 

After #607  is completed

allyhawkins:

Closes #621 

Here I'm adding a script that takes as input a single SCE object and the reference tumor SCE created in #622 and runs `SingleR` on all non-tumor cells in the object. The goal is to be able to obtain annotations for all cells in the object where confident tumor cell calls are kept and all other cells are annotated using both the combined tumor reference and `BlueprintEncodeData` from `celldex`. 

The following steps are taken in the script: 

- The input SCE object is filtered to only retain any non-tumor cells. Tumor cells are those that are present in the reference tumor object for that library (these were called based on the `AUCell` output). 
- The reference tumor SCE is filtered to remove any tumor cells from the specified library. This way we avoid `SingleR` assigning cells to that tumor cell population due to them being from the same sample and having similar gene expression. 
- SingleR is run on all remaining cells with the filtered tumor ref and `BlueprintEncodeData` using the `label.ont` column.
- The results are extracted into a data frame with both the ontology ID and the human readable value for any normal cells. The annotations from tumor cells are returned as `Tumor-library_id`. 
- The results from `SingleR` are combined with the rest of the cells that were already labeled as tumor and output to a TSV file.  

Yes 

I have only tested this with one sample and was able to run it locally using `--threads 4`. 

Just noting that my plan next is to create a pretty small template report to visualize the `SingleR` results and then put that into a workflow to run all of the samples. 

### Author checklists

_Check all those that apply._


allyhawkins:
I made a slight update to this to also add using `HumanPrimarCellAtlas` to the normal references that we use. In my initial analysis with using just `BlueprintEncodeData`, I was seeing a lot of cells get assigned as tumor that were previously classified as endothelial by `SingleR` and `CellAssign` and as normal when we manually classified cells for `SCPCL000822`. I remembered that Anand said in one of his posts that he planned to use the HPCA reference and has had good luck with it and there are a lot more cell types included. When I updated to use both HPCA and Blueprint, I saw cell types that lined up more with what I expected. 

These are pretty rough plots that I've just been playing around with, sorry for the small text! I just wanted to give you some rationale as to why I added in HPCA in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/627/commits/994801c80a36ee1bdfc992eb1241281cb0967ecb

Here's using both HPCA and celldex: 
image

I looked at common markers of endothelial cells and immune cells (taken from the supplemental in [Visser et al.](https://doi.org/10.1158/2767-9764.CRC-23-0027)) and I see that they are lining up with what I would expect: 
PECAM1 - Endothelial cells
image

PTPRC - Immune 
image


In contrast this is using Blueprint alone: 
image

PECAM1 - Endothelial, it looks like the "tumor_new" cells have some expression and are probably misclassified here 
image

jashapiro:
> * The input SCE object is filtered to only retain any non-tumor cells. Tumor cells are those that are present in the reference tumor object for that library (these were called based on the `AUCell` output).

Just starting to review this, but I wonder if we want this filtering here? My thought is that we may want to see if the tumor cells from AUCell are being annotated as tumor cells by SingleR as well as a partial sanity check.

analyses/cell-type-ewings/scripts/singler-workflow/02-run-singler.R
```diff
@@ -0,0 +1,129 @@
...
+# Prep references --------------------------------------------------------------
+
+# grab celldex references
+blueprint_ref <- celldex::BlueprintEncodeData(ensembl = TRUE)
+hpca_ref <- celldex::HumanPrimaryCellAtlasData(ensembl = TRUE)
+
+# remove tumor cells from SCE object for annotation 
+library_id <- metadata(sce)$library_id
+sce_tumor_cells <- tumor_ref$barcodes[which(tumor_ref$library_id == library_id)]
+filtered_sce <- sce[ , !(colnames(sce) %in% sce_tumor_cells)]
```
jashapiro, line 77: I already said this in a separate comment, but I think we may not want to do this filtering; we'd like to be able to compare how these cells are annotated by SingleR as a positive control.

analyses/cell-type-ewings/renv.lock
jashapiro, line 1: Are you using any of these additions? Seems like this may have gotten added for some reason unrelated to the scripts below.

analyses/cell-type-ewings/scripts/singler-workflow/02-run-singler.R
```diff
@@ -0,0 +1,129 @@
...
+  # replace ont labels with full human readable labels
+  dplyr::left_join(cl_df, by = c("ontology")) |>
+  # if its not found in blueprint its tumor so keep the original annotation 
+  dplyr::mutate(annotation = dplyr::if_else(is.na(annotation), ontology, annotation))
+
+# combine back with tumor cells 
+tumor_annotations_df <- data.frame(barcodes = sce_tumor_cells) |> 
+  dplyr::mutate(
+  annotation = glue::glue("tumor-{library_id}")
+)
```
jashapiro, line 119: Just cleaning up some odd spacing.
```suggestion
  dplyr::mutate(annotation = glue::glue("tumor-{library_id}"))
```

analyses/cell-type-ewings/scripts/singler-workflow/02-run-singler.R
```diff
@@ -0,0 +1,129 @@
...
+  annotation = cl_ont$name, # CL ID
+  ontology = names(cl_ont$name) # human readable name
+)
+
+# grab results from data frame and add human readable values for normal ont labels 
+results_df <- data.frame(
+  barcodes = rownames(singler_results),
+  ontology = singler_results$pruned.labels
+) |>
+  # replace ont labels with full human readable labels
```
jashapiro, line 110: replacing or adding? I would assume we want both?

analyses/cell-type-ewings/scripts/singler-workflow/02-run-singler.R
```diff
@@ -0,0 +1,129 @@
...
+             tumor_ref = filtered_tumor_ref),
+  labels = list(blueprint_ref$label.ont, hpca_ref$label.ont, filtered_tumor_ref$ref_tumor_label),
+  BPPARAM = bp_param
+)
+
+# get ontology labels
+cl_ont <- ontoProc::getOnto("cellOnto")
+cl_df <- data.frame(
+  annotation = cl_ont$name, # CL ID
+  ontology = names(cl_ont$name) # human readable name
```
jashapiro, line 102: I'm confused here: you are later joining on `ontology`, which I would expect to the the CL ID?

analyses/cell-type-ewings/scripts/singler-workflow/02-run-singler.R
```diff
@@ -0,0 +1,129 @@
...
+
+# grab results from data frame and add human readable values for normal ont labels 
+results_df <- data.frame(
+  barcodes = rownames(singler_results),
+  ontology = singler_results$pruned.labels
+) |>
+  # replace ont labels with full human readable labels
+  dplyr::left_join(cl_df, by = c("ontology")) |>
+  # if its not found in blueprint its tumor so keep the original annotation 
+  dplyr::mutate(annotation = dplyr::if_else(is.na(annotation), ontology, annotation))
```
jashapiro, line 113: An alternative to do the same thing, I think:
```suggestion
  dplyr::mutate(annotation = dplyr::coalesce(annotation, ontology))
```

analyses/cell-type-ewings/scripts/singler-workflow/02-run-singler.R
```diff
@@ -0,0 +1,129 @@
...
+
+# combine back with tumor cells 
+tumor_annotations_df <- data.frame(barcodes = sce_tumor_cells) |> 
+  dplyr::mutate(
+  annotation = glue::glue("tumor-{library_id}")
+)
+
+all_results_df <- dplyr::bind_rows(list(results_df, tumor_annotations_df)) |>
```
jashapiro, line 121: I think we would might to keep these original annotations as a separate column, if we call _all_ the cells.

jashapiro (commented):
The overall scheme here looks good. I wasn't sure about the expanded `renv` but the more important comment is the one I already made and expanded a bit here. I think we want to call _all_ the cells, but still leave the current library out of the reference set. 

As I write that probably want to be even more extreme in filtering: remove all samples from the same participant, if we can do that easily. If not, we should be able do the same sample, rather than library, which I think would be a minimum. This situation may not apply to this project at all if we only have one library per sample and one sample per participcant, but I think it should be noted that if it did, we would want to filter at the higher level.

analyses/cell-type-ewings/renv.lock
allyhawkins: These are all `scpcaTools` dependencies that should have been added in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/622 and I just forgot, sorry about the confusion!

allyhawkins:
> The overall scheme here looks good. I wasn't sure about the expanded renv but the more important comment is the one I already made and expanded a bit here. I think we want to call all the cells, but still leave the current library out of the reference set.

This makes sense to me! I don't anticipate those to change calls, but you are right that's a good positive control. I think this also means making some code updates to #634. Mainly I'll probably add a confusion matrix between the AUCell calls and the SingleR calls, but will keep a lot of the remaining plots. 

> As I write that probably want to be even more extreme in filtering: remove all samples from the same participant, if we can do that easily. If not, we should be able do the same sample, rather than library, which I think would be a minimum. This situation may not apply to this project at all if we only have one library per sample and one sample per participcant, but I think it should be noted that if it did, we would want to filter at the higher level.

I think this is a great idea! Especially for this project since we do have 2-4 samples per participant. I will make this adjustment which will also include adding a column for `participant_id` to the merged reference so we can easily subset it.

allyhawkins:
@jashapiro I updated this to perform annotation of all cells in the original object rather than removing the already annotated cells. I also removed any cells from libraries from the same participant ID in the reference instead of just library ID. This should be ready for another look.

analyses/cell-type-ewings/scripts/singler-workflow/02-run-singler.R
```diff
@@ -98,32 +101,27 @@ singler_results <- SingleR::SingleR(
 # get ontology labels
 cl_ont <- ontoProc::getOnto("cellOnto")
 cl_df <- data.frame(
-  annotation = cl_ont$name, # CL ID
-  ontology = names(cl_ont$name) # human readable name
+  annotation = cl_ont$name, # human readable name 
+  ontology = names(cl_ont$name)# CL ID
```
jashapiro, line 105: okay, this makes more sense!

jashapiro (approved):
LGTM! Thanks for the updates!

PR #633, opened 2024-07-19
Add more test benchmark datasets

sjspielman:
Towards #467 

Yesterday, I filed #624 to add a subsetted benchmarking dataset to use when testing doublets. When I then went to implement the updates to use this pushed dataset, I realized it was not going to be enough to test all the code, since one of the benchmarking notebooks necessarily compares datasets to each other. 

In this PR, I'm revisiting the benchmarking test data to add a subsetted version of _all four datasets_ in a `.zip` file, which much more closely mirrors the actual benchmarking data. This should allow all benchmarking code to run properly! Each subsetted dataset contains 50 droplets (45 singlet and 5 doublet) and 100 genes.


allyhawkins (approved):
LGTM!

sjspielman:
Since these changes have no bearing on the code currently run in the module test GHA, I'm going to merge before that check completed.

PR #634, opened 2024-07-19
Add template for SingleR evaluation

### LINKED ISSUE #632: Create a template report for evaluating SingleR results

#292 

After all libraries are run through `SingleR` with the combined tumor reference and normal references from `celldex`, we will need some way to evaluate the assignments. This should be a template notebook that produces informative plots for letting us know how `SingleR` did. 

This will be a template notebook with relevant plots for evaluating the SingleR assignments. Some ideas that I have been playing around with include: 
- UMAPs showing the "original tumor cells", "newly assigned tumor cells", and the top 5 normal cell types 
- Heatmap/ density plot looking at marker gene expression and gene set scores in tumor and normal cell types 
- Heatmap/ density plot that looks at expression of some key normal tumor markers (e.g., PECAM in endothelial cells). Although we don't know that the same groups of normal cells are being found in all samples, we can start with using the markers identified in Visser et al for the general normal cell type groups (immune, endothelial, mesenchymal). We already specify these markers in [`references/visser-all-marker-genes.tsv`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-ewings/references/visser-all-marker-genes.tsv). 

No 

No

_No response_

allyhawkins:

Closes #632 

This PR adds a template report to run after running the SingleR script being added in #627. The goal is to be able to have some thing that we can use to help evaluate the output from annotating each sample using SingleR. 

There are a few things that I think we want to be able to get from looking at this notebook: 

- Are all the tumor cells being called distinctly different from normal cells, as in do they have higher expression of marker genes and gene set scores then the rest of the cells? 
- How do the tumor cells being annotated using our combined reference look compared to the tumor cells annotated by `AUCell`. Mainly, I am concerned that we could have some over fitting here since it is likely that the cells are more similar to cells from other Ewing's libraries over normal cells. On the other hand I do expect most cells to be tumor cells. 
- Do the normal cell types look expected? What I've added here is not a rigorous exploration of normal cell types at all, but I do think it's a good starting point for seeing how well SingleR is doing across all samples. I think we can come back and tweak the normal cell evaluation as a next step. To do this I am using the normal cell markers that were used in Visser et al., and we already have saved in `references/visser-all-marker-genes.tsv`. 

In the notebook I included a few plots that I think help us look at this: 

- Heatmaps looking at the expression of all marker gene sets across all cell types. Here I would expect tumor cells to be distinct from normal cells. I also expect endothelial cells and immune cell types to show higher expression of both of those marker gene sets. 
- Density plots looking at the sum of the expression of all marker gene sets across cell types. I would expect tumor cells to show increased expression of the tumor markers, endothelial cells increased expression of those markers, and so forth. 

I included these plots with a few different annotations: 

- Just looking at all tumor vs. all normal 
- Looking at all tumor (both the original tumor cells and any newly annotated tumor cells by SingleR) vs all individual normal cell types (As I'm putting this together, I think I could get rid of this section of the notebook since it's not that different from the third bullet point). 
- Separately looking at original tumor, new tumor, and all normal cell types 

This might be overkill, but I think it will be helpful for when we look at other samples. 

Yes 

This is a template notebook, so I haven't run it on all the samples yet, but I did test this out with `SCPCL000822`, which also helped me evaluate how well this approach of using the combined tumor reference would work. I think maybe there are some cells that appear more "normal" that are in fact getting called "tumor", BUT there is so much heterogeneity in these tumors that I am not that surprised by this. A lot of literature has shown subsets of tumor cells to have low EWS-FLI1 expression and, therefore, low expression of key EWS-FLI1 targets, which are most of our marker genes. 

Here's a rendered copy of the report for SCPCL000822: 
[01-singler-results.html.zip](https://github.com/user-attachments/files/16316673/01-singler-results.html.zip)

I ran this locally. 


analyses/cell-type-ewings/scripts/utils/tumor-validation-helpers.R
```diff
@@ -199,3 +237,86 @@ plot_cnv_heatmap <- function(
...
+  
+  # build matrix for heatmap cells x gene set sum or mean 
+  heatmap_mtx <- classification_df |>
+    dplyr::select(barcodes, gene_exp_columns) |> 
+    tibble::column_to_rownames("barcodes") |> 
+    as.matrix()
+  colnames(heatmap_mtx) <- stringr::str_remove(colnames(heatmap_mtx), "_sum|mean-")
+  
+  # plot heatmap of marker genes 
+  plot_gene_heatmap(t(heatmap_mtx),
```
jashapiro, line 271: I'd probably move this `t()` up to the processing pipeline above. (and then change `colnames()` to `rownames()`, I guess)

analyses/cell-type-ewings/template_notebooks/singler-workflow/01-singler-results.Rmd
```diff
@@ -0,0 +1,335 @@
...
+
+# generate classification df to use for plots 
+classification_df <- sce |> 
+  scuttle::makePerCellDF(use.dimred = "UMAP") |> 
+  # replace UMAP.1 with UMAP1
+  dplyr::rename_with(
+    \(x) stringr::str_replace(x, "^UMAP\\.", "UMAP")
+  ) |>
+  # get rid of excess columns
+  dplyr::select(barcodes, UMAP1, UMAP2) |>
```
jashapiro, line 91: I don't know why I haven't suggested this before, but maybe could to simplify this to:
```suggestion
  # get rid of excess columns and rename UMAPs
  dplyr::select(barcodes, UMAP1 = UMAP.1, UMAP2 = UMAP.2) |> 
```

analyses/cell-type-ewings/template_notebooks/singler-workflow/01-singler-results.Rmd
```diff
@@ -0,0 +1,335 @@
...
+# get the sum of expression of all genes for each cell type
+gene_exp_df <- cell_types |> 
+  purrr::map(\(type){
+    calculate_sum_markers(marker_genes_df, type)
+  }) |> 
+  purrr::reduce(dplyr::inner_join, by = "barcodes")
+
+# join sum expression columns with classification df 
+classification_df <- classification_df |> 
+  dplyr::left_join(gene_exp_df)
```
jashapiro, line 144: Avoid warning
```suggestion
  dplyr::left_join(gene_exp_df, by = "barcodes")
```

jashapiro (commented):
I started to leave a couple comments, but I realized with my comments on #621 you may want to make some changes here? In particular if you want to compare the AUCells tumor calls to singleR calls made on _all_ cells in a sample, there would be some changes here. I'm not sure the best way to represent those; maybe a separate set of UMAP plots with  `tumor_AUCell_only`, `tumor_SingleR_only`, `tumor_both` or maybe just an upset plot? Either way, I think we want to settle #621 first.

allyhawkins:
> I started to leave a couple comments, but I realized with my comments on https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/621 you may want to make some changes here? In particular if you want to compare the AUCells tumor calls to singleR calls made on all cells in a sample, there would be some changes here. I'm not sure the best way to represent those; maybe a separate set of UMAP plots with tumor_AUCell_only, tumor_SingleR_only, tumor_both or maybe just an upset plot? Either way, I think we want to settle https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/621 first.

I've updated this based on the changes in #621. I added a section that has a UMAP showing tumor cells annotated by both, AUCell only, and SingleR only. I also added in a confusion matrix between tumor/normal in AUcell and SingleR. I will note that generally I don't anticipate these things to line up, but for SCPCL000822 I do so I think it's a nice positive control to see a high kappa value in this sample. 

Also since we no longer have `tumor_new` and `tumor_original` as all cells were reannotated with `SingleR`, I made some adjustments to the sections. I show the marker gene and gene set scores for all tumor vs. all normal, all tumor vs. individual normal cell types, and then within tumor cell groups (both SingleR and AUCell, AUCell only, and SingleR only). 

Here's the updated report: 
[01-singler-results.html.zip](https://github.com/user-attachments/files/16339923/01-singler-results.html.zip)

analyses/cell-type-ewings/template_notebooks/singler-workflow/01-singler-results.Rmd
```diff
@@ -0,0 +1,314 @@
...
+
+**Gene set scores**
+
+```{r}
+full_celltype_heatmap(classification_df, geneset_columns, "singler_tumor_normal")
+```
+
+```{r}
+geneset_columns |> 
+  purrr::map(\(column){
```
jashapiro, line 230: Just noting that the code here is writing out the list names, which I think can be avoided by capturing the `map` output:
```suggestion
plots <- geneset_columns |> 
  purrr::map(\(column){
```

I don't _think_ that would prevent the plots from printing, but if it did you could use `patchwork:: wrap_plots()` on the resulting list.

jashapiro (approved):
This looks good overall. I just had the one aesthetic concern about using `map()` with the plotting function as that seems to give some ugly-ish code chunks in between the plots. Not sure if you might need `patchwork` to get rid of that, but it is also going to be fine to just leave it as is.

PR #636, opened 2024-07-22
Final update to doublet test data

sjspielman:
As briefly discussed last Friday, I still need to do one more update to the test benchmark data towards #467 
This PR updates the test data to select high enough information genes & cells to be able to run algorithms. I select genes based on highest means, and cells based on highest total counts. This works way better than random selection - I have thoroughly tested and these subsetted datasets will work to run the benchmarking code through (next PR!)


sjspielman:
noting that I canceled the doublet GHA workflow, since code here isn't run through the workflow.

allyhawkins (commented):
This is fine, but I think we still want to have a way to actually handle the error you mentioned about not having enough PCs in the actual script for scrublet. I think part of having test data is to make sure we handle those errors and avoid things breaking in the future.

sjspielman:
> This is fine, but I think we still want to have a way to actually handle the error you mentioned about not having enough PCs in the actual script for scrublet.

I have this waiting to be filed in the next PR. See here in the next branch, which is ready to be filed: https://github.com/sjspielman/OpenScPCA-analysis/blob/904abd26b582da4a5ff1e1b6fd187027f5935f01/analyses/doublet-detection/scripts/01b_run-scrublet.py#L25-L30

There are a couple other code changes in that branch too to facilitate test data. Do you me to cherry pick that `scrublet` change and/or other small code changes here?

allyhawkins (approved):
I just wanted to make sure that those errors were being handled. If they are coming in the next PR that is fine with me and I will go ahead and approve this.

PR #637, opened 2024-07-22
Add doublet benchmark analysis to GHA

### LINKED ISSUE #467: Test benchmark data for the doublet GHA

sjspielman:
Closes #467 

This PR adds a step to the doublet detection GHA to run the benchmark script through now that we have benchmarking test data to use. There are several additional changes I made to enable this:

- I added a `test` argument to the benchmark bash script so the test data can be invoked instead of the real data. While I was here, I also updated the `CORES` variable in the bash scripts to be an argument in the same style.
- I updated the scrublet script to use only 2 pcs if there is a really low number of neighbors. We only ever expect this to happen on test data.
- I updated notebooks to make sure they still run even if 0 doublets are detected (which happens in some of the test data!), and I specifically updated the comparison notebook to be a bit more robust with specifying input dataset names - these are now a param rather than being pulled out of result file names.
- Changes are documented in the README.

Let me know where I can clarify any of the changes I made here!


.github/workflows/run_doublet-detection.yml
```diff
@@ -43,11 +43,17 @@ jobs:
           conda activate openscpca-doublet-detection
           ./download-data.py --test-data --format SCE
 
-      - name: Run doublet-detection module
+      - name: Run doublet-detection module on benchmark test data
+        run: |
+          cd ${MODULE_PATH}
+          conda activate openscpca-doublet-detection
```
allyhawkins, line 49:
```suggestion
```
I don't think you need this since the environment was activated in the download test data step.

.github/workflows/run_doublet-detection.yml
```diff
@@ -43,11 +43,17 @@ jobs:
...
+          test=1 cores=1 ./run_doublet-detection-benchmark.sh
```
allyhawkins, line 50: I don't think you actually need to specify using 1 core? I use 4 in the ewing's action without an issue.

analyses/doublet-detection/scripts/01b_run-scrublet.py
```diff
@@ -21,8 +21,16 @@ def run_scrublet(adata: anndata.AnnData, random_seed: int) -> pandas.DataFrame:
     """
 
     scrub = scrublet.Scrublet(adata.X, random_state = random_seed)
+
+    # Handle test datasets, which are very small
+    # The real datasets used in this module will always have > 10
+    if (scrub.n_neighbors < 10):
+        n_pc = 2
```
allyhawkins, line 28: How did you figure out the cutoff to use here was 10? Is there a way to specifically grab the number of pcs present in the `scrub` object and use that rather than `n_neighbors`?

allyhawkins (commented):
This looks good, just a few question that I had. The biggest thing is how you determined the `n_neighbors` cut off of 10?

.github/workflows/run_doublet-detection.yml
sjspielman: > I use 4 in the ewing's action without an issue.

Ah cool, I'll remove all this

analyses/doublet-detection/scripts/01b_run-scrublet.py
sjspielman: > Is there a way to specifically grab the number of pcs present in the scrub object and use that rather than n_neighbors?

I know, this 10 is less than ideal; I set it because the test datasets that were failing were below 10. I agree this should have been done in a systematic way, but I didn't see an immediate way how and didn't think it was worth the effort to dig in further. The `__init__` function for the `Scrublet` class doesn't actually perform the PCA yet - it doesn't do much of anything besides set the object up for use: https://github.com/swolock/scrublet/blob/67f8ecbad14e8e1aa9c89b43dac6638cebe38640/src/scrublet/scrublet.py#L100-L126

So, that information isn't available in the `scrub` object. I'm definitely open to other ideas though - for example, we might prefer a `try`/`except` somehow on the associated `ValueError`?

.github/workflows/run_doublet-detection.yml
sjspielman: let's see!

analyses/doublet-detection/README.md
```diff
@@ -28,15 +28,24 @@ There are two portions of this module, which can be run as follows:
...
 
 - ScPCA analysis: Detect doublets in ScPCA data using `scDblFinder` for a given project id (`SCPCPXXXXXX`).
   - The conda environment is not needed for this analysis, since it is only R-based
   ```sh
   ./run_doublet-detection-scpca.sh {scpca project id}
 
-  # for example, detect doublets on all libraries in ScPCA project SCPCP000001:
-  ./run_doublet-detection-scpca.sh SCPCP000001
+  # the script uses 4 cores by default. Use the cores argument to customize:
+  cores=2 run_doublet-detection-scpca.sh {scpca project id}
```
sjspielman, line 45:
```suggestion
  cores=2 ./run_doublet-detection-scpca.sh {scpca project id}
```

analyses/doublet-detection/scripts/01b_run-scrublet.py
allyhawkins: > So, that information isn't available in the scrub object. I'm definitely open to other ideas though - for example, we might prefer a try/except somehow on the associated ValueError?

Hmm I think this could be a good approach! I think I would set it up so you run the whole thing and then if it fails, you just output a dataframe with barcodes and `NA` for the scrublet results columns rather than guessing at the best `n_pcs` to use.

sjspielman: better parallel with `scDblFinder` too!

sjspielman: Ah, this will mean I need to further update notebooks to handle cases where scrublet results are `NA`. Hopefully that's not too involved but if it is I'll let you know...

sjspielman: Not too involved! ✅

sjspielman:
This PR is on pause until https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/639 goes in and the Docker image is updated.

analyses/doublet-detection/scripts/01b_run-scrublet.py
```diff
@@ -21,11 +21,20 @@ def run_scrublet(adata: anndata.AnnData, random_seed: int) -> pandas.DataFrame:
...
+        doublet_scores, predicted_doublets = scrub.scrub_doublets()
+        # convert True/False to string values
+        predicted_doublets_str = [("doublet" if x else "singlet") for x in predicted_doublets]
+    except:
+        print(
+            "Could not run scrublet on the provided dataset."
+            "The output TSV file will contain only NA values."
+        )
+        doublet_scores = ["NA"]*len(adata.obs_names)
+        predicted_doublets_str = ["NA"]*len(adata.obs_names)
```
sjspielman, line 37: Review question: is there another preferred way here to get lists of `"NA"` strings?

analyses/doublet-detection/README.md
```diff
@@ -28,15 +28,24 @@ There are two portions of this module, which can be run as follows:
...
   ```sh
   ./run_doublet-detection-scpca.sh {scpca project id}
 
-  # for example, detect doublets on all libraries in ScPCA project SCPCP000001:
-  ./run_doublet-detection-scpca.sh SCPCP000001
+  # the script uses 4 cores by default. Use the cores argument to customize:
+  cores=2 ./run_doublet-detection-scpca.sh {scpca project id}
+
+  # for example, detect doublets on all libraries in ScPCA project SCPCP000001 with 6 cores:
+  cores=6 ./run_doublet-detection-scpca.sh SCPCP000001
```
allyhawkins, line 48: I think you could simplify this to one line and just show the last example, but up to you.

analyses/doublet-detection/scripts/01b_run-scrublet.py
jashapiro: Do you need it to be a string? Or would it be better to use `pandas.NA` here? Either way, you should be able to use a single value for these, as the number of rows  of the data frame will be determined by the barcodes and pandas will recycle single values as needed. 

(You may want to use `numpy.nan` for the scores to have them be floats, but since you are just outputting a file and probably don't care too much about the types as you would if you were going to process the data frame itself.)

allyhawkins: I think this is fine, but if you wanted to load in `numpy` then you could just do: 
`numpy.full(shape=len(adata.obs_names), fill_value ="NA")`. 

I think either one is okay.

analyses/doublet-detection/template-notebooks/02_explore-benchmark-results.Rmd
```diff
@@ -260,31 +272,37 @@ scdbl_df <- read_tsv(scdbl_tsv) |>
...
+    sce_df,
     by = "barcodes"
   )
 ```
 
+```{r}
+# Check if scores are present, to determine if certain chunks should be run
+has_scdbl <- !all(is.na(doublet_df$scdbl_score))
+has_cxds <- !all(is.na(doublet_df$cxds_score))
+has_scrublet <- !all(is.na(doublet_df$scrublet_score))
```
allyhawkins, line 304: I think this is a good approach.

analyses/doublet-detection/scripts/01b_run-scrublet.py
allyhawkins: Just saw Josh's suggestion and I like that better. If you just need a single value then use that.

sjspielman: > but since you are just outputting a file and probably don't care too much about the types as you would if you were going to process the data frame itself.)

this is definitely the thought I had that me think `"NA"` as a plain string was good enough! but yeah a single value seems better

sjspielman:
Noting I replaced the `scrublet` `NA` code with a single `"NA"`. When I tried `pandas.NA`, the output TSV had fully blank values for those columns.

jashapiro:
One more request before you submit this: can you run the python code through `black` or `ruff` for formatting?

PR #638, opened 2024-07-22
Use a subset of samples for testing doublets in PRs

### LINKED ISSUE #631: Consider options to reduce test times
When making modifications to the `detect-doublets` module, I became acutely aware of how long that module is taking to test. 

This is mostly my fault, as I requested that it be run on _all_ test data during an early review. I stand by that decision, as my memory is that it did discover some edge cases that would have otherwise not been caught. However, at this stage with that module largely "mature" module, that level of testing seems like a bit of overkill.

I think the best option is to select a "representative" subset of samples that span projects, with special focus on cases that have caused errors in the past or are likely to cause errors in the future (looking at you, multiplexed data!). 

We might retain an option to run on all data for periodic testing runs. In other words, PRs could run a shorter check, with monthly checks using the full data (simulated) sets. This seems to me to be a good compromise between speed of development and catching edge cases. 

What do others think? 

Tagging folks: @allyhawkins @jaclyn-taroni @sjspielman

sjspielman:
Closes #631 

This PR updates the doublet GHA to use a set of representative samples for testing in PRs only. I define an environment variable `SAMPLES` based on whether or not the triggering event is a PR. 

I grabbed these sample IDs as either:
- First sample ID in each project, or
- For SCPCP000011, use the small sample that the doublet module previously failed on (`SCPCS000435`)
- For SCPCP000009, I made sure to grab multiplexed samples/libraries. I am using this set of multiplexed samples, which have 2 associated libraries -  `SCPCS000133,SCPCS000134,SCPCS000135,SCPCS000136`.

----

Edit: GHA passed very quickly 🎉 and this is now ready for review! Let me know if this is about the idea we had in mind here!


.github/workflows/run_doublet-detection.yml
```diff
@@ -10,6 +10,8 @@ name: Run doublet-detection analysis module
 env:
   MODULE_PATH: analyses/doublet-detection
   AWS_DEFAULT_REGION: us-east-2
+  # use all samples for testing unless this is triggered by a PR, in which case only use a subset of samples for faster testing
+  SAMPLES: ${{ github.event_name != 'pull_request' && 'all' || 'SCPCS000435,SCPCS000133,SCPCS000134,SCPCS000135,SCPCS000136,SCPCS000001,SCPCS000024,SCPCS000050,SCPCS000101,SCPCS000250,SCPCS000168,SCPCS000216,SCPCS000124,SCPCS000298,SCPCS000481,SCPCS000484,SCPCS000514,SCPCS000490,SCPCS000502,SCPCS000246,SCPCS000616,SCPCS000632,SCPCS000758' }}
```
jashapiro, line 14: I would move this down to the step where it is used. You can make an `env` block within that.

jashapiro: For `SCPCS000133,SCPCS000134,SCPCS000135,SCPCS000136`  you should be able to use just one sample; the script will get all libraries that contain the sample, even if multiplexed.

jashapiro (commented):
This looks good, with just a couple modifications. 

I had a thought about turning the list of samples into an array and joining it, but I think it would get messy enough to probably not be worth the editability. I think it would loook something like this though:

```
SAMPLES: |
  ${{ github.event_name != 'pull_request' &&
      'all' ||
      join(
        fromJSON('[
          "SCPCS000001",
          "SCPCS000024",
          "SCPCS000050"
        ]'),
        ','
     )
  }}
```

Not pretty...
 

What I would do though is sort the samples, and probably include in the comments how the samples were selected.

.github/workflows/run_doublet-detection.yml
```diff
@@ -39,14 +40,21 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Download test data in SCE format
+        env:  |
```
jashapiro, line 43: No pipe needed here
```suggestion
        env: 
```

sjspielman: beat you https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/638/commits/43954b245421795881671f37b065f029d6242d1c

.github/workflows/run_doublet-detection.yml
```diff
@@ -39,14 +40,21 @@ jobs:
...
+        env:
+          # Use all samples for testing unless this is triggered by a PR, in which case only use a subset of samples for faster testing
+          # A single sample (or combination of samples for multiplexed data) was chosen for each ScPCA project as:
+          #  Unless otherwise stated, the first sample ID in the given project
+          #  For the multiplexed project SCPCP000009, set of samples `SCPCS000133,SCPCS000134,SCPCS000135,SCPCS000136` (this group of samples has 2 associated libraries). Only `SCPCS000133` is specifed in the SAMPLES string here.
+          #   For the project SCPCP000011, the sample SCPCS000435 was specifically selected as an edge case sample with <10 cells
+          SAMPLES: ${{ github.event_name != 'pull_request' && 'all' || 'SCPCS000001,SCPCS000024,SCPCS000050,SCPCS000101,SCPCS000124,SCPCS000133,SCPCS000168,SCPCS000216,SCPCS000246,SCPCS000250,SCPCS000298,SCPCS000435,SCPCS000481,SCPCS000484,SCPCS000490,SCPCS000502,SCPCS000514,SCPCS000616,SCPCS000632,SCPCS000758' }}
+
         run: |
           conda activate openscpca-doublet-detection
```
jashapiro, line 52: Is this `conda` line still required? I ask for two reasons: we did update the container (which still works in general here, yay!) but I noticed that it wasn't needed in the next step, which makes me think we maybe don't need it any more? Or maybe because there is an explicit call to `bash` within that chunk, it works out and does call `.bashrc` as it launches the script. 

Or, as I think about it, it is probably the fact that you are just using R there...

jashapiro (approved):
LGTM. I had one thought but it is probably wrong as I realized while writing.

.github/workflows/run_doublet-detection.yml
sjspielman: Yep, it's b/c R+python in benchmark step, but just R in scpca step. Your comment here suggests that a comment in the GHA couldn't hurt!

sjspielman: (but I suppose the benchmark step isn't in this PR...)

PR #639, opened 2024-07-22
Add pandoc to Dockerfile

sjspielman:
Related to #637 
Now that the benchmarking notebooks are being run in the doublets GHA, it has become [apparent that the Docker image needs `pandoc`](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/10047296868/job/27768675342#step:5:516). I've added it here to the Dockerfile, so once this image makes it to the registry hopefully #637 will pass.

### DIFF
--- a/analyses/doublet-detection/Dockerfile
+++ b/analyses/doublet-detection/Dockerfile
@@ -52,6 +52,13 @@ RUN Rscript -e 'renv::restore()' \
   && rm -rf /tmp/downloaded_packages \
   && rm -rf /tmp/Rtmp*
 
+# Install pandoc dependency
+RUN apt-get -y update &&  \
+  DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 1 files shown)


analyses/doublet-detection/Dockerfile
```diff
@@ -52,6 +52,13 @@ RUN Rscript -e 'renv::restore()' \
   && rm -rf /tmp/downloaded_packages \
   && rm -rf /tmp/Rtmp*
 
+# Install pandoc dependency
+RUN apt-get -y update &&  \
+  DEBIAN_FRONTEND=noninteractive \
+  apt-get install --no-install-recommends -y \
+  pandoc \
+  && rm -rf /var/lib/apt/lists/*
+
```
jashapiro, line 61: Move this near the top; system dependencies should be earlier layers.

allyhawkins (approved):
LGTM

PR #641, opened 2024-07-22
Update Dockerfile and docs to use CMD instead of ENTRYPOINT

jashapiro:
Here I am applying the changes in from #626 in `analyses/doublet-detection/Dockerfile` to all of the Dockerfiles, replacing `ENTRYPOINT` with `CMD`. This seems to solve the problems we were having with proper launching and activation of conda environments in `openscpca-nf`.

I also updated the docs to reflect the same suggestions.

### DIFF
--- a/analyses/cell-type-ewings/Dockerfile
+++ b/analyses/cell-type-ewings/Dockerfile
@@ -65,5 +65,5 @@ RUN Rscript -e 'renv::restore()' && \
   rm -rf /tmp/downloaded_packages && \
   rm -rf /tmp/Rtmp*
 
-# Set entrypoint to bash to activate the environment for any commands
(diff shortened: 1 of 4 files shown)


sjspielman (approved):
One more spot, but I don't need to see again:
https://github.com/jashapiro/OpenScPCA-analysis/blob/9972a4f7c3d35cc820e55c0093402f47351fc12a/docs/ensuring-repro/docker/docker-images.md?plain=1#L106

~Just kidding, this was about conda, not Docker. I just got excited by the grep result~

**Just kidding about kidding, I was right the first time.** And it's time for me to call it a day!

sjspielman (approved):
meant to approve!
edit - I did approve? either way, it's approved.

PR #643, opened 2024-07-23
Fix path in periodic release issue

jashapiro:
I noticed that our periodic release action was running but not filing issues. Turns out the path was wrong, but that doesn't result in failure, just a silent exit.

I updated the action with the correct path, which seemed sufficient for now. Should I also add a step that fails if the file does not exist?

I also updated the docker action to push all images for periodic releases as well as semantic versions.


jaclyn-taroni (commented):
Since this is _scheduled_, I do think we want it to explicitly fail and tell someone.

jashapiro:
@jaclyn-taroni I added failure if the file does not exist. This will tell _somebody_ but we don't currently have a mechanism other than filing an issue for notification in this repo (no secrets). At least not one that I can think of.  And filing an issue because we can't file an issue might be a bit circular!

.github/workflows/docker_all-modules.yml
```diff
@@ -18,6 +18,7 @@ on:
     # build and push all on tagged release
     tags:
       - "v*.*.*"
+      - "20[0-9][0-9]-[0-1][0-9]-[0-3][0-9]"
```
jaclyn-taroni, line 21: I don't know if I agree with this. Can you tell me more about why you think we need this?

jashapiro: I mostly just thought it might be nice to have periodically tagged docker images in addition to semantic, but I don't feel too strongly about it and think it is fine to remove this.

I also just realized that I'm not sure the current docker action would push these if they are not semantic, so I'm going to revert this change.

jaclyn-taroni (approved):
👍🏻

PR #651, opened 2024-07-23
Add modules to run_all

jashapiro:
Here I am adding `detect-doublets` and `cell-type-ewings` to the `run_all` workflow.

I also added some instructions to be sure this workflow is up to date before completing a versioned release. 

I wasn't totally sure we needed to add `cell-type-ewings` yet, but I figured it wouldn't hurt.


sjspielman (approved):
👍

PR #653, opened 2024-07-23
handle explicit "all" options for download scripts

jashapiro:
In https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/10066556460 there was an error because the download script does not actually handle "all" as an option for samples or projects, expecting instead a blank string if the option was set. Because blank strings are hard to deal with at the command line, I added a step to explicitly parse "all" as an option in both the data and results scripts.

### DIFF
--- a/download-data.py
+++ b/download-data.py
@@ -436,6 +436,11 @@ def main() -> None:
     if validation_error:
         sys.exit(1)
 
+    # convert "all" to empty string for later processing
+    if args.projects.casefold() == "all":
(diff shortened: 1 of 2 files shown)


jashapiro:
If somebody does get the chance to review this small PR tonight, that would be great!

jaclyn-taroni (approved):
👍🏻  Tested that `download-data.py` worked as expected locally with `--dryrun`

PR #657, opened 2024-07-24
Fix docker action repository naming

### LINKED ISSUE #655: Analysis module failing in CI
At least one of the Docker images that is part of the [Build all Docker images](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/workflows/docker_all-modules.yml) workflow has failed.

https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/workflows/docker_all-modules.yml

Then assign an OpenScPCA admin to this issue and add to this issue a list of the modules with failing docker builds.
Alternatively, file a new issue for each failing Docker image.

jashapiro:
Here I am removing a couple of legacy triggers for tagged versions from docker actions, and fixing the error that came up when trying to make a new repo for the `hello-R` module. 

closes #655

The error there was that `hello-R` has a capital letter in it, which is not allowed in a docker tag. So I added a step to convert that before checking for the presence of the tag in AWS ECR. 

The docker metadata action itself already [sanitizes image tag](https://github.com/docker/metadata-action/tree/v5/?tab=readme-ov-file#image-name-and-tag-sanitization), so I believe that I only have to change the AWS portion of the action.

I considered pulling out the sanitized version from the metadata action, but that actually seems like it will be a fair bit more complex. So there is still a risk of module names that include unallowed characters, but that seems like a low risk, and one that may be best addressed by changing the module names.

### DIFF
--- a/.github/workflows/build-push-docker-module.yml
+++ b/.github/workflows/build-push-docker-module.yml
@@ -41,8 +41,9 @@ jobs:
       - name: Create ECR repository if needed
         id: create-ecr
         run: |
-          aws ecr-public describe-repositories --repository-names ${{ inputs.module }} \
-          || aws ecr-public create-repository --repository-name ${{ inputs.module }}
+          module=echo "${{ inputs.module }}" | tr '[:upper:]' '[:lower:]'
+          aws ecr-public describe-repositories --repository-names $module
+          || aws ecr-public create-repository --repository-name $module
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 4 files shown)


allyhawkins (approved):
LGTM

sjspielman:
Does this transformation mean these docs are now deprecated? https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/61a24b067682d540a0b9443ef7ca0133c8e6ef8d/docs/ensuring-repro/docker/using-images.md?plain=1#L20-L22

sjspielman (approved):
👍

PR #658, opened 2024-07-24
610 Add script to apply SEACells to AnnData objects

### LINKED ISSUE #610: Add script to apply SEACells to AnnData objects

https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/362

As the first step in creating and analyzing metacells, we will create a script to run the SEACells algorithm on an individual object. 

The goal at this is to have a reusable component that can be applied to each object and can be used for benchmarking and future analyis.

The plan is a single python script that takes an AnnData object as input and outputs another object containing a set of metacells. This script should also allow for the selection of input parameters to the SEACells algorithm. The script will also include a standalone function that can be used when imported into another script or notebook. 

This may also include creating a plot showing the metacells overlaid on the individual cells in a UMAP, but that may come as a second PR with the addition of other benchmarking statistics.

Optionally, the script may output timing benchmarks to a log file.

None expected

Unknown at the moment: some initial computational benchmarks at this stage may be applied to get a sense for this in future PRs

Expected by July 17

jashapiro:
Closes #610

This PR adds a python script to apply the SEACells algorithm to AnnData objects.

The main steps and default parameters are derived from https://github.com/dpeerlab/SEACells/blob/main/notebooks/SEACell_computation.ipynb and associated SEACells source code. 

One thing to note was that there are few subtleties to the expected formats and associated metadata for things like PCA and UMAP, so it was necessary to recalculate those values, so I wrote a small function to do that. I initally thought I might be able to get away with just renaming the matrices, but it seemed like some of the metadata stored when `scanpy` calculates the dimension reduction matrices was required downstream, so recalculation seemed the way to go. In my testing, the PCA results were nearly identical, but UMAP were subtly different, as might be expected. 

I am currently writing out both the updated AnnData object and the SEACells model object (as a pickle) in order to have access to both for the later analysis. 
The model object seems most useful for checking convergence, so it may not really be needed in most cases, but it is relatively small, so easy to keep. 

The wrapper script currently runs only on one sample, and exists mostly just for tracking the commands used for testing.


analyses/metacells/README.md
```diff
@@ -9,20 +9,38 @@ Initially, this module will contain some scripts and notebooks for applying anal
...
+```sh
+bash run-metacells.sh
+```
+
 
 ## Input files
 
 Initial analysis is expected use processed `SingleCellExperiment` objects from the ScPCA Portal, in AnnData format.
 
+For initial testing, files were downloaded with:
```
sjspielman, line 31: Just confirming this code is right - since you wrote _testing_, I wondered if you meant to include the test flag? If not, then not!

analyses/metacells/run-metacells.sh
```diff
@@ -0,0 +1,25 @@
...
+sample="SCPCS000001"
+library="SCPCL000001"
+
+test_file="${data_dir}/${project}/${sample}/${library}_processed_rna.h5ad"
+results_dir="results/${project}/${sample}"
+mkdir -p ${results_dir}
+
+# run the SEACells script
+python scripts/run-seacells.py \
+    $test_file \
```
sjspielman, line 21:
```suggestion
    ${test_file} \
```

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python3
+
+"""
+Seacells Analysis Script
```
sjspielman, line 4:
```suggestion
SEAcells Analysis Script
```

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,159 @@
...
+Joshua Shapiro
+2024-07-18
```
sjspielman, line 6: This seems like it will get out of date real fast, for example, now.

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,159 @@
...
+# Load modules
+#
+# Load required Python modules at the top of your script
+# We have included the standard `pathlib` module and the `session_info` module
+# that we will be using at the bottom of this notebook to record the versions of
+# the modules used in this analysis.
+#
+# Do not install modules here; only load them with `import` statements.
+# Avoid renaming modules with `as` statements, unless you are performing a
+# standard renaming (e.g., `import pandas as pd`).
```
sjspielman, line 20: left over from template, i assume

```suggestion
```

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    Convert an ScPCA AnnData object to the formatting expected by scanpy and SEACells.
+    """
+    # put the highly variable genes in a column of adata.var
+    adata.var["highly_variable"] = adata.var.gene_ids.isin(
+        adata.uns["highly_variable_genes"]
+    )
+
+    # recompute principal components and umap (adds additional metadata)
+    sc.tl.pca(adata, n_comps=50, mask_var="highly_variable")
```
sjspielman, line 45: any specific reason for 50?

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,159 @@
...
+def run_seacells(
+    adata: anndata.AnnData, cell_ratio: float = 75, verbose: bool = False
+) -> tuple[anndata.AnnData, SEACells.core.SEACells]:
+    """
+    Run the SEACells algorithm on the given dataset.
+
+    Parameters
+    ----------
+    adata : anndata.AnnData
+        An annData object containing the data to run the SEACells algorithm on.
```
sjspielman, line 61:
```suggestion
        An AnnData object containing the data to run the SEACells algorithm on.
```

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,159 @@
...
+        Should contain an X_pca field with the PCA coordinates of the cells.
```
sjspielman, line 62: We might want to formally check that `X_pca` if present. It doesn't seem necessary at this stage, but it might be helpful later, so can't hurt now?

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,159 @@
...
+    cell_ratio : float
```
sjspielman, line 63: Can you add a bit here about where the 75 default comes from?

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,159 @@
...
+        build_kernel_on="X_pca",
+        n_SEACells=n_metacells,
+        n_waypoint_eigs=n_eigs,
+        convergence_epsilon=1e-5,
+        verbose=verbose,
+    )
+    # initialize and fit model
+    model.construct_kernel_matrix()
+    model.initialize_archetypes()
+    model.fit(min_iter=10, max_iter=50)
```
sjspielman, line 89: I think this is the function - any reason to not keep their `max_iter=100`? Is it super slow?
https://github.com/dpeerlab/SEACells/blob/3462c624ffae0df6d3930490f345f00196c3503e/SEACells/cpu.py#L653-L659

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,159 @@
...
+    )
+    parser.add_argument(
+        "--seed",
+        type=int,
+        help="The random seed to use for reproducibility.",
+        default=2024,
+    )
+    parser.add_argument(
+        "--logfile", type=pathlib.Path, help="File path for log outputs"
+    )
```
sjspielman, line 121: I'd mention that this also controls the verbosity of SEAcells

analyses/metacells/README.md
```diff
@@ -9,20 +9,38 @@ Initially, this module will contain some scripts and notebooks for applying anal
...
+
+```sh
+./download-data.py --format "anndata" --projects SCPCP000001
+```
 
 ## Output files
 
-TBD
+Currently, the script outputs three files per library: an updated AnnData object, a pickled SEACells model object, and a log file.
```
sjspielman, line 39: Currently it looks like it only outputs a logfile if one specified

sjspielman (commented):
This looks good to me overall! A few of my comments focus on some choices that the script makes, so more docs may be helpful there. 
In addition, I do think we want to explain somewhere in the docs the values that you had to recalculate to run SEAcells. The main purpose of that is to be clear how the input data for (at least 1 script in) this module slightly differs from the processed objects. 

I'm also think it makes sense to activate the module GHA before developing further, but that should probably be in a separate PR.

analyses/metacells/README.md
jashapiro: Which it is in the run script!

analyses/metacells/run-metacells.sh
jashapiro: Sure, but not really important here, as there is not really any ambiguity in the variable.

analyses/metacells/scripts/run-seacells.py
jashapiro: Intent here is to keep it the same as the values already present, but I will make that explicit and check.

analyses/metacells/scripts/run-seacells.py
jashapiro: I'm going to modify this to actually call the convert function from within this function.

analyses/metacells/scripts/run-seacells.py
jashapiro: I don't think it is necessary for the argument definition. The point is that if you want logs I will put all the info there. If you don't want logs, the verbosity is just annoying.

jashapiro (commented):
Thanks for the review. I will be implementing many of your suggestions

analyses/metacells/README.md
sjspielman: yeah, fair! Since I was mostly reviewing python, "the script" meant the python script to me, but that's not what it means here! 👍

jashapiro:
I think this is pretty much ready for another review. I briefly tried activating the action, but I ran into trouble which I expect is related to small samples sizes. I will investigate those as part of a separate PR when I activate the action for real.

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,168 @@
...
+    # check if PCA and metadata were already calculated with scanpy
+    pca_present = (
+        "X_pca" in adata.obsm and "pca" in adata.uns and "variance" in adata.uns["pca"]
+    )
+
+    # recalculate PCA if not present
+    if not pca_present:
+        if "X_pca" in adata.obsm:
+            n_comps = adata.obsm["X_PCA"].shape[1]
```
sjspielman, line 39: I assume?
```suggestion
            n_comps = adata.obsm["X_pca"].shape[1]
```

analyses/metacells/scripts/run-seacells.py
```diff
@@ -0,0 +1,168 @@
...
+        elif "X_PCA" in adata.obsm:
+            n_comps = adata.obsm["X_PCA"].shape[1]
+        else:
+            # set PCA to default 50, or min dimension if smaller
+            n_comps = min(50, adata.n_vars, adata.n_obs)
+        sc.tl.pca(adata, n_comps=n_comps, mask_var="highly_variable")
+
+    # recompute principal components and umap
```
sjspielman, line 47: left over comment?

sjspielman (approved):
LGTM! Left a few small comments but nothing I need to see again. Makes sense to deal with GHA separately, as well.

PR #659, opened 2024-07-24
Add running `SingleR` to annotation workflow 

### LINKED ISSUE #640: Create shell script to run SingleR on all samples

#292 

We need a way to run SingleR on all the samples in SCPCP000015 and output reports to evaluate results. 

This will contain a shell script that creates the reference object and then runs the SingleR script and renders the SingleR template on all samples in SCPCP000015. 

We should also add this workflow to the GHA for this module and make sure things work with the test data. 

No

No 

Once #634 is wrapped up.

allyhawkins:
Closes #640 

Here I am adding the ability to run `SingleR` and generate the template report for all samples in SCPCP000015. I thought about making a separate workflow for this. However we do utilize the output from `AUCell` to build the reference for `SingleR`, so I think it actually made more sense to extend the existing `aucell-annotation.sh` workflow to include running `SingleR`. This allows us to also keep all the outputs together for each library. 

I renamed the existing `aucell-annotation.sh` workflow to now be `aucell-singler-workflow.sh`, which includes first identifying tumor cells with `AUCell`, using those calls to build a reference for tumor cells, and then running `SingleR`.  

- The only new changes to the workflow file are in the section on SingleR and then I added some checks for existing AUCell and SingleR results to help speed things up when re-running this. 
- I moved all the scripts and template notebooks that were previously in the `singler-workflow` to be in a folder, `aucell-singler-workflow` and updated the numbering to reflect the order they are in the shell script. 
- Documentation has been updated to reflect the new name and order of scripts. 
- I tested this with the test data (and updated the GHA to run the script with the new name) and had to make one change to the notebook to accommodate the scenario of having either all normal or all tumor cells. 

I will note that some of these samples have a lot of cells so `SingleR` is kind of slow (~ 1 hour to run with 4 cpus). So I wanted to get things set up to run on LSfR rather than having to do it locally. Those vcs don't have access to internet, so I did make a few changes to read in the celldex and cell ontology references if the file exists rather than trying to access them. I save them to the `scratch` directory and only read in if the file exists, otherwise they are still downloaded. 

Yes 

### Results

I haven't run this through on all samples yet, but when that happens I will sync them to S3. 

This should run without issue using 4 CPUs locally it's just going to take a long time. I was able to run the test data through the whole workflow locally. 

I was able to get this to run successfully on LSfR for a few samples, but ultimately my session keeps timing out and it doesn't keep things running, so I have not run the entire thing through yet. Once the session is timed out the computer is automatically stopped and I have to restart it. It's also possible that there was an error, stopping the CPUs and shutting down the computer. But I wouldn't know if there was an error because when that happens you have to restart and opening up the terminal opens up a fresh terminal, not where you left off. I'll try again and see if I can get it to run for longer... just wanted to share my experience. 

- Are we okay with my decision to combine this with the AUCell workflow? 
- I added some spelling and styling pre-commit hooks to try and avoid those issues in this module in the future, so there's some updates in the README that are style, sorry about that! I can also revert that commit if need be. 


allyhawkins:
I'm at a little bit of a loss here with why this keeps failing. It looks like `celldex` is having trouble downloading the reference files? 

I'm seeing this error message: 
Screenshot 2024-07-25 at 12 23 35 PM

Which correspond to these lines of code: 
```
# grab celldex references
if (!file.exists(blueprint_file)) {
  blueprint_ref <- celldex::BlueprintEncodeData(ensembl = TRUE)
  readr::write_rds(blueprint_ref, blueprint_file)
} else {
  blueprint_ref <- readr::read_rds(blueprint_file)
}

if (!file.exists(hpca_file)) {
  hpca_ref <- celldex::HumanPrimaryCellAtlasData(ensembl = TRUE)
  readr::write_rds(hpca_ref, hpca_file)
} else {
  hpca_ref <- readr::read_rds(hpca_file)
}

```

I did get the workflow for SingleR to run through on all samples locally though!

jashapiro:
> I did get the workflow for SingleR to run through on all samples locally though!

That is good at least. The link in those screenshots does give a not found error, so something there is going wrong. I wonder if there might be something that was fixed in some of those renv updates that you added here. Maybe break those out into a separate PR, get the docker image to build, then retry this?

analyses/cell-type-ewings/aucell-singler-annotation.sh
```diff
@@ -0,0 +1,168 @@
...
+
+        # define library ID
+        library_id=$(basename $sce_file | sed 's/_processed.rds$//')
+
+        # define library specific input/output files
+        geneset_results="${sample_results_dir}/${library_id}_gene-set-scores.tsv"
+        singler_results="${sample_results_dir}/${library_id}_singler-classifications.tsv"
+
+        # only run SingleR if results don't already exist
+        if [[ ! -f $singler_results ]]; then
```
jashapiro, line 142: I'd probably remove this `if`; I would be a bit worried about accidentally not updating these. I suppose you could add something to check if the ref file was modified more recently than the output, but at that point you are basically getting into writing a workflow manager. If you really think you need those features, I would probably go ahead and just turn this into a Nextflow workflow (with minimal config so it just runs locally)!

analyses/cell-type-ewings/scripts/auc-singler-workflow/04-run-singler.R
```diff
@@ -61,15 +76,32 @@ if (opt$threads > 1) {
...
 
 # grab celldex references
-blueprint_ref <- celldex::BlueprintEncodeData(ensembl = TRUE)
-hpca_ref <- celldex::HumanPrimaryCellAtlasData(ensembl = TRUE)
+if (!file.exists(blueprint_file)) {
+  blueprint_ref <- celldex::BlueprintEncodeData(ensembl = TRUE)
+  readr::write_rds(blueprint_ref, blueprint_file)
+} else {
+  blueprint_ref <- readr::read_rds(blueprint_file)
+}
```
jashapiro, line 97: How much time is this saving? I would expect built-in caching in `celldex` and `ensembldb` to handle most of this most of the time, but maybe I am wrong about that?

analyses/cell-type-ewings/scripts/utils/tumor-validation-helpers.R
```diff
@@ -122,6 +122,7 @@ create_marker_gene_df <- function(
 # For any genes that are in the specified `cell_type`, sum of the logcounts is calculated
 # output is a data frame with barcodes and `{cell_type}_sum`
 calculate_sum_markers <- function(marker_genes_df,
+                                  sce,
                                   type) {
```
jashapiro, line 126: Just noting that I think `styler` might complain about this tab pattern, so you might want to run it here to save some future revisions.

jashapiro (approved):
Overall this looks good and it does seem like the last run worked. I just had some questions about whether you really need to be saving reference data files or if the built-in caches are sufficient. Other than that, and if it is still passing, it looks good to me.

analyses/cell-type-ewings/scripts/auc-singler-workflow/04-run-singler.R
allyhawkins: This wasn't about saving time, it was about being able to run this on LSfR. I was having trouble getting it to download the celldex packages so I saved the file to allow for reuse, but I was also using the old Docker image, so perhaps that was part of the problem and this isn't needed anymore.

analyses/cell-type-ewings/aucell-singler-annotation.sh
allyhawkins: This was really just to save time whenever I had to restart because of a failure. I can remove this before merging in.

analyses/cell-type-ewings/scripts/utils/tumor-validation-helpers.R
allyhawkins: I have `styler` set up in my local pre commit file and I just ran it again just on this file and it didn't make any changes.

PR #660, opened 2024-07-25
Add ensembldb to lock file and Docker

allyhawkins:
It looks like the failure in https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/10084373932/job/27883026469 is because we are missing `ensembldb` in the docker image. I think that is a dependency of `celldex` since we are setting `ensembl=TRUE` and it doesn't appear to get automatically captured. I added it to the `dependencies.R` sheet and then updated the lock file here.

### DIFF
--- a/analyses/cell-type-ewings/components/dependencies.R
+++ b/analyses/cell-type-ewings/components/dependencies.R
@@ -1,2 +1,3 @@
 # R dependencies not captured by `renv`
 # library("missing_package")
+library(ensembldb)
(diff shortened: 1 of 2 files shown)


analyses/cell-type-ewings/components/dependencies.R
```diff
@@ -1,2 +1,3 @@
 # R dependencies not captured by `renv`
 # library("missing_package")
```
jashapiro, line 2: Now this is populated, you can remove the example comment
```suggestion
```

jashapiro (approved):
lgtm, assuming actions pass.

PR #661, opened 2024-07-25
Fix failed GHA for building and pushing docker images 

allyhawkins:
I was seeing an error in the action to build and push the docker images after merging and it looks like we might just be missing the `$()` around the call to actually assign the module variable. 

This is from the failed ewings build on the PR I just merged, but it looks like some other builds failed after merging yesterday too. 

Screenshot 2024-07-25 at 10 34 17 AM


jashapiro (approved):
oops

PR #664, opened 2024-07-25
Add module results FAQ

### LINKED ISSUE #515: Consider adding FAQ about using results from modules that aren't completed
> I do think we want to say something about using results from other modules that are _not_ in the workflow yet. I'm not sure we have a specific rec here, but it is something we should address. Essential, I expect an FAQ of "What if the results I want to use are from a module that isn't in the workflow yet?"

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/503#discussion_r1629497867_

sjspielman:
Closes #515 

This PR adds an FAQ about using results from existing modules. It's pretty short and sweet to start, but there are additional considerations we might add (or add them elsewhere?). Here's what I thought of that we might add (can you think of more?):
- How to handle paths when running a module locally for use. For example, symlink from the `data/current/results` directory in some way?
- How to handle GHAs: If they need an in-progress module's results, they'll need to run that module in the testing GHA too. I didn't get into the weeds here to start b/c I figure that's more of an internal consideration for us. How transparent would we want to be?


jashapiro (approved):
I think the general level of detail here is about right, but I would add just a bit more about the paths, especially noting that depending on where you are it might be different. I would _not_ recommend saying anything about links though, as that is likely to cause trouble no matter what people do. 

As for GHA, I agree that it does not belong in this FAQ, and moreover I don't necessarily think we know the answers yet.

PR #665, opened 2024-07-25
Remove prettier from precommit recommendations

### LINKED ISSUE #663: Update optional precommit recommendation to use prettier
In our pre-commit docs, we recommend `prettier` as an _optional_ hook that advanced users might want to add locally:
https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/e853bd0995a1b6bb5f15a46b2a80883da336113d/docs/technical-setup/environment-setup/setup-precommit.md?plain=1#L109-L113

I've learned some things about this hook such that maybe we don't want to recommend it anymore:

1. It is now [now read-only](https://github.com/pre-commit/mirrors-prettier).
2. I just triggered this hook when updating a docs markdown file, and it broke _a bunch_ of formatting and code chunk rendering, such that I decided to remove it from my local precommit config.

sjspielman:
Closes #663 
does what it says, and uses a great branch name to boot.

### DIFF
--- a/docs/technical-setup/environment-setup/setup-precommit.md
+++ b/docs/technical-setup/environment-setup/setup-precommit.md
@@ -106,11 +106,6 @@ You can add these hooks by adding the following code to your `.pre-commit-local.
     rev: v0.4.0
     hooks:
       - id: style-files
-  # prettier formatter for many other languages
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
🚫💅

PR #667, opened 2024-07-25
Make sure renv is up to date 

allyhawkins:
I had been getting errors that my lock file was out of sync, so I had updated the lock file in #659. I'm breaking those changes out to get the Docker image to build so we can test changes from #659 on the most up to date image.

### DIFF
--- a/analyses/cell-type-ewings/renv.lock
+++ b/analyses/cell-type-ewings/renv.lock
@@ -595,7 +595,7 @@
     },
     "IRanges": {
       "Package": "IRanges",
-      "Version": "2.38.0",
+      "Version": "2.38.1",
       "Source": "Bioconductor",
       "Repository": "Bioconductor 3.19",
       "Requirements": [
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
LGTM, but I think you probably want the dependency update in here as well.

allyhawkins:
> LGTM, but I think you probably want the dependency update in here as well.

I think it is in here? You're talking about the change in #660 right? This branch has the `library(ensembldb)` call and I opened the project locally and re-ran `renv::snapshot()` and it doesn't appear to update anything additional.

jashapiro:
> > LGTM, but I think you probably want the dependency update in here as well.
> 
> I think it is in here? You're talking about the change in #660 right? This branch has the `library(ensembldb)` call and I opened the project locally and re-ran `renv::snapshot()` and it doesn't appear to update anything additional.

Oh, right!

PR #670, opened 2024-07-26
Add metacells to GHA and handle small data

### LINKED ISSUE #669: Activate GHA for metacells and fix edge case failures

_No response_

The metacells action does not currently run test on pull requests, and it would be good if it did! 

This follows https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/658, where a brief attempt was made to activate the action, but something in the test data was causing failures. My guess is that it was the metacell ratio, where reducing to one cell caused issues (as it should have!) We will probably need to add a minimum metacell number cutoff, and perhaps other checks for extremely small datasets with few cells.

An activated GHA and modified scripts to deal with whatever was causing initial failures with test data!

_No response_

_No response_

_No response_

jashapiro:

Closes #669 (probably)

Here I am adding the metacell module to the github actions, with just a very simple test at this point of only a single sample. 

The primary error was, as I expected, that with 100 cells in a test dataset and a default ratio of 75, we were telling SEACells to create 1 cell, which caused failure. To fix this, I added minimum cells argument, which I set at 20, arbitrarily. This applies to both the output number of metacells and the minimum input number.

For now, I am just having the function throw an error if there are fewer than 20 input cells, but I catch that in the outer script and simply don't modify the input object. Since I don't yet know what the final output of this module will be, this is likely to change in the future. 

Many.

The purpose here is to run in CI, so there should not be particular requirements at this stage.

Do the defaults seem generally reasonable? I chose a pretty arbitrary cutoff! I kind of expect this to change in the future. 

### Author checklists

_Check all those that apply._


analyses/metacells/scripts/run-seacells.py
```diff
@@ -74,10 +80,13 @@ def run_seacells(
     SEACells.core.SEACells
         The SEACells model object
     """
+
+    if adata.n_obs < min_cells:
+        raise ValueError("The dataset must have at least 20 cells to run SEACells")
```
sjspielman, line 85:
```suggestion
        raise ValueError(f"The dataset must have at least {min_cells} cells to run SEACells")
```

sjspielman (approved):
The arbitrary n=20 doesn't bother me much, tbh. I might add some kind of comment for future us that this value might change in the future, but I also see how that comment could get lost in the shuffle and not have much of an effect, so I don't feel strongly. 

The main question I have though is how testing moving forward is going to work. Since the test data is all <=100 cells, will running SEAcells on the test data actually be sufficient to get good code coverage when testing? I don't feel this question should hold up the PR so happy to approve now, but it's a thing I am wondering as this module begins to be built up.

jashapiro:
> The arbitrary n=20 doesn't bother me much, tbh. I might add some kind of comment for future us that this value might change in the future, but I also see how that comment could get lost in the shuffle and not have much of an effect, so I don't feel strongly.
> 
> The main question I have though is how testing moving forward is going to work. Since the test data is all <=100 cells, will running SEAcells on the test data actually be sufficient to get good code coverage when testing? I don't feel this question should hold up the PR so happy to approve now, but it's a thing I am wondering as this module begins to be built up.

I'm not sure I expect a problem. The modification is deliberately not designed to skip functional code; just to change the output number of cells.

PR #683, opened 2024-08-01
Assign clusters to SCPCL000822 and SCPCL000824

### LINKED ISSUE #402: Cluster Ewing sarcoma samples - SCPCL000822 and SCPCL000824

#292 

Right now the Ewing sarcoma samples on the Portal contain clusters that were calculated using the default clustering in `scpca-nf`. These clusters were obtained using `scran::clusterCells()` with Jaccard Louvain clustering and a nearest neighbors parameter set to 20. These may not be the most optimal cluster assignments for these specific sample, so we should identify the best parameters for this project. 

- One option is to use the [`bluster::clusterSweep`](https://rdrr.io/github/LTLA/bluster/man/clusterSweep.html) function to evaluate the best parameters for clustering. 
- Another option is to try [`Dbscan`](https://rdrr.io/github/LTLA/bluster/man/DbscanParam-class.html) for clustering. 

We should also think about if we want to pick parameters that work across all samples in the project or do this for each individual sample. We also will want to identify some metrics that can be used to evaluate the clustering. 

This PR will contain a template notebook that evaluates clustering for a single sample. We should then run this template notebook across multiple samples to identify the best parameters. 

Any new R packages that are used will be added to the `renv.lock` file. 

_No response_

_No response_

allyhawkins:

Closes #402 

Here I am starting the process of assigning clusters to the Ewing sarcoma samples. I started with two that we have been working with for examples, SCPCL000822 and SCPCL000824. I didn't want to do just one since SCPCL000822 is abnormal for these samples where we actually see a good mix of tumor and normal cells. So I included both samples to have some comparison of how this approach works across more than one sample. 

Generally I am aiming to: 

- Get cluster assignments for these samples so that we can use them in downstream analysis. In particular I expect tumor and normal cells to cluster separately. 
- Get cluster assignments that can be used to help define subpopulations of tumor cells. I expect that tumor cells that cluster differently from other tumor cells are potentially part of different sub-types of tumor cells. This will help us classify the different tumor cell phenotypes/ cell states that may be present. 
- Obtain general cell type assignments based on the cluster. Any clusters that have high tumor marker gene expression over other clusters should be tumor cells and then any clusters with high expression of specific normal cell markers are labeled as that normal cell. For example, there are clusters with high endothelial cell marker gene expression, so we can label the cells in that cluster as endothelial. 

I am starting by just adding an exploratory notebook that looks at clustering using different values of k and identifies the most "optimal" value of k to use for cluster assignments. I anticipate eventually turning this into something that can be applied across multiple samples, but for now I just wanted to play around with clustering and figure out what metrics and plots might be useful in identifying the best value of k. Additionally, I wanted to see if we could identify tumor cell clusters and normal cell clusters based on the chosen cluster assignments. 

This notebook includes: 

- Louvain, Jaccard clustering across a range of k (5-40, with step size of 5). Here I use the `clusterSweep` function mentioned in [OSCA](https://bioconductor.org/books/3.18/OSCA.advanced/clustering-redux.html#clustering-parameter-sweeps). 
- Metrics to evaluate this clustering: 
  - [Silhouette width](https://bioconductor.org/books/3.18/OSCA.advanced/clustering-redux.html#silhouette-width)
  - [Cluster purity](https://bioconductor.org/books/3.18/OSCA.advanced/clustering-redux.html#cluster-purity)
  - [Cluster stability](https://bioconductor.org/books/3.18/OSCA.advanced/clustering-redux.html#cluster-bootstrapping). Here I used a lot of the [code we had set up in `scpca-downstream-analysis`](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/60d265d971582edff54af5f06d956c34d686706f/utils/clustering-functions.R#L171). 
- I then compared the cluster assignments to the cell type assignments obtained by running our SingleR workflow (`aucell-singler-annotation.sh`). This is shown by looking at the Jaccard index between the two methods, where I expect to see that clusters mostly contain one cell type or groups of similar cell types (e.g., fibroblasts and muscle cells together).  
- Looking at the above stats, I chose to use k=25 for the remaining analysis. This included: 
  - Plotting the distribution of marker gene expression across clusters for tumor, endothelial, immune, and MSC like cells. Here we see that each cluster shows high expression of one of these marker gene lists compared to all other clusters. 
  - I assigned cell types based on expression of the marker genes in each cluster. This means that each cluster is a specific cell type. I don't know that we want to keep that as our final annotations, but I think at minimum clusters with tumor marker gene expression should be "tumor", while normal cells could be labeled with the more general label or with their `SingleR` annotation. 
- I then exported the cluster assignments, cluster cell type, and consensus cell type between clustering and SingleR. 

Sure do. 

This is just a rendered notebook, but if you want to be able to render it yourself, you will need to download the `aucell_singler_annotation` results for these two samples/libraries. 

These are stored in `s3://researcher-211125375652-us-east-2/cell-type-ewings/results/aucell_singler_annotation`. 

For these two samples I felt pretty confident about the K that I chose and I feel like the clusters look fairly good? There seems to be nice separation between them. 

Because of this I think at minimum we should label all cells that belong to any clusters with high tumor marker gene expression as tumor. If cells don't belong to those clusters then they should not be tumor cells and should belong to a normal cell group. 

Here's a rendered copy of the report for easy review: 
[05-cluster-exploration.html.zip](https://github.com/user-attachments/files/16458866/05-cluster-exploration.html.zip)

This can be run locally just by rendering, but the cluster stability calculations do take ~ 10 minutes. 

How do we feel about using these metrics to evaluate clusters? I am thinking we might want to make a template notebook that reports these metrics for each sample and outputs all cluster assignments. Then in follow up notebooks, we can choose which parameter provides the "best" clusters and use that to refine our tumor cell assignments and assign tumor cell states. I would focus on the clustering metrics as the main part of review so we can determine how best to apply this to other samples. 

In the original issue (#402) there was mention of using density based clustering and DbScan. I did try this and honestly the clusters did not look good. I also played around with some consensus clustering methods (like [SC3](https://github.com/hemberg-lab/SC3)) but was hitting some pretty cryptic errors just getting it to run. I didn't feel like it was worth pursuing at this point. 

### Author checklists

_Check all those that apply._


analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+# source in helper functions: plot_gene_heatmap() and plot_cnv_heatmap()
+# create_classification_df() and create_marker_gene_df()
+validation_functions <- file.path(module_base, "scripts", "utils", "tumor-validation-helpers.R")
+source(validation_functions)
+```
+
+## Define functions 
+
+```{r, assign_clusters_functions}
+# get clusters for a specified value of k
```
sjspielman, line 71: Expand this "docs" a bit here that it's louvain w/ jaccard since `k` means many different things in clustering land!

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+    # remove combined params column
+    dplyr::select(-params)
+
+  return(cluster_df)
+}
+```
+
+
+```{r, cluster_stats_functions}
+# cluster statistics functions
```
sjspielman, line 124:
```suggestion
# cluster statistics functions


```

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+  }
+
+  ari_df <- data.frame(
+    ari = ari,
+    k_value = k
+  )
+}
+
+# get cluster stability for each nn_param cluster results are available for
+# calculated the ari between original clusters and shuffled clusters 20 times
```
sjspielman, line 186: this was left over from copy/paste probably 
```suggestion
```

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+
+```{r}
+# source in helper functions for make_jaccard_matrix() and jaccard()
+jaccard_functions <- file.path(module_base, "scripts", "utils", "jaccard-functions.R")
+source(jaccard_functions)
+
+# source in helper functions: plot_gene_heatmap() and plot_cnv_heatmap()
+# create_classification_df() and create_marker_gene_df()
+validation_functions <- file.path(module_base, "scripts", "utils", "tumor-validation-helpers.R")
+source(validation_functions)
```
sjspielman, line 65: It doesn't look like these functions are used? 

```suggestion
```

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+  }) |>
+  dplyr::bind_rows(.id = "library_id") |>
+  dplyr::left_join(all_cluster_results, by = c("library_id", "barcodes"))
+```
+
+Below we visualize the cluster assignments using each parameter on a UMAP. 
+This isn't particularly useful because of clusters, but it can be helpful to see any values of K that may have obvious over or under clustering. 
+
+```{r, fig.height=15, fig.width=10}
+# look at clustering results across all params
```
sjspielman, line 357: For this plot you're going to need to tweak the factors, since `k_5` is last because the computer thinks 5>40 😞

sjspielman: Actually also, this plot is really hard to take in because it's so long. Why don't we just plot the two datasets separately so each has its own `facet_wrap`? We aren't comparing dataset UMAPs directly to each other anyways and a shared legend isn't particularly meaningful.

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+```
+
+```{r}
+# summarize the silhouette width and purity for each nn_param used
+summary_stats_df <- all_stats_df |>
+  dplyr::group_by(library_id, nn_param) |>
+  dplyr::summarise(
+    mean_width = mean(width),
+    mean_purity = mean(purity)
+  ) |>
```
sjspielman, line 394: I'd like to see some error lines in these plots too to assess overlap, if we're plotting means (that said, did you try maybe a `sina` with a red mean dot like the downstream analysis clustering report?). On one hand I'd say toss an a standard error calculation here too, but I actually you can probably skip a lot (all?) of this wrangling using `stat_summary()` which should plot your mean & SE automagically.

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+  dplyr::bind_rows(.id = "library_id") |>
+  dplyr::mutate(
+    k_value = as.factor(k_value),
+    # make sure that k = 5 comes first
+    k_value = forcats::fct_relevel(k_value, "5", after = 0)
+  )
+
+# plot stability across all values of k
+ggplot(stability_df, aes(x = k_value, y = ari)) +
+  geom_point() +
```
sjspielman, line 441: just so we can make sure we're seeing the points, and I suggested my personal favorite width parameter

```suggestion
  geom_jitter(width = 0.1) +
```

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+```
+
+
+```{r, fig.height=10, fig.width=7}
+# get heatmap for each library showing cluster assignments vs. singler assignments
+names(sce_list) |>
+  purrr::map(\(id){
+    cluster_celltype_heatmap(
+      cluster_classification_df,
+      id
```
sjspielman, line 485: need to set factors for this plot, 5 is last again.

sjspielman: I wonder if we should turn off (at least y-axis) clustering for these heatmaps so that the cell labels are always in the same order. I think that will make it easier to interpret.

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+    )
+  })
+```
+
+From these plots, it's obvious that K = 5 is over clustering and cell types are spread across a lot of clusters. 
+However, for the other values of K, there doesn't seem to be any striking difference that are obvious by eye.
```
sjspielman, line 491:
```suggestion
However, for the other values of K, there don't seem to be any striking differences that are obvious by eye.  
```

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+umap_df <- umap_df |>
+  dplyr::select(library_id, barcodes, UMAP1, UMAP2) |>
+  dplyr::distinct()
+
+# assign clusters based on marker gene expression and cell types that are prominent in that cluster
+# use the jaccard matrices to help guide assignments
+cluster_umap_df <- cluster_classification_df |>
+  dplyr::left_join(umap_df, by = c("library_id", "barcodes")) |>
+  dplyr::filter(nn_param == "k_25") |>
+  dplyr::mutate(
```
sjspielman, line 536: Somewhere it is worth noting (comments? bullets? you decide!) which clusters are _not_ being assigned. Looks like 10&11 for `SCPCL000822` and 9 for `SCPCL000824`.

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+  scale_color_brewer(palette = "Dark2")
+```
+
+We can also validate that these cells do in fact have higher expression of the marker genes as compared to other cells by plotting the marker gene expression on UMAPs. 
+
+```{r}
+# grab columns that contain marker gene sums
+marker_gene_columns <- colnames(cluster_umap_df)[which(endsWith(colnames(cluster_umap_df), "_sum"))]
+
+# create individual density plots and combine into one
```
sjspielman, line 576:
```suggestion
# create individual umaps and combine into one
```

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+      library_id == "SCPCL000824" & cluster == 7 ~ "mesenchymal-like cell",
+      # if they didn't have high expression of any markers, we label them as unknown
+      .default = "unknown"
+    ),
+    # make tumor first
+    cluster_celltype = forcats::fct_relevel(cluster_celltype, "tumor", after = 0)
+  )
+```
+
+```{r}
```
sjspielman, line 560: I'd make this plot bigger, maybe move legend to bottom as needed for spacing. Can't directly suggest changes to the code chunk though since too many backticks...

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,626 @@
...
+# look at cell type assignments on the UMAP
+ggplot(cluster_umap_df, aes(x = UMAP1, y = UMAP2, color = cluster_celltype)) +
+  geom_point(alpha = 0.5, size = 0.5) +
+  facet_grid(cols = vars(library_id)) +
+  theme(aspect.ratio = 1) +
+  guides(color = guide_legend(override.aes = list(alpha = 1, size = 1.5))) +
+  scale_color_brewer(palette = "Dark2")
+```
+
+We can also validate that these cells do in fact have higher expression of the marker genes as compared to other cells by plotting the marker gene expression on UMAPs.
```
sjspielman, line 570: It's hard to interpret these plots since they paired by cell type, not by sample. Is it possible to instead have two plots - one for each sample and faceted by cell type? (right now it's 3 plots, one for each cell type but faceted by sample)

sjspielman (commented):
Overall looks really good! To some of your questions, I think k=25 is no unreasonable and the dbscan situation is fine imo. Let me know if any of the inline comments I left are unclear!

The main review I have is more conceptual/organizational since I'm not sure I precisely follow the exact next steps. Related to this comment you made:

> How do we feel about using these metrics to evaluate clusters? I am thinking we might want to make a template notebook that reports these metrics for each sample and outputs all cluster assignments. Then in follow up notebooks, we can choose which parameter provides the "best" clusters and use that to refine our tumor cell assignments and assign tumor cell states. I would focus on the clustering metrics as the main part of review so we can determine how best to apply this to other samples.

(First, I feel fine about using the metrics you used). This notebook serves as a "roadmap" for how clustering can be performed and evaluated for each sample. I envision, in the end, this becoming essentially two notebooks per sample (is this what you envision? if not, where did I miss the boat?): 1) a template notebook to perform and evaluate clusters which can be run across all samples. This is everything before you chose k=25., and 2) a sample-specific notebook (which might be based on a template, but should not be fully automated since it will involve more interactive plotting and interpretation) that uses a specific clustering result to explore marker gene expression and/or assign clusters to cell types (the final 2 notebook sections) in the clusters that were chosen from the previous notebook.

How does ☝️ sound relative to what you are thinking? It also seems like this is a special case that 2 samples are being evaluated in the same notebook - how general do you see that being?

allyhawkins:
> (First, I feel fine about using the metrics you used). This notebook serves as a "roadmap" for how clustering can be performed and evaluated for each sample. I envision, in the end, this becoming essentially two notebooks per sample (is this what you envision? if not, where did I miss the boat?): 1) a template notebook to perform and evaluate clusters which can be run across all samples. This is everything before you chose k=25., and 2) a sample-specific notebook (which might be based on a template, but should not be fully automated since it will involve more interactive plotting and interpretation) that uses a specific clustering result to explore marker gene expression and/or assign clusters to cell types (the final 2 notebook sections) in the clusters that were chosen from the previous notebook.

Yes this is exactly what I was thinking! I see the following PRs coming up: 

- taking part 1 (pre choosing K) and turn it into a template notebook. 
- write a script to render the notebook on all samples in the project 

Then one approach would be to go sample by sample and look at the output, but I would like to figure out a way to at least somewhat automate it. But first the clustering. 

> It also seems like this is a special case that 2 samples are being evaluated in the same notebook - how general do you see that being?

When we convert to a template I want it to only run on one sample. Here I'm looking at 2 mainly because I wanted to make sure what I was doing was going to be applicable on more than one sample before going too far down this road. 

I'll go through your review and make adjustments, but wanted to be sure I addressed these overall questions first.

allyhawkins:
@sjspielman thanks for the helpful comments. I went through and mostly made adjustments to some of the plots based on your comments. 

- The UMAPs are now separated by sample and faceted by K. These should be a lot easier to interpret now! 
- For all the cluster statistics I am now showing a sina plot and plotting the median value and IQR. This is what we did in downstream analyses. I think it's harder to see the differences, but also maybe better because the differences are smaller than what the previous plot seemed to portray. 
- I turned off clustering of rows for the Jaccard heatmaps. 
- The last set of UMAPs that looks at marker gene expression is now faceted by cell type with one plot per sample. The only problem here is that I think it's hard to see the gene expression for some of the marker gene sets since tumor expression is so high. But I do agree that it's easier to interpret. 

Here's a copy of the updated rendered report: 
[05-cluster-exploration.html.zip](https://github.com/user-attachments/files/16473284/05-cluster-exploration.html.zip)

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,672 @@
...
+    nn_param = forcats::fct_relevel(nn_param, "k_5", after = 0)
+  )
+```
+
+
+```{r}
+# silhouette width for different params
+library_ids |>
+  purrr::map(\(id){
+    plot_cluster_stats(all_stats_df, width, id)
```
sjspielman, line 430: Can we do a ....

```suggestion
    plot_cluster_stats(all_stats_df, width, id) + ggtitle(id)
```

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,672 @@
...
+      quantile(x, 0.25)
+    },
+    fun.max = function(x) {
+      quantile(x, 0.75)
+    }
+  )
+```
+
+Here we see that cluster stability increases as k increases for both libraries. 
+We see a plateau start to appear around k = 20 and k = 25.
```
sjspielman, line 479: Looking at this again, I actually now wonder if we should be at k=30/35 for `SCPCL000824` with this metric? What do you think?

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,672 @@
...
+  purrr::map(\(id){
+    cluster_celltype_heatmap(
+      cluster_classification_df,
+      id
+    )
+  })
+```
+
+From these plots, it's obvious that K = 5 is over clustering and cell types are spread across a lot of clusters. 
+However, for the other values of K, there don't seem to be any striking differences that are obvious by eye.
```
sjspielman, line 524: im intrigued by the macrophage/cluster 9 interaction for `SCPCL000824`, as well as some of what's going on with cluster 10. Looks like things are bouncing back and forth from k=20 -> k=40 for a decent number of the cells?

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
```diff
@@ -0,0 +1,672 @@
...
+      tidyr::pivot_longer(
+        cols = all_of(marker_gene_columns),
+        names_to = "cell_type",
+        values_to = "gene_expression"
+      ) |>
+      dplyr::mutate(cell_type = stringr::str_remove(cell_type, "_sum"))
+
+    ggplot(plot_df, aes(x = UMAP1, y = UMAP2, color = gene_expression)) +
+      geom_point(alpha = 0.5, size = 0.2) +
+      facet_wrap(vars(cell_type)) +
```
sjspielman, line 629: I wonder if `scales="free"` would help with the color scale, but also it might just do something bonkers, or do nothing at all!...never tried it before for such a circumstance..

sjspielman (commented):
Code-wise I think this is in a pretty good place, so not much to say there! I left a few comments about some of the interpretations, in particular that I'm not sure I'm entirely sold on k=25 for both datasets. (I'd like to hear what you think there before I approve this, but I think we're right about there)

I also have another question (which shouldn't hold this PR up, just wanted to ask, and this question may just be a product of how I haven't been the primary module reviewer!) - when we think about the final.final cell type labels for this module, how do you envision reporting labels derived from clusters and then applied to all cells, vs labels made at the level of individual cells?

analyses/cell-type-ewings/exploratory_analysis/05-cluster-exploration.Rmd
allyhawkins: I think the exact numbers of the clusters here can't really be compared between multiple parameters. We don't actually know that the cells that were in cluster 9 when we cluster with one value of k are the same cells that are in cluster 9 or 10 in another value of k. What I see is there is a cluster that has both macrophages and some cells from the "All remaining cell types" class. With k=25, this is cluster 10 and with k=30 this is cluster 9. 

That being said I do agree that maybe k=30 is a little bit better for SCPCL000824. Mostly because of the increase in purity and increase in stability that is shown between K=25 to 30.

sjspielman: > I think the exact numbers of the clusters here can't really be compared between multiple parameters. We don't actually know that the cells that were in cluster 9 when we cluster with one value of k are the same cells that are in cluster 9 or 10 in another value of k.

Haha of course.. Friday afternoon brain 😄

sjspielman: > That being said I do agree that maybe k=30 is a little bit better for SCPCL000824. Mostly because of the increase in purity and increase in stability that is shown between K=25 to 30.

Do you plan to run these libraries again through the template notebook? If so, I'd leave this one is, and just indicate in the text that 30 could be good too, and save changes for later. Otherwise if this is it for these libraries, I'd just go ahead and use 30.

allyhawkins:
> I also have another question (which shouldn't hold this PR up, just wanted to ask, and this question may just be a product of how I haven't been the primary module reviewer!) - when we think about the final.final cell type labels for this module, how do you envision reporting labels derived from clusters and then applied to all cells, vs labels made at the level of individual cells?

This is a GREAT question that I do not have the answer too yet. I think I see the cluster assignments more as generalized cell types where we categorize them into broader groups, like immune cells, endothelial cells, etc. I think the benefit of the individual cell type assignment with `SingleR` is that we may have some that have more granular assignments, like maybe the immune cell group contains both macrophages and a few other cells that belong to other immune cell types. 

The main thing I think we should get from these cluster assignments is which clusters should be tumor cells. There are some cases where I see tumor cells that were assigned by SingleR that are very closely mixed with normal cells and I think in many of those cases, clustering will help refine which cells should actually be tumor cells vs. normal cells. 

I will say it is nice to see that the cluster assignments seem to match up with what is obtained from SingleR (note this is using a modified reference that contains both normal and tumor cell types). 

SCPCL000822 SingleR assignments 
image


SCPCL000824 SingleR assignments 
image

One other side note here is that by having the cluster assignments and pulling out all the tumor clusters, we can then identify whether each cluster belongs to a specific tumor cell state and provide further classification. 

> Looking at this again, I actually now wonder if we should be at k=30/35 for SCPCL000824 with this metric? What do you think?

I agree that K=30 might be a slightly better choice based on cluster stats, but I tried it and got some very weird and wonky results for the assignments. Pretty much everything is tumor or MSC cells and the macrophages are moved. K=25 tends to line up with the SingleR results better and I do trust those over using only 2-3 marker genes per cell type. I do also plan to run these again through the template since all libraries should be run through that. Maybe we need to dig into the assignments for this sample more thoroughly, but for now I do think the K=25 is the better choice. 

image

Because of this I didn't end up changing anything other than adding that id as the title for one of the plots like you suggested.

analyses/cell-type-ewings/renv.lock
sjspielman, line 1: 👍

sjspielman (approved):
Ok, your comment has convinced me all around! Sounds like for the final annotations, we'll have a more holistic look at all the different analyses across the module to make a lot of case-by-case decisions on labels where needed either way.

PR #688, opened 2024-08-05
Add push trigger to Wilms tumor cell typing Docker build/push workflow & proposed change to template

jaclyn-taroni:

Here, I'm adding a GitHub Actions workflow trigger that will push the Docker images for the `cell-type-wilms-tumor-06` module to ECR upon merge to `main`.

I meant to accomplish this in https://github.com/maud-p/OpenScPCA-analysis/pull/3, but I was (arguably overly!) focused on aligning the changes with the template.

That made me wonder why these commented-out steps are not in the template, so I've added them. I could see an argument for not including them–we want to make sure something is sufficiently mature before pushing–but I don't know that the pull request trigger and push trigger need to be uncommented at the same time.

N/A

What do you think of this change to the template? I admit there maybe should be docs changes as well, but I wanted to discuss this first!

### DIFF
--- a/.github/workflows/docker_cell-type-wilms-tumor-06.yml
+++ b/.github/workflows/docker_cell-type-wilms-tumor-06.yml
@@ -21,6 +21,14 @@ on:
       - "analyses/cell-type-wilms-tumor-06/.dockerignore"
       - "analyses/cell-type-wilms-tumor-06/renv.lock"
       - "analyses/cell-type-wilms-tumor-06/conda-lock.yml"
+  push:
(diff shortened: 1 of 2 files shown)


jashapiro (approved):
I am fine with the change to the template; I think I simply forgot to add it when I was trying different approaches to building and pushing.

jaclyn-taroni:
Thanks! I don't think there are any docs changes required, having reviewed the comments at the top of the YAML + the ensuring reproducibility docs.

PR #695, opened 2024-08-05
Disable second docker build summary 

allyhawkins:
Since we don't need two summaries when building Docker images, this disables the summary for the job that does the pushing.


jashapiro (approved):
LGTM

PR #697, opened 2024-08-06
Style code for periodic releases

sjspielman:
This PR adds code styling into the periodic release issue template, as described in https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/678#issuecomment-2272054896


jashapiro:
I don't think styling is worth including in the periodic releases. At least not on all code. The potential to introduce conflicts in actively developed modules just seems like too much of a hassle. 

We might include it in versioned releases, but even there, I would think we should limit the action to the modules that we think are relatively stable.

sjspielman:
Nope!

Issue #679, opened 2024-08-01
Wilms Tumor Dataset Annotation (SCPCP000006) _ clustering


maud-p:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/635

### Describe the goals of the changes to the analysis module.

The main addition to the module is a RMardown report for three Wilms tumor sample from the dataset SCPCP000006:

- SCPCS000168,
- SCPCS000169,
- SCPCS000170. 

We tested in this report ways to 
- cluster cells
- characterize/identify each cell types

The aim would be to discuss the report and improvement before adapting and rendering it to all samples in the dataset.

The analysis will be as the following:

[0] We build a seurat object based on the counts data and went through the seurat workflow [normalization --> reduction --> clustering] following the Seurat workflow.

[1] We perform some quality check to assess any QC-induced clustering (nFeature, nCount, percent.mito).

[2] We add cell cycle information, as we know that in a specific cell cycle state, the transcriptional program is mostly/exclusively related to cell cycle genes and the identity of cells is difficult to determine. We expect these cells to cluster together in a cluster of proliferating cells.

[3] We look at specific marker genes that we reported in the table marker.sets/CellType_metadata.csv to check the relevance of the clustering.

[4] We look at specific pathways that we reported in the table marker.sets/Pathways_metadata.csv to check the relevance of the clustering.

[5] We run DElegate::FindAllMarkers2 to find markers of the different clusters and manually check if they do make sense. DElegate::FindAllMarkers2 is an improved version of Seurat::FindAllMarkers based on pseudobulk differential expression method.

[6] We perform enrichment analysis of marker genes for each seurat clusters. We defined all the genes from the seurat object as the universe and used the MSigDB gene sets.

[7] We plot pca/umap reduction grouping with available annotations (singler_, cellassign_). We expect at least immune cells to be correctly label and fall into a few set of clusters.

[8] We run label transfer (Azimuth) to transfer annotation from the fetal kidney atlas human reference. We plot pca/umap reduction grouping with latest labels. We expect it to be the most representative of the cell types in the sample.

### What will your pull request contain?

The pull request will contain the rmd file 01-clustering_SCPCS000xxx.Rmd in the cell-type-wilms-tumor-06 folder and the html report in the notebook folder.

The renv.lock file needed to update-build-run the docker container required for the analysis. 

[...]

PR #699, opened 2024-08-07
01 clustering wilms 06

maud-p:

isuue #679 

In this pull request, I would like to introduce one notebook that I use to get a first impression on a sample.
I went from the basic approach (look at known markers), to more advances label transfer or enrichment analysis of marker genes. 
Across the 3 samples I started to look into, I realized some of the steps are crucial, some less, but would be gratefull for your input. 

For this analysis, we worked with the _processed.rds data. We builded a Seurat object based on the counts data and re-perform the analysis [normalization –> reduction –> clustering] following the Seurat workflow.

We transferred meta.data to keep:

- QC data computed by the DataLab
- annotation data computed by the DataLab
- raw annotation and gene_symbol conversion

We perform the following analysis to assess for the quality of clustering and get a first impression on the sample:

[1] We perform some quality check to assess any QC-induced clustering (nFeature, nCount, percent.mito).

[2] We add cell cycle information, as we know that in a specific cell cycle state, the transcriptional program is mostly/exclusively related to cell cycle genes and the identity of cells is difficult to determine. We expect these cells to cluster together in a cluster of proliferating cells.

[3] We look at specific marker genes that we reported in the table marker.sets/CellType_metadata.csv to check the relevance of the clustering.

[4] We look at specific pathways that we reported in the table marker.sets/Pathways_metadata.csv to check the relevance of the clustering.

[5] We run DElegate::FindAllMarkers2 to find markers of the different clusters and manually check if they do make sense. DElegate::FindAllMarkers2 is an improved version of Seurat::FindAllMarkers based on pseudobulk differential expression method.

[6] We perform enrichment analysis of marker genes for each seurat clusters. We defined all the genes from the seurat object as the universe and used the MSigDB gene sets.

[7] We plot pca/umap reduction grouping with available annotations (singler_, cellassign_). We expect at least immune cells to be correctly label and fall into a few set of clusters.

[8] We run label transfer (Azimuth) to transfer annotation from the fetal kidney atlas human reference. We plot pca/umap reduction grouping with latest labels. We expect it to be the most representative of the cell types in the sample.

Yes! Based on the present PR, I would like to refine the approach to annotate the cells in each sample. More PR are coming!

### Results

I haven't send anything to S3. But I am wondering if I should load the fetal kidney reference for easy use of it?
Else, I can load the different steps I used to get the fetal kidney atlas compatible woth azimut. I did it 2 years ago, were some steps with python and R. I could try to clean the docker container used for it and publish it in the module. To give you a feeling what to expect, I added the script I used to generate the fetal kidney reference : 01_fetal_reference_kidney.Rmd

Let me know what do you think!

RMarkdowm - html notebooks

I tested different approaches/tools to get a feeling of the cells composing the samples. 

I think the first clustering is not too bad and could be used for visualization. This was the first aim of the PR :)

I found the most helpfull approach is to transfer label from the healthy fetal kidney reference. 
This allow to identify with high confidency (i) immune cells (normal), (ii) endothelial cells (normal), (iii) stromal cells (cancer+normal) and (iv) nephron cells (blastema+epithelial cancer + normal cells).

The use of known marker genes is usefull to check clustering but cannot be used imho to annotate cells. 

Pathway enrichment analysis using the MSiGDB C8 gene set can be used imho to annotate cells, as cancer cells, blastema or primitive epithelium must be enriched in fetal kidney pathway while healthy/normal kidney should be enriched in mature/adulte differentiated kidney pathways.

To annotate the cells, I would re-evaluate my strategy as the following:
1) label transfer from the fetal kidney reference and label cells with the predicted compartments, i.e. immune (normal), endothelial (normal), stroma (normal+cancer) and fetal nephron (normal+cancer).

2) I would subset the stroma and fetal nephron cells separetly and re-perform clustering step. I expect then normal/cancer subtype to be better segregated in a dataset of stronal cells or epithelial cells only. 
To check the new clustering, I will use a similar approach including marker genes, pathways, label transfer from the fetal kidney atlas. 

3) run inferCNV to have an additional information that can allow to conclude normal versus cancer depending on low versus high number of rearrangments.  

I used RStudio and renv. Th erenv.lock file has been updated!

The strategy for the next step, if you think it makes sense!

### Author checklists

_Check all those that apply._


maud-p:
Thank you!

FYI, I just opened a ney issue (https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/703) and plan to add few steps to allow you and the community to re-build the azimuth compatible kidney reference. 
In the present PR, this is missing, as I wasn't able to upload the fetal kidney reference (files too large).

sjspielman (commented):
Hi Maud, 

Nice to e-meet you, I'm one of the Data Scientists at the CCDL :wave:. Glad to start working with you on your contribution!

It looks like you've made some great headway into exploring these samples and getting the analysis started. You've also provided a lot of helpful details in your opening PR comment, so thanks for taking the time to explain everything! 

Generally speaking, I think you've put together some solid initial analysis that can be built on more generally for all samples. Although it can be tempting (reasonably so!) to jump right into the science, for this kind of large analysis we'd probably like to get some of the code nailed down before we really dive into interpreting each sample. This way, we can develop code we're confident will flexibly work across the full set of samples, and then we can proceed to quickly run through a "pipeline" all in one go. 

Bearing this in mind, I'm going to recommend that we take a _bit_ of a step back - and also a step forward all at once! One of the main things that I see in your code is that you have three separate notebooks which contains quite a bit of shared code. Given this, it would be great to modularize the code by instead starting off with a template notebook. Essentially, this will be very similar to the notebooks you've written, but it will be agnostic about which sample is being analyzed, and we can use an R Markdown parameter to pass in which sample to process (and export clustering results for). This means you can still include much (but not all) of the code in the template and then run it on the same three samples, but you wouldn't have the same kind of interspersed text interpretation since it would be automated.

There are a few reasons I make this suggestion -  when performing complex analyses, and in particular on multiple samples!, we're going to be writing the same code over and over again. This can often lead to longer-term problems with code maintenance and reproducibility. For example, if you have the same line of code in 5 files and realize you need to change that line, odds are low that you (and I speak from experience...!) will remember each time you used that line of code and change them all. Having less repeated code not only makes it easier for you to write and maintain your code, but in the longer term it will also make it easier for others to run and understand your code. And, in between, "refactoring" some of this code into a template with fewer repeated pieces of code will make it easier for me to review. 

Something I want to emphasize is that I am not necessarily asking you to change any code, just rearrange such that we end up with _one_ notebook that could be used for any sample. This applies even to any clustering analyses or plots you don't necessarily think will be relevant to final clustering, or plots that you might not think are the most "informative." I'm looking at this aspect of your PR comment:

> I tested different approaches/tools to get a feeling of the cells composing the samples.

What I'm saying is, please keep those different approaches/tools you've already used! A template notebook to get a sense of the data, no matter what moves onto the next stage of analysis, is great.
 

#### Next steps

- For now, let's leave this PR open but pivot to creating a different one in parallel with a template notebook. For this, I'd create a new branch (off of `main`) focuses only on transitioning your current code into a template notebook that can be generally run on any sample (to be clear that will be a fully separate branch from the one you used for this PR). 
	- You can then file a new PR from that branch. At that point, I can begin to review the code more carefully, and we can start to think more carefully about next steps and all the good science stuff.
- You might be interested in referring to this [template notebook which @allyhawkins wrote for the `cell-type-ewings` module to perform clustering](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/559d739f49dfcfc48528b22f60c2d871b02ab0b9/analyses/cell-type-ewings/template_notebooks/clustering-workflow/01-clustering.Rmd). You can see how she has used R Markdown parameters to pass in the input sample information that the code then uses. Of course, your code will be rather different from hers - this is only an example to help show what I mean by "template notebook"! 
	- Some of the coding/organizational strategies that she uses in this notebook are outlined [in our documentation for writing notebooks](https://openscpca.readthedocs.io/en/stable/contributing-to-analyses/analysis-modules/notebook-structure/), which you might also find useful.
- Similar to the [`cell-type-ewings` module](https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-ewings), I would save that notebook in a folder called `template-notebooks` as, maybe `01_perform-clustering.Rmd` (or something like that!).
- Also, for now, I would also hold off on the fetal kidney reference code/results. I would expect that to actually be its own PR (or even 2: One to add the code that creates the reference, and one to use the reference in a given analysis). 
	- That means this code would not end up in the template notebook. But, that certainly doesn't mean we can't update the template notebook in a future PR to include analysis with that reference! It's just all part of the step-wise process :) 

In the new PR you file, you will likely have the following changed files:

- Your new template notebook in `template-notebooks`
- Associated [documentation](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/analysis-modules/documenting-analysis/) for the notebook
- An updated `renv.lock` file to cover the notebook's dependencies
- [Maybe] HTMLs from running the notebook on a few samples
	- _That said_ we can also talk about the best way to generate these once you file your PR too. Ideally, you might use a shell script to call the notebook and render it for certain datasets of interest. If you want to just run them "manually" (e.g. just writing in different `param` values for the sample to proceed) to start that's totally fine, and we can work towards making it reproducible in a separate PR!

Again, please let me know how I can help with this process, where I can provide more feedback, or what I can clarify from what I've written above! Thanks again for your work towards OpenScPCA :)

maud-p:
Thank you @sjspielman for your detailed comments and advice!
It is all clear and I expect to fill the discussed PR either today or tomorrow :)

PR #704, opened 2024-08-08
SCPCP000006 - Add 01-clustering template analysis

maud-p:
#### Please link to the GitHub issue that this pull request addresses.

https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/635

#### What is the goal of this pull request?

As discussed in the previous PR (https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/699) I wanted to have one notebook that is rendered over the entire Wilms tumor dataset.

#### Briefly describe the general approach you took to achieve this goal.

1. I basically took the previous .Rmd file, 
2. remove all the comments specific to one sample, 
3. remove the label transfer using the fetal kidney atlas (for the next PR)
4. write a R script to render the notebook on all samples. 

#### If known, do you anticipate filing additional pull requests to complete this analysis module?

Yes! More than one. At least:

- [ ] A PR to build the fetal kidney reference for Azimuth,
- [ ] A PR to include the label transfer from the fetal kidney reference to the clustering template.
- [ ] A PR to include inferCNV and/or copyKAT to the template.
- [ ] A PR to subdivide and re-cluster stroma and fetal nephron cells, taht contain nromal + cancer cells

#### What types of results does your code produce (e.g., table, figure)?

HTML reports 

### Provide directions for reviewers

_In this section, tell reviewers what kind of feedback you are looking for._
_This information will help guide their review._

#### What are the software and computational requirements needed to be able to run the code in this PR?

_This information will help reviewers run the code during review, if applicable._
_For software, how should reviewers set up their environment (e.g., `renv` or `conda`) to run this code?_
_For compute, can reviewers run this code on their laptop, or do they need additional computational resources such as RAM or storage?_
_Please make sure this information, if applicable, is documented in the README.md file._

#### Is there anything that you want to discuss further?

If you have any comments, ideas, very welcome. Either on the content of the template (as I said, I think some sections will not be so usefull) or the future plans:
- [ ] build the fetal kidney reference for Azimuth,
- [ ] include the label transfer from the fetal kidney reference to the clustering template.
- [ ] include inferCNV and/or copyKAT to the template.
- [ ] subdivide and re-cluster stroma and fetal nephron cells, taht contain nromal + cancer cells

#### Analysis module and review

- [x] This analysis module [uses the analysis template and has the expected directory structure](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/analysis-modules/).
- [x] The analysis module `README.md` has been updated to reflect code changes in this pull request.
[...]

Issue #703, opened 2024-08-08
Wilms-06: Download and build the Azimuth compatible fetl kidney reference


maud-p:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/635

### Describe the goals of the changes to the analysis module.

One important step in the module is the label transfer from the fetal idney reference (https://www.kidneycellatlas.org/).
Data are available from the website in a .h5ad format. 
The structure of the .h5ad file might not follow the standard structure because conversion to Seurat object using SeuratDisk functions didn’t work. 
This problem has been solved by the community using Python, R and reticulate (https://github.com/satijalab/seurat/issues/4711).

- [ ] I will thus adapt my Dockerfile to be able to use python and scanpy within my RMarkdown.

- [ ] I will add a 00-build_kidney_reference.Rmd to allow everyone building the azimuth compatible reference that will be used in the module.

### What will your pull request contain?

In the next PR, I would like to add a RMardown file and detail in the READ.ME file the steps to
1) download the fetal kidney atlas reference 
2) convert to Seurat object
3) create an Azimuth compatible reference that will be used all along the module for transfer ammotation.

The Dockerfile will be updated to allowthe use of python and scanpy. 

### Will you require additional software beyond what is already in the analysis module?

python

### If known, when do you expect to file the pull request?

09/08/2024

jaclyn-taroni:
Hi @maud-p, I wanted to provide some links to docs that you may find helpful.

We have some recommendations about Docker images that use R and Python: https://openscpca.readthedocs.io/en/latest/ensuring-repro/docker/docker-images/#images-with-both-r-and-conda-environments

We are specifically recommending that Python dependencies are managed with conda, and there's more information about using conda in OpenScPCA available in the docs as well: https://openscpca.readthedocs.io/en/latest/ensuring-repro/managing-software/using-conda/

Looking ahead a bit, I would expect that the Azimuth-compatible reference will [go into your S3 bucket as a results file](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/creating-pull-requests/before-file-pr/#add-results-files-to-s3).

As always, let us know if we can be helpful!

jaclyn-taroni:
You should just be able to leave that line as is 👍🏻

sjspielman:
This reference was prepared in #737.

PR #706, opened 2024-08-09
Adding 00-reference to build azimuth kidney reference

maud-p:

#703 

I write a RMarkdown script to 

1. download the data from the fetal kidney atlas
2. process it and create an azimuth compatible reference,

This will be used in the module to perform label transfer from the human fetal kidney atlas to the Wilms tumor samples.

I used a different approach than described in the issue #703 .
I figured out that I can download the human fetal kidney data from cellxgene as a rds object.
url = "https://datasets.cellxgene.cziscience.com/40ebb8e4-1a25-4a33-b8ff-02d1156e4e9b.rds"

Like this, I didn't need to create a conda/renv enrironment. 
The dockerfile from this PR is the same as the dockerfile from the PR #704 .

Of note however, using your documentation, I have been able to build a conda/renv container that actually allow to run the scripts I did so far, in case we need it in the future :)

Yes! The next one will be to implement the label transfer in the sample report. 

here is the command I used to upload the result to the bucket:

```shell
scripts/sync-results.py cell-type-wilms-tumor-06 \
    --bucket researcher-008971640512-us-east-2 \
```

The azimuth compatible reference in a format of 2 files:
- ref.Rds
- idx.annoy

When running the 00_fetal_reference_kidney.Rmd, these two files will be saved in the module folder/marker-genes. 

We build a reference that is compatible with the azimuth label transfer. 

I have one issue, using the RMarkdown 00_fetal_reference_kidney.Rmd, we can only build the reference manually running each chunk, but it do not work when we want to knittr as html report. 
I could isolate the problem to the AzimuthReference function, might be related to this issue: https://github.com/satijalab/azimuth/issues/219

1. Run the docker container, 
2. open RStudio from http://localhost:8787/
3. in the module folder open the 00_fetal_reference_kidney.Rmd and run chunk by chunk (do not knittr!!)

### Author checklists

_Check all those that apply._


analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+
+
+## Base directories
+
+```{r base paths, eval=TRUE, include=TRUE}
+# The base path for the OpenScPCA repository, found by its (hidden) .git directory
+repository_base <- rprojroot::find_root(rprojroot::is_git_root)
+
+# The current data directory, found within the repository base directory
+data_dir <- file.path(repository_base, "data", "2024-07-08", "SCPCP000006")
```
jaclyn-taroni, line 49: We always want to point to `data/current/SCPCP000006` so we can re-run this notebook without code changes.

```suggestion
data_dir <- file.path(repository_base, "data", "current", "SCPCP000006")
```

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+Note to the DataLab: should we save it somewhere else? 
+I suggest this rds file could be placed here transiently and removed once the reference is build?
+
+
+```{r path_to_data}
+path_to_data <- file.path(module_base, "marker-sets/fetal_full.rds")
+
+url = "https://datasets.cellxgene.cziscience.com/40ebb8e4-1a25-4a33-b8ff-02d1156e4e9b.rds"
+
+download.file(url, path_to_data)
```
jaclyn-taroni, line 74: I think we could move this URL to a parameter with this as the default. I'll comment above with that suggestion.

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+```
+## Output file
+
+The azimuth compatible fetal kidney reference will be saved in the marker-sets folder from the module.
+
+
+```{r path_to_output}
+path_to_output <- file.path(module_base, "marker-sets")
```
jaclyn-taroni, line 82: Given what I expect the size to be like, you probably want to put this in `results` and then sync to your bucket:

```suggestion
path_to_output <- file.path(module_base, "results", "references")
```

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+Please note that this download link permanently references the current version of the dataset (08/2024). 
+If this dataset is updated, a new download link will be created that permanently references the next version of this dataset.
+
+We save it in the marker-sets folder of the module.
+Note to the DataLab: should we save it somewhere else? 
+I suggest this rds file could be placed here transiently and removed once the reference is build?
+
+
+```{r path_to_data}
+path_to_data <- file.path(module_base, "marker-sets/fetal_full.rds")
```
jaclyn-taroni, line 70: For transient files, you can [put them in `scratch`](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/analysis-modules/#skeleton-analysis-module-contents).

```suggestion
We place it in the `scratch` folder of the module because it is a temporary file.


```{r path_to_data}
path_to_data <- file.path(module_base, "scratch", fetal_full.rds")
```

Files in `scratch` are ignored, which is what we want here!

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+
+
+```{r cfg}
+cfg = list()
+cfg$padj_thershold = 0.05
+cfg$lfc_threshold = 1
+cfg$rate1_threshold = 0.5
+
+
+set.seed(12345)
```
jaclyn-taroni, line 98: These can also be moved to `params` and then you could do the following:

```suggestion
cfg <- list()
cfg$padj_threshold <- params$padj_threshold
cfg$lfc_threshold <- params$lfc_threshold
cfg$rate1_threshold <- params$rate1_threshold


set.seed(params$seed)
```

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+
+# subset for plotting
+Idents(s) <- s$compartment
+cells <- WhichCells(s, downsample = 100)
+ss <- subset(s, cells = cells)
+ss <- ScaleData(ss, features = top5$feature)
+
+p1 <- SCpubr::do_DimPlot(s, reduction="umap", group.by = "compartment", label = TRUE, repel = TRUE) + ggtitle("compartment")
+p2 <- DoHeatmap(ss, features = top5$feature,  cells = cells, group.by = "compartment") + NoLegend() + 
+  scale_fill_gradientn(colors =  c("#01665e","#35978f",'darkslategray3', "#f7f7f7", "#fee391","#fec44f","#F9AD03"))
```
jaclyn-taroni, line 191: I think this is the same color palette as what is used below? So I'd recommend saving it as a variable so you only would need to change it one place:

```r
heatmap_color_palette <- c("#01665e","#35978f",'darkslategray3', "#f7f7f7", "#fee391","#fec44f","#F9AD03")
```

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+```
+
+
+## Input files
+
+
+rds data can be downloaded using the URLhttps://datasets.cellxgene.cziscience.com/40ebb8e4-1a25-4a33-b8ff-02d1156e4e9b.rds
+
+Please note that this download link permanently references the current version of the dataset (08/2024). 
+If this dataset is updated, a new download link will be created that permanently references the next version of this dataset.
```
jaclyn-taroni, line 62: You would need to change this text if using `params`.

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+title: "Build azimuth compatible fetal kidney reference from the kidney cell atlas"
+author: "Maud PLASCHKA"
+date: '2024-08-07'
+output: 
+  html_document: 
+    toc: yes
+    toc_float: yes
+    code_folding: hide
+    highlight: pygments
+    df_print: paged
```
jaclyn-taroni, line 11:
```suggestion
date: '2024-08-07'
params:
  ref_url: "https://datasets.cellxgene.cziscience.com/40ebb8e4-1a25-4a33-b8ff-02d1156e4e9b.rds"
  padj_threshold: 0.05
  lfc_threshold: 1
  rate1_threshold: 0.5
  seed: 12345
output: 
  html_document: 
    toc: yes
    toc_float: yes
    code_folding: hide
    highlight: pygments
    df_print: paged
```

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+```{r path_to_data}
+path_to_data <- file.path(module_base, "marker-sets/fetal_full.rds")
+
+url = "https://datasets.cellxgene.cziscience.com/40ebb8e4-1a25-4a33-b8ff-02d1156e4e9b.rds"
+
+download.file(url, path_to_data)
+```
+## Output file
+
+The azimuth compatible fetal kidney reference will be saved in the marker-sets folder from the module.
```
jaclyn-taroni, line 78: Update this text to reflect whatever happens in the chunk below!

analyses/cell-type-wilms-tumor-06/README.md
```diff
@@ -67,6 +67,25 @@ Some differenices are expected, some marker genes or pathways are associated wit
...
+
+REF: 
+[1] https://www.ncbi.nlm.nih.gov/books/NBK373356/
+[2] https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9915828/
+[3] https://www.science.org/doi/10.1126/science.aat5031
+
+The first step of the module is thus to download and build the azimuth compatible reference. 
+This can be achieved running 00_fetal_reference_kidney.Rmd. 
+Briefly, this will download the reference data from the cellxgene platform: https://datasets.cellxgene.cziscience.com/40ebb8e4-1a25-4a33-b8ff-02d1156e4e9b.rds
+and create an azimuth compatible Seurat object that will be saved in the marker-sets forlder as ref.Rds and idx.annoy files.
```
jaclyn-taroni, line 86: Update to reflect the use of `params` (to let folks know where to look to see what is being used!) and where it will be saved.

analyses/cell-type-wilms-tumor-06/results/README.md
```diff
@@ -3,3 +3,11 @@
...
 
 Instead, copy results files to an S3 bucket and add a link to the S3 location in this README file.
+
+
+# Azimuth compatible fetal kidney reference
+
+To perform label transfer using Azimuth and the fetal kidney atlas, a reference is build (in 00_fetal_reference_kidney.Rmd) using the fetal_full.Rds object download from :
+"https://datasets.cellxgene.cziscience.com/40ebb8e4-1a25-4a33-b8ff-02d1156e4e9b.rds"
+
+The result of 00_fetal_reference_kidney.Rmd can be found in the S3 bucket name researcher-maud-p-us-east-2 names ref.Rds and idx.annoy
```
jaclyn-taroni, line 13:
```suggestion
The result of 00_fetal_reference_kidney.Rmd can be found in the S3 bucket name `researcher-008971640512-us-east-2` named ref.Rds and idx.annoy
```

jaclyn-taroni (commented):
Hi @maud-p, thanks for filing this! It looks like there is some overlap with the files changed in #704, which makes sense – we'll just need to be careful about how we merge both pull requests in.

I am returning some initial feedback here that I'd put in two main categories:

- **Where to use `params`.** If we wanted a better understanding of how the parameters that are currently hardcoded in the notebook affected the results, it would be easier to test if we could pass new values (e.g., the seed being set) to `rmarkdown::render()` and compare the resulting notebooks. Using these also has the benefit of organizing values in one place, which can be very helpful!
  - This advice might be less impactful if we can't get the Azimuth issue sorted (I will need to take a closer look at that), but having things in one place is still helpful, in my opinion.
- **What paths to use.** I think `scratch` is what you will want to use for the transient file you download, and you'll want to use `results` instead of `marker-sets` for the result files.

Let me know if you have any questions 😄

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
jaclyn-taroni: I did check your results bucket, so I see the resulting files 👍🏻 

Did you need to move them to `results` manually? I think what I am proposing cuts out that step (if my assumption is correct, of course!)

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+Do not install packages here; only load them with the `library()` function.
+
+```{r packages, message=FALSE, warning=FALSE}
+library("Seurat")
+library(sctransform)
+library(Azimuth)
+library(SCpubr)
+library(tidyverse)
+library(patchwork)
+library(SeuratWrappers)
```
jaclyn-taroni, line 38: I was poking around locally to see if I could better understand the Azimuth problem, and I think we need to get `SeuratWrappers` into the `renv.lock` file.

I was able to install from RStudio within the Docker container with:

```
remotes::install_github("satijalab/seurat-wrappers@8d46d6c47c089e193fe5c02a8c23970715918aa9")
```

(This is the most recent commit.)

If you run `renv::snapshot()`, I expect some packages might get removed because the code in this branch doesn't account for what is getting added in #704. That would be okay, though; we'd need to resolve it in whichever branch gets merged second.

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
```diff
@@ -0,0 +1,240 @@
...
+
+options(future.globals.maxSize= 891289600000)
+s <- SCTransform(seurat, verbose = FALSE, method = "glmGamPoi", conserve.memory = TRUE)
+s <- RunPCA(s, npcs = 50, verbose = FALSE)
+s <- RunUMAP(s, dims = 1:50, verbose = FALSE, return.model = TRUE)
+```
+
+```{r create_ref, echo=TRUE, fig.height=7, fig.width=12, message=FALSE, warning=FALSE, out.width='100%'}
+options(future.globals.maxSize= 891289600000)
+Fetal_kidney <- AzimuthReference(
```
jaclyn-taroni, line 120: I am wondering if you have had any luck running `AzimuthReference` within a script? 🤔 Because if so, maybe we make this a script instead of a notebook.

jaclyn-taroni: I know very little about Azimuth, so forgive me if this is a naive question! How is the reference generated here different from what is available on Zenodo? https://zenodo.org/records/4738021#.YJIW4C2ZNQI

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
maud-p: yes you are right, makes sense! I'll correct this!

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
maud-p: not sure if I understand what this params is... Is that for later run of the all scripts in the modules in a .sh or .py execution file?

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
maud-p: Oh I see now :) thanks!

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
jaclyn-taroni: The `params` set in the YAML header (https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/706#discussion_r1711490706) can be referenced this way – `params` becomes a list.

analyses/cell-type-wilms-tumor-06/00_fetal_reference_kidney.Rmd
maud-p: I tried to run the same in a script instead of a RMarkdown but same, I couldn't run it using one clic on "Source". 🤔

jaclyn-taroni: I am new to this, but I am now under the impression we can use Seurat to accomplish many of the same things as Azimuth: https://azimuth.hubmapconsortium.org/#General

>#### Can I run the app myself?
> The source code is available [here](https://github.com/satijalab/azimuth). However, for users interested in performing these analyses outside the context of the Azimuth app, we suggest using Seurat v4 and using our vignette on [Mapping and annotating query datasets](https://satijalab.org/seurat/articles/integration_mapping.html) as an example. You can also download a Seurat v4 R script from the app once your analysis is complete to reproduce the results locally.

(h/t @jashapiro)

Following those links, I assume we'd want to use this section as a reference: https://satijalab.org/seurat/articles/integration_mapping.html#cell-type-classification-using-an-integrated-reference

Being unable to run this successfully except for chunk by chunk gives me pause – it would be hard to test this notebook in GitHub Actions.

maud-p: The reference you suggest can also be of interest but it contains 15 different organs fron a quick look. The one I wanted to use is only composed of cells from the kidney and from what I understood the annotation have been done by kidney experts.
But I could give a try with the one you suggest, might be more straightforward!
I'll compare the two on few samples.

jaclyn-taroni: > But I could give a try with the one you suggest, might be more straightforward! I'll compare the two on few samples.

I would be very curious about this result in general! 

Another option would be to try to use Seurat https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/706#discussion_r1712065012 with the kidney dataset.

I am trying to avoid the `AzimuthReference()` bug if at all possible 😄

maud-p: Thank you very much for looking into Seurat/Azimuth! 
The label transfer using FindTransferAnchors and TransferData (described in the link you suggested https://satijalab.org/seurat/articles/integration_mapping.html#cell-type-classification-using-an-integrated-reference) is what I used before Seurat and Azimuth v5! I'll go back to it!

maud-p:
Thank you very much @jaclyn-taroni for all your feedback and suggestions, all clear so far and make lot of sense!

I'd like to try:
- use Azimuth v5 for label transfer from the fetal kidney atlas 
- use seurat v4 for label transfer from the fetal kidney atlas
- try the azimuth fetal reference for label transfer
- compare the three on few samples. 

You are right, we might not need to build a reference in the end :)
I'll try to provide you feedback next Tuesday or Wednesday!

maud-p:
Hi @jaclyn-taroni !
Some feedback about the label transfer from two fetal (kidney) references using either `Seurat v4`or  `Azimuth`.  

# My approaches

I tested 3 different approaches:

-	Using the Stewart et al. reference, I tested the Seurat and Azimuth workflow. 
-	Using the Cao et al. azimuth reference, I used the Azimuth workflow. 
-
You can find one RMarkdown template (in notebook_template) per approach. The idea was to then select 1 or 2 to be rendered for all samples. 

I knitr the 3 templates for the sample 176 and you can find the html reports in the folder notebook/00-reference.

Of note: I started to wrote a R script `00_reference.R´ that would be used to render the notebook for all samples, following the advices from @sjspielman on the other PR. I am new to the `params`, I might have missed something because the R Script with `params` render html report without the plots. This does not happen when I render without `params`. Do you maybe have experience with this issue? 

# Result 1 : `Seurat`and `Azimuth`workflow gives similar results

First using the human fetal kidney atlas from Stewart et al. (https://www.kidneycellatlas.org/) = the reference I used first in the PR, I couldn’t build the reference using the `Azimuth` function ` AzimuthReference()`within a RMarkdown.
I thus tried to use the `Seurat`workflow (`FindTransferAnchors` --> ` TransferData`) that works smoothly (see one example of label transfer, on the right) and the results are highly comparable with the `Azimuth`approach (see one example of label transfer, on the left). 
 
image


# Result 2: The azimuth fetal reference (Cao et al) seems to be appropriate for the Wilms tumor dataset.

I also tried to use the azimuth fetal reference (Cao et al. 15 organs including kidney). I was first sceptic using this reference as it contained cells from 15 different organs. To my surprise, most of the cells label to kidney cells! The additional tissues in the reference dataset do not seems to be an issue 😊 
Here the `Azimuth` workflow works well. The label transfer appears to be quite similar to the one based on Stewart et al reference. 

# Conclusion: In a future PR, I’d like to render both label transfer (Stewart and Cao) on all samples.

jaclyn-taroni:
Thanks, @maud-p! It's very interesting that the "irrelevant" tissues are not a problem when using the entire fetal reference. I plan to take a closer look this afternoon (Eastern US).

maud-p:
I just wanted to let you know that I solved the plot rendering problem by using the `selfcontained` mode in the yaml parameters of the output: html_document :)
```r
output:
  html_document: 
    toc: yes
    toc_float: yes
    code_folding: hide
    highlight: pygments
    df_print: kable
    self_contained: yes
    mode: selfcontained 
```

jaclyn-taroni (commented):
Hi @maud-p,

With #704 and these new additions, I’m getting a fuller picture of where the module will end up. As a result, I’m going to propose some changes to the structure to help us reduce runtimes and ensure consistency across notebooks. 

I’ll summarize the high-level thoughts here:

* The download and preparation of the fetal kidney reference should go into its own script, so you don’t have to repeat that across notebooks or for individual samples. I’ve added a comment that contains what I think that script would look like and described how I would use it/how it fits together with the notebooks in other comments.
  * This would be the first thing you would run in `00_reference.R`
  * The second thing you’d run in `00_reference.R` would be a version of `00_fetal_reference_kidney.Rmd` that does the exploration and marker gene calculations.
* Create functions for common Seurat steps, such as converting from `SingleCellExperiment` and normalization, dimension reduction, etc. — then you can source these in whichever notebook you need them and be consistent in their application without repeating code.
  * In the future, you might create a script just for converting to Seurat objects. I would make that a *later pull request* if you think it’s a good idea; this one is fairly large and complex without those changes.
* Make sure you save any outputs of notebooks you expect to use in later steps, like the Seurat objects and results of transfer, in `results`!
* Remove data that you can get from the data release from the Git repository.
* I think there’s value in using `RunAzimuth()` consistently across the two atlases if you’re going to use both atlases. So, I would only run the `Seuratv4` notebook on a handful of samples to keep a record of the consistency between Seurat and Azimuth results.
* In terms of naming and organization:
  * I would rename the Cao and Stewart templates to be prefixed with `00a`, `00b`, etc., so it’s clear they are part of the reference stuff. I would also use the atlas name instead of the author’s name in the filenames.
  * I thought it might be helpful if each sample gets its own directory in `notebook/00-reference`.

General comment — if you’re going to be comparing the two atlases down the line, I think we should be considering how to do that quantitatively in addition to the qualitative plots you have here in future pull requests. All the more reason to be saving the outputs!

Once you make these changes, I will file a PR to this branch to set up running CI to test the module. I think that’s the best way to demonstrate how that should work, and then you could maintain that workflow in future PRs if you are interested in learning about it!

Thanks again for your contributions 😄 ! If you have any questions or run into issues, let us know! 

I was glad to see that the plotting issue was resolved because it was on my list to figure out before returning this review, so nice work on that 👍 🚀 I had not ever seen that before!

analyses/cell-type-wilms-tumor-06/notebook_template/01_fetal_reference_Stewart_Seuratv4.Rmd
```diff
@@ -0,0 +1,280 @@
...
+
+### Human fetal kidney atlas Stewart et al.
+
+We first wanted to try the human fetal kidney atlas to transfer label into the Wilms tumor samples using azimuth. 
+You can find more about the human kidney atlas here: https://www.kidneycellatlas.org/ [3]
+
+Using this reference, we test two workflows for label transfer:
+
+- `Azimuth v5`,
+- `Seurat v4` as described here  https://satijalab.org/seurat/articles/integration_mapping.html#cell-type-classification-using-an-integrated-reference
```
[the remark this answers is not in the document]
maud-p: hm sorry for the confusion, you are right I used `Seurat v5`, but the worflow was developed with lower version of `Seurat`. I should remove the version! (Or even not, as it seems that we will only go with `Azimuth` :) )
Sorry for the confusion!

analyses/cell-type-wilms-tumor-06/notebook_template/01_fetal_reference_Stewart_Azimuthv5.Rmd
```diff
@@ -0,0 +1,318 @@
...
+  k.param = 31,
+  plotref = "umap",
+  plot.metadata = NULL,
+  ori.index = NULL,
+  colormap = NULL,
+  assays = NULL,
+  metadata = c("compartment", "cell_type"),
+  reference.version = "0.0.0",
+  verbose = FALSE
+)
```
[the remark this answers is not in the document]
maud-p: Thank you so much !

analyses/cell-type-wilms-tumor-06/notebook_template/01_fetal_reference_Cao_Azimuth.Rmd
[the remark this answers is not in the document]
maud-p: While renaming I just though, would it be OK to go with 
- "00a_fetal_all_reference_Cao.Rmd"
- "00b_fetal_kidney_reference_Stewart.Rmd"?

Just to make sure it is clear it is not from the same reference, i.e. fetal_kidney is not a subset of fetal_all?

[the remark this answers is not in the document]
jaclyn-taroni: Yes, sounds good 👍🏻

maud-p:
Thank you very much @jaclyn-taroni !
I am working on the few remaining changes and will let you know once it is done!

Thank you so much, your reviews and the ones from @sjspielman are super clear and useful!

maud-p:
Hi @jaclyn-taroni ,

Thank you for your review, I tried to apply the changes :)
I reorganized a bit (hope for something better/clearer).

The `00_run_workflow.R` is the starting point. It calls the script to build the reference `download-and-create-fetal-kidney-ref.R` and the notebok to characterize it `00b_characterize_fetal_kidney_reference_Stewart.Rmd`. Results are saved in `results/reference`.

The the loop over the samples perform in three steps:
- `Seurat workflow`, nornalization and clustering: `01_seurat-processing.Rmd` in `notebook_template`
- `Azimuth` label transfer from the fetal full reference (Cao et al.) in `notebook_template`
- `Azimuth` label transfer from the fetal kidney reference (Stewart et al.) in `notebook_template`

For the input and output, I went for the following strategy:
We start with the `_process.Rds` data to run `01_seurat-processing.Rmd`. 
The output of `01_seurat-processing.Rmd` is saved in `results` in a subfolder for each sample and is the input of the second step `02a_label-transfer_fetal_full_reference_Cao.Rmd`.
The output of `02a_label-transfer_fetal_full_reference_Cao.Rmd` is then the input of `02b_label-transfer_fetal_kidney_reference_Stewart.Rmd`.

At the end of the workflow, we have a `Seurat`object that contains:
- normalization and clustering, dimensional reductions
- label transfer from the fetal full reference
- label transfer from the fetal kidney reference

I tried to run it on 11 samples, seems to work :-)

Results should be loaded on the S3 bucket `researcher-008971640512-us-east-2`


Please let me know if something is unclear!

Thank you!

analyses/cell-type-wilms-tumor-06/notebook_template/02b_label-transfer_fetal_kidney_reference_Stewart.Rmd
```diff
@@ -0,0 +1,204 @@
...
+
+
+### Azimuth annotation from fetal kidney
+
+```{r run_azimuth, message=FALSE, warnings=FALSE}
+
+data <-  RunAzimuth(srat, path_to_reference, assay = 'RNA')
+# We transfer the annotation to the pre-processed `Seurat`object as we don't want to keep changes done on the query by `RunAzimuth`
+metadata_to_trasfer <- data@meta.data[, c("predicted.compartment.score", "predicted.compartment", "predicted.cell_type.score", "predicted.cell_type")]
+srat <- AddMetaData(srat, metadata_to_trasfer, col.name = c("fetal_kidney_predicted.compartment.score", "fetal_kidney_predicted.compartment", "fetal_kidney_predicted.cell_type.score", "fetal_kidney_predicted.cell_type"))
```
jaclyn-taroni, line 141: Similar to my comment on the other notebook:
```suggestion
# We transfer the annotation to the pre-processed `Seurat`object as we don't want to keep changes done on the query by `RunAzimuth`
metadata_vec <- c("predicted.compartment.score", "predicted.compartment", "predicted.cell_type.score", "predicted.cell_type")
metadata_to_trasfer <- data@meta.data[, metadata_vec]
srat <- AddMetaData(srat, metadata_to_trasfer, col.name = paste0("fetal_kidney_", metadata_vec))
```

analyses/cell-type-wilms-tumor-06/notebook_template/02b_label-transfer_fetal_kidney_reference_Stewart.Rmd
```diff
@@ -0,0 +1,204 @@
...
+
+- the fetal nephron / mesenchymal cells are cap-mesenchyme cells. 
+In our case, cap-mesenchyme contains blastema and primitive epitheliul cancer cells. 
+
+- the stroma / mesenchymal stem are likelly mesenchymal cancer or normal cells. 
+
+## Save the `Seurat`object
+
+```{r save}
+saveRDS(object = srat, file = file.path(output_dir, paste0("00b-fetal_kidey_label-transfer_",params$sample_id,".Rds")))
```
jaclyn-taroni, line 187:
```suggestion
saveRDS(object = srat, file = file.path(output_dir, paste0("00b-fetal_kidney_label-transfer_",params$sample_id,".Rds")))
```

analyses/cell-type-wilms-tumor-06/results/README.md
```diff
@@ -3,3 +3,15 @@
...
 
 Instead, copy results files to an S3 bucket and add a link to the S3 location in this README file.
+
+
+# Azimuth compatible fetal kidney reference
+
+To perform label transfer using Azimuth and the fetal kidney atlas, a reference is build (in 00_fetal_reference_kidney.Rmd) using the fetal_full.Rds object download from :
+"https://datasets.cellxgene.cziscience.com/40ebb8e4-1a25-4a33-b8ff-02d1156e4e9b.rds"
+
+The result of 00_fetal_reference_kidney.Rmd can be found in the S3 bucket name `researcher-008971640512-us-east-2` named ref.Rds and idx.anno
```
jaclyn-taroni, line 13:
```suggestion
The result of 00_fetal_reference_kidney.Rmd can be found in the S3 bucket name `researcher-008971640512-us-east-2` named ref.Rds and idx.annoy
```

maud-p:
Hi @jaclyn-taroni , Thank you so much! I will test this now! 
Thank you!

analyses/cell-type-wilms-tumor-06/notebook_template/02a_label-transfer_fetal_full_reference_Cao.Rmd
```diff
@@ -0,0 +1,208 @@
...
+
+### Label transfer from fetal kidney reference using Azimuth
+
+```{r run_azimuth, message=FALSE, warnings=FALSE}
+DefaultAssay(srat) <- "RNA"
+s <- RunAzimuth(srat, reference ="fetusref" )
+
+# We transfer the annotation to the pre-processed `Seurat`iobject as we don't want to keep changes done on the query by `RunAzimuth`
+metadata_to_trasfer <- s@meta.data[, c("predicted.annotation.l1.score", "predicted.annotation.l1", "predicted.annotation.l2.score", "predicted.annotation.l2", "predicted.organ.score", "predicted.organ")]
+srat <- AddMetaData(srat, metadata_to_trasfer, col.name = c("fetal_full_predicted.annotation.l1.score", "fetal_full_predicted.annotation.l1", "fetal_full_predicted.annotation.l2.score", "fetal_full_predicted.annotation.l2", "fetal_full_predicted.organ.score", "fetal_full_predicted.organ"))
```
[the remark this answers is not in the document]
maud-p: it works, thanks!

maud-p:
Hi @jaclyn-taroni ,

Looking at the `Action` workflow, it seems that the `Renv Set-up` is not working because of  ‘Rhtslib’.
Googling a bit on it, it might be an issue of `linux/x86_84`. https://github.com/Bioconductor/Rhtslib/issues/23

Is that related to your note on Apple Silicon?
`If you are on a Mac with an M series chip, you will not be able to use RStudio Server if you are using a linux/amd64 or linux/x86_84 (like the ones available from ECR). You must build an ARM image locally to be able to use RStudio Server within the container.`

I am wondering because the image has been successfully build outside of this workflow. 

Thank you :)

jaclyn-taroni:
Hi @maud-p, I believe the problem is system dependencies on the Ubuntu machine running the module test. 

The image-building workflow is passing (https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/10452889901); the use of `renv` outside the Docker container is the problem. 

We need to install certain system dependencies. I was testing this out in a branch here, and this step seems to fix it: https://github.com/jaclyn-taroni/OpenScPCA-analysis/blob/38c0de66f4248516a6245b718cc1818f315eff44/.github/workflows/run_cell-type-wilms-tumor-06.yml#L46-L48

But then there's another failure that I haven't figured out yet: https://github.com/jaclyn-taroni/OpenScPCA-analysis/actions/runs/10461410602/job/28969673082. I _suspect_ we might need to move the script building the reference outside of `00_run_workflow.R`.

I am out of the office with the rest of the Data Lab team this week (#709), so I apologize in advance for the slow iteration!

maud-p:
FYI, I ran the workflow (for loop for `Seurat` and label transfer) on my machine. 

It took around 4h to complete the analysis for the 40 samples. 

Results with the R objects per samples are on S3 bucket `008971640512`

🎉

jaclyn-taroni:
Hi @maud-p - I've been continuing to work through the CI errors in a branch. You can see the latest GitHub Actions run here: https://github.com/jaclyn-taroni/OpenScPCA-analysis/actions/runs/10578864325/job/29309947677

It looks like we need to make sure `ggplotify` is installed using `renv`. It is not currently in the `renv.lock` file. I might suggest adding the following to `analyses/cell-type-wilms-tumor-06/components/dependencies.R`:

```R
library(ggplotify)
```

And then running `renv::snapshot()` and committing the results.

maud-p:
Hi @jaclyn-taroni , Thank you very much, I'll do it in a minute :)

Did you solde the `AzimuthReference` issue or you just ignore this step to proceed further? 
I tried to change/trick a bit the `AzimuthReference` function but without success. Not sure if I understood how to use this advice https://github.com/satijalab/azimuth/issues/155

Thank you!

jaclyn-taroni:
The way I solved this was to add it to its own step: https://github.com/jaclyn-taroni/OpenScPCA-analysis/blob/cbb4f6165b598835bf10a542190b072c955b4012/.github/workflows/run_cell-type-wilms-tumor-06.yml#L59-L65

And for now, commented out the `source()` step: https://github.com/jaclyn-taroni/OpenScPCA-analysis/blob/cbb4f6165b598835bf10a542190b072c955b4012/analyses/cell-type-wilms-tumor-06/00_run_workflow.R#L14-L15

This is temporary, though. I want to make sure we can resolve all the issues (e.g., missing dependencies). The solution might be to run `download-and-create-fetal-kidney-ref.R` from a shell script instead of an Rscript.

jaclyn-taroni:
Hi @maud-p - I wanted to let you know I am looking into the latest failure. We have used a strategy where we use `params` to change behavior based on whether or not a notebook is running in CI previously, and I am trying that out (can see in this branch: https://github.com/jaclyn-taroni/OpenScPCA-analysis/tree/jaclyn-taroni/install-rhtslib/analyses/cell-type-wilms-tumor-06). I think finding an appropriate value for `k.weight` in CI might be tricky, but I'll keep at it!

maud-p:
I understand now, thank you very much @jaclyn-taroni !! I didn't realized that in CI the samples were downsampled. 

But then I am a bit afraid, it seems that decreasing `k.weight` might not work https://github.com/satijalab/azimuth/issues/197

I'll also try to play with this and keep you updated.

maud-p:
@jaclyn-taroni I am on it, I might have found a solution modifiying a bit the `RunAzimuth` function.

Another parameter to reduce is `mapping.score.k` to 80 (instead of 100).

But then comes another ERROR a bit later because of a bug in the `RunAzimuth` function I guess.
The problem is that during `FindTransferAnchors`, `mapping.score.k` is reinitialized to `ksmooth` (default = 100).

As `ksmooth` is not a parameter of `RunAzimuth`, I copy/pasted the `source` of the function nd added it :) Seems to work on my RSession, I'll commit the changes to github in about an hour hopefully :)

maud-p:
FYI @jaclyn-taroni , I tryied first to copy your strategy to use params to change behavior based on whether or not a notebook is running in CI previously, and I am trying that out (https://github.com/jaclyn-taroni/OpenScPCA-analysis/tree/jaclyn-taroni/install-rhtslib/analyses/cell-type-wilms-tumor-06) but it failed. I might have messed up something... 

I thus decided to adapt the behavior depending on the size of the sample. Not as elegant but temporary :) 
What is the CI strategy for downsampling? How many cells are left?  I think my first threshold with `n= 100 cells` might have been too low, I hope increasing it will help :)

jaclyn-taroni:
I will take a look! Here's the docs page about test data: https://openscpca.readthedocs.io/en/latest/getting-started/accessing-resources/getting-access-to-data/#accessing-test-data. But you are probably looking for the module instead: https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/simulate-sce We use the default 100 cells to simulate. The problem could be the nature of the simulated data itself, too.

maud-p:
Thanks, I have the impression the workflow do not go to the `RunAzimuth_small` as expected, `if(dim(srat)[2] < 500)` might be the wrong command line. 
I try now with the only option to use `RunAzimuth_small`, let's see :)
Then I do not have much more idea :/

jaclyn-taroni:
Okay, @maud-p - let's do this:

- Can you roll back this branch to what you had in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/706/commits/f98da21d6f936791da10ca2d903e9661d9f409d0? (i.e., before all the k.weight debugging, etc.)
- Also use `system()` instead of `source()` for `download-and-create-fetal-kidney-ref.R` in `00_run_workflow.R`

Then I will give this a final review for correctness, etc. and merge it without CI. We'll figure out CI in a follow-up PR.

maud-p:
@jaclyn-taroni I created a new PR https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/737 with a new branch at the stage of `[f98da21](https://github.com/AlexsLemonade/OpenScPCA-analysis/commit/f98da21d6f936791da10ca2d903e9661d9f409d0)` using `system` instead of `source` 

Sorry I wasn't sure how to rollback to this state. Also sorry that I couldn't find a way to use this `RunAzimuth`. Is there a way that the downsample on CI results in >200 cells objects?

Thank you very much for your help!!

One question, after you merge it, should I rather try to solve this issue or continue the analysis? I guess these could be two independant PR?

Thank you again!!

jaclyn-taroni:
> @jaclyn-taroni I created a new PR https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/737 with a new branch at the stage of `[f98da21](https://github.com/AlexsLemonade/OpenScPCA-analysis/commit/f98da21d6f936791da10ca2d903e9661d9f409d0)` using `system` instead of `source` 
> 
> Sorry I wasn't sure how to rollback to this state.

I think you picked the easiest way to do it, so that totally works :+1:

> Also sorry that I couldn't find a way to use this `RunAzimuth`.

No worries – sorry it has been such a pain! 😅 

> Is there a way that the downsample on CI results in >200 cells objects?
> 
> One question, after you merge it, should I rather try to solve this issue or continue the analysis? I guess these could be two independant PR?

I'd recommend moving on to clustering, and we will try to figure out the best thing to do internally at the Data Lab. If increasing the number of cells is what is necessary, we'd have to figure out the best approach.

jaclyn-taroni:
Closing in favor of #737

PR #707, opened 2024-08-12
Update periodic release issue template to minimize steps

jaclyn-taroni:
We would like to limit the work involved in a dated release to just making sure that the most recent run of all modules ran correctly rather than triggering it manually and also running spellcheck.

I left the part about any blocking issues, as I could imagine that would come up from time to time.


jashapiro (approved):
LGTM

Issue #717, opened 2024-08-14
Allow increased space for docker build/push


jashapiro:
We are occasionally running into space issues when building Docker images, or specifically when pushing them to repositories. See https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/10393802804 for the latest example.

One way to solve this is to make sure Docker images don't get too large, but that may not be possible for all cases. 

The other option is to use larger runners for building. We might be able to do this by updating only the https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/build-push-docker-module.yml file; it should not require too much more space, as the image was able to build at least. It is only in the packaging for push that it seems things have gone wrong. 

Another possibility may be to see if we can add any cleanup steps to free space after the initial build, but I am not sure what the best way to do that would be at this time.

PR #718, opened 2024-08-14
revert cell-type-ewings conda-lock

jashapiro:
The changes in #714 for the `cell-type-ewings` module resulted in a docker image that was too big to build and push on our current GHA setup (see #717). 

In the interests of not having a docker image that doesn't match the `conda-lock` file, I am reverting the conda-lock changes in that module. I retained one change, which was to remove the `defaults` channel from the `conda-lock.yml` file manually. No packages were using this channel due to its low priority, so removing it from the list did require altering any installed packages. This change should allow future updates to proceed without errors (unless the channel list changes again).

### DIFF
--- a/analyses/cell-type-ewings/conda-lock.yml
+++ b/analyses/cell-type-ewings/conda-lock.yml
@@ -13,13 +13,15 @@
 version: 1
 metadata:
   content_hash:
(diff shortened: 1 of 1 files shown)


sjspielman (approved):
meant to approve!

PR #720, opened 2024-08-15
Don't have spellcheck action fail

jashapiro:
(Posting this after I got a workflow failure email that it took me a bit too long to figure out was not really because of something I had done.)

Since the spellcheck action already posts any spelling errors to an issue, we probably don't also need the action to fail as well. It seems like the workflow failure notice maybe goes to the last person to commit to main (did anybody else get it for the most recent failure?), which seems pretty arbitrary.

### DIFF
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -67,8 +67,8 @@ jobs:
             OpenScPCA admin
             spelling
 
-      - name: Fail if there are spelling errors
-        if: steps.spell.outputs.error_count > 0
+      - name: Fail if there are spelling errors for non-scheduled jobs
+        if: github.event_name != 'schedule' && steps.spell.outputs.error_count > 0
         run: |
           echo "There were ${{ steps.spell.outputs.error_count }} errors"
           column -t spell_check_errors.tsv
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -66,10 +66,3 @@ jobs:
           labels: |
             OpenScPCA admin
             spelling
-
-      - name: Fail if there are spelling errors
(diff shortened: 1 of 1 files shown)


jaclyn-taroni:
I would expect this to include logic that prevents failure only on the scheduled job. I think there's value to failing on relevant pull requests.

jashapiro:
> I would expect this to include logic that prevents failure only on the scheduled job. I think there's value to failing on relevant pull requests.

Of course you are correct. Updated to just add an exclusion of that step for scheduled jobs.

.github/workflows/spellcheck.yml
```diff
@@ -67,8 +67,8 @@ jobs:
             OpenScPCA admin
             spelling
 
-      - name: Fail if there are spelling errors
-        if: steps.spell.outputs.error_count > 0
+      - name: Fail if there are spelling errors for non-scheduled jobs
+        if: github.event_name != 'schedule' && steps.spell.outputs.error_count > 0
```
jaclyn-taroni, line 71: Does this need to be:

```suggestion
        if: ${{ github.event_name != 'schedule' && steps.spell.outputs.error_count > 0 }}
```

Going off line 60 here.

jaclyn-taroni (approved):
I agree with the logic, but the syntax might be incorrect. See my one comment. I don't need to see this again.

.github/workflows/spellcheck.yml
jashapiro: The syntax for `if:` statements is special in that GHA assumes the content is an expression, so you don't need the wrapper. See the first note in: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions

PR #728, opened 2024-08-27
Update code styling workflow

### LINKED ISSUE #708: Style code on a per module basis
The [code styling workflow](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/.github/workflows/code-styling.yml) is pretty broad – it will style any relevant file in `analyses`. If we use it as currently written, we may cause conflicts in analysis modules being actively developed.

We could rework this to only style analysis modules that are deemed sufficiently mature, similar to how we list modules in the Docker all modules workflow: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/4d2cbd550db9a5090d51a205ac7dcaddbb37c5ae/.github/workflows/docker_all-modules.yml#L35-L41

sjspielman:
Closes #708 

This (draft!) PR updates the code styling workflow to run on only mature modules. 

- Currently in my testing, `ruff` seems to want to look at all code under `analyses` even with the provided path. I've tried all kinds of quotes combinations... 
- The action is not turned on for testing _here_ yet; will do when we think ^ is fixed.
- For now I commented out ewings - is it mature enough to include yet?
- Let me know what we think about the number of branches/PRs. Right now I'm sending each module's code styling to its own branch/PR, because I'm concerned about conflicts from concurrent matrix jobs. I don't want to change our setting of `delete-branch: true` in the PR step, because we would want to make sure any _stale_ branch is definitely deleted. But at the same time, this setting might risk the branch "deleting itself" via different matrix job. Given this, I thought one branch per module was the safer bet?


jashapiro (commented):
I think my line comment should solve the issue with picking a directory not working. In essesence I think you are ending up with a command that looks like `ruff format <module-dir> .` which is then doing both the module directory and the whole repo. I think the fix is just to use the `with: src:` arg for the action to replace the `'.'` rather than add to it.

I don't know how exactly I feel about the matrix jobs with multiple PRs. It probably won't be too bad, and it does allow for more focused reviews. But I wonder if we might want to take an alternative approach where we have a file we maintain with a list of the "mature" modules that we could read in an early step, parse, and pass along as an env variable. Then we could have a single set of jobs and a single PR per run if we wanted that.

In practice, that approach might not exclude using a matrix, but we might have to get a bit fancy passing the list of modules from a file to the matrix definition. Pretty sure it is possible though with multiple jobs where only the second job uses a matrix strategy and then using `matrix: include:` to pass the list of modules to the second job. Syntax might be a bit tricky, but I think it is possible.

A note for implementation of the single-PR version is that I also think you can pass a list of directories to the `src:` argument in `ruff-action` as that value gets [split internally](https://github.com/ChartBoost/ruff-action/blob/491342200cdd1cf4d5132a30ddc546b3b5bc531b/action/main.py#L26).

.github/workflows/code-styling.yml
```diff
@@ -45,25 +57,25 @@ jobs:
...
 
-
       ## Second, style the python code ##
-      - name: Format python files
+      - name: Style Python code for each analysis module
         uses: chartboost/ruff-action@v1
         with:
-          args: format analyses/
+          # TODO: This is grabbing everything under analyses, with and without quotes
+          args: format analyses/${{ matrix.module }}/
```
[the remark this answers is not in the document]
sjspielman: Ah! I remember seeing this but didn't try it, since I went to look for more docs on whether `src:` would do what I thought...and then promptly didn't follow up...

sjspielman:
> In practice, that approach might not exclude using a matrix, but we might have to get a bit fancy passing the list of modules from a file to the matrix definition. Pretty sure it is possible though with multiple jobs where only the second job uses a matrix strategy and then using matrix: include: to pass the list of modules to the second job. Syntax might be a bit tricky, but I think it is possible.

I think this sort of situation is about what you're describing, no? https://github.com/AlexsLemonade/AdminItems/blob/d3feaff26e9b803195b23ee25e2d5c617f3e5f10/.github/workflows/file-annual-creds-rotations.yml#L55-L81

sjspielman:
Just a note that `src:` was good for getting `ruff` to run in the right directories.

jashapiro:
> > In practice, that approach might not exclude using a matrix, but we might have to get a bit fancy passing the list of modules from a file to the matrix definition. Pretty sure it is possible though with multiple jobs where only the second job uses a matrix strategy and then using matrix: include: to pass the list of modules to the second job. Syntax might be a bit tricky, but I think it is possible.
> 
> I think this sort of situation is about what you're describing, no? https://github.com/AlexsLemonade/AdminItems/blob/d3feaff26e9b803195b23ee25e2d5c617f3e5f10/.github/workflows/file-annual-creds-rotations.yml#L55-L81

Yes, though I would probably just store the list as JSON to avoid the ugly awk

sjspielman:
Well this is interesting... https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/10618217531/job/29432900775#step:7:61

jashapiro:
> Well this is interesting... https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/10618217531/job/29432900775#step:7:61

80% sure this is because it is running in a fork.

.github/workflows/code-styling.yml
```diff
@@ -15,9 +19,30 @@ concurrency:
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Read in list of mature analysis modules
+        id: read-modules
+        run: |
+          # From input `mature-modules.json`, read in modules which should be styled and save to $GITHUB_OUTPUT
+          echo "matrix=$(jq -cr . < ./.github/workflows/mature-modules.json)" > $GITHUB_OUTPUT
```
jashapiro, line 37: Adding an additional comment, removing unneeded (I think) `./` and `.`. Also always `>>` for `GITHUB_OUTPUT`
```suggestion
        run: |
          # From input `mature-modules.json`, read in modules which should be styled, reformat, and save to $GITHUB_OUTPUT
          echo "matrix=$(jq -cr < .github/workflows/mature-modules.json)" >> $GITHUB_OUTPUT
```

.github/workflows/code-styling.yml
```diff
@@ -15,9 +19,30 @@ concurrency:
...
+
   style-code:
     runs-on: ubuntu-latest
     name: Perform code styling
+    needs: setup-module-matrix
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.setup-module-matrix.outputs.matrix) }}
```
jashapiro, line 45: I think for clarity it is probably worth making this a bit more explicit
```suggestion
      matrix: 
        module: ${{ fromJSON(needs.setup-module-matrix.outputs.matrix.modules) }}
```

Though see my overall comment about simplifying a bit so you can maybe just use `needs.setup-module-matrix.outputs.modules`

.github/workflows/code-styling.yml
```diff
@@ -45,25 +73,25 @@ jobs:
...
-          branch: style-analysis-modules
+          branch: style-module_${{ matrix.module }}
           base: main
           delete-branch: true
-          title: "GHA: Automated analysis module code styling"
+          title: "GHA: Automated code styling for analysis module: ${{ matrix.module }}"
           body: |
             ### Description:
-            This PR was auto-generated by GitHub Actions to style all R (scripts and notebooks) and Python (scripts only) code in `analyses`.
+            This PR was auto-generated by GitHub Actions to style all R (scripts and notebooks) and Python (scripts only) code in `analyses/${{ matrix.module }}`.
```
jashapiro, line 94: ruff does notebooks too as of version 0.6.0, and I am pretty sure this action gets the latest version by default. 
```suggestion
            This PR was auto-generated by GitHub Actions to style all R and Python code (in scripts and notebooks) in `analyses/${{ matrix.module }}`.
```

jashapiro (commented):
I might actually simplify this a bit by making the json file simply look like this: 

```
[
    "hello-R",
    "hello-python",
    "simulate-sce",
    "doublet-detection"
]
```
 and then save not to `matrix` but directly to a `modules` output.

Or maybe even simpler for maintenance, use `mature-modules.txt`: 
```
hello-R
hello-python
simulate-sce
doublet-detection
```

then:

```
echo "modules=$(jq -nRc '[inputs]|map(select(length>0))' mature-modules.txt) >> $GITHUB_OUTPUT
```

(the map/select removes any blank lines, including trailing newline).

I'm also not sure where we should put the mature modules file; where you have it is probably fine for now, but I am not sure it should be buried in a hidden directory.

sjspielman:
> I'm also not sure where we should put the mature modules file; where you have it is probably fine for now, but I am not sure it should be buried in a hidden directory.

I too am not sure about this, and similarly I'm not sure about calling this "mature modules," either, because modules are at different levels of maturity. This list is specifically "what's ready to be styled," so I now think we want a file name that more specifically reflects that too.

.github/workflows/code-styling.yml
sjspielman: yup, default: https://github.com/ChartBoost/ruff-action?tab=readme-ov-file#advanced-usage

sjspielman:
This should be ready for another look! Let me know what you think about the mature modules list, which I've renamed for now (also do we want ewings in there yet?).

Note that this passed checks here (no PRs filed, but matrix jobs were run): https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/728/commits/0e51c5a991a0f709d7cedd93377106d2ae2b2921

.github/workflows/code-styling.yml
```diff
@@ -1,3 +1,7 @@
+# GitHub Action to perform code styling
+# This action reads in a list of modules that should be styled from the file `mature-modules.json`.
+# A PR is opened for each module whose code was changed by the stylers.
+
 name: Code styling
 on:
   #pull_request:
```
jashapiro, line 7: I would remove this trigger. Since this action files separate PRs, we are never going to want to run it on submitted PRs without substantial changes.

.github/workflows/code-styling.yml
```diff
@@ -15,9 +19,31 @@ concurrency:
...
+    runs-on: ubuntu-latest
+    name: Setup the module matrix
+    outputs:
+      module: ${{ steps.read-modules.outputs.module }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Read in list of mature analysis modules
```
jashapiro, line 33:
```suggestion
      - name: Read in list of analysis modules to style
```

jashapiro (approved):
LGTM, with just a couple of minor suggestions.

PR #735, opened 2024-08-29
Initiate ropenspca package

### LINKED ISSUE #732: Initiate rOpenScPCA package
This issue tracks initiating the package `rOpenScPCA` as discussed in #731. It should live in `/packages/rOpenScPCA`.
As part of this, a `/packages/README.md` file will also be added.

sjspielman:
Closes #732 

This PR initiates the directory for the `rOpenScPCA` package. Some notes:

- I've also added a workflow to `R CMD CHECK` the package, and a _temporary_ function to test as part of that to make sure it works.
- I was on the fence for whether or not we should use `renv` to track packages used during package development. For now, I've included `renv`, but I'm not really tied to keeping it. What do you think?
- In `DESCRIPTION` how's my syntax for referencing the repo's top-level LICENSE file? Surprising nobody, `devtools::check()` gives it a `WARNING`, but could be ok for this circumstance?


sjspielman:
If `rcmdcheck` is loading up the `renv` environment, that would explain the CI failure as indeed `rcmdcheck` isn't in `renv.lock`...

jashapiro:
I think I would start this without `renv` as that adds some unnecessary complexity.

.github/workflows/test_ropenscpca.yml
```diff
@@ -0,0 +1,39 @@
...
+          r-version: 4.4.0
+          use-public-rspm: true
+
+      - name: Remove `.Rprofile` to not activate renv
+        run: |
+          rm ./packages/rOpenScPCA/.Rprofile
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: |
```
jashapiro, line 30: I think you need to specify the directory here. (And you should not need to specify `testthat` as that is part of the DESCRIPTION)
```suggestion
        with:
          working-directory: "packages/rOpenScPCA/"
          extra-packages: |
```

sjspielman:
This build is [now succeeding for tests](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/10634131342/job/29480772657), so I went ahead and removed my temporary function+test which were implemented to make sure GHA was working as expected. 

Of course, `R CMD CHECK` also failing but in an entirely expected way. It doesn't like my LICENSE syntax, which is rather fair... What license syntax should we go for here?

packages/rOpenScPCA/DESCRIPTION
```diff
@@ -0,0 +1,25 @@
...
+           role = c("aut"))
+    )
+Maintainer: Stephanie J. Spielman <stephanie.spielman@ccdatalab.org>
+Description: This package contains utility functions that support single-cell RNA-seq
+    analysis in R-based OpenScPCA analysis module code.
+License: BSD_3_clause + ../../LICENSE file
+Encoding: UTF-8
+LazyData: true
+Suggests:
+    devtools,
```
jashapiro, line 22: `devtools` is only used for building, so it should not be needed in the package spec
```suggestion
```

jashapiro (commented):
Looks good overall, but we do want it to pass all the checks.

packages/rOpenScPCA/DESCRIPTION
```diff
@@ -0,0 +1,25 @@
...
+           role = c("aut", "cre")),
+    person(c("Joshua", "A."), "Shapiro",
+           email = "josh.shapiro@ccdatalab.org",
+           comment = list(ORCID = "0000-0002-6224-0347"),
+           role = c("aut"))
+    )
+Maintainer: Stephanie J. Spielman <stephanie.spielman@ccdatalab.org>
+Description: This package contains utility functions that support single-cell RNA-seq
+    analysis in R-based OpenScPCA analysis module code.
+License: BSD_3_clause + ../../LICENSE file
```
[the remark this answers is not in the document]
sjspielman: > To quiet this error, I would probably just copy the license file into the package.

👍

jashapiro (approved):
LGTM

PR #736, opened 2024-08-29
Fix linking to use relative links

### LINKED ISSUE #733: Improve handling of symlinks within Docker containers
WE have had reports that symlinks do not work when used within mounted containers. In particular, this affects the symlink to the `current` data directory that we create/update when downloading data.

We may need to make these links absolute, which can be done with a small modification of the download scripts to use different links.  The original thought was that we were creating relative links and might need absolute, but looking at my recent download, it seems we are perhaps creating absolute links, which is not what I had intended!

I will make the change I expect we need, then test that it works within a docker image.

jashapiro:
Closes #733

It seems that the problem in #733 was not relative path symlinks being a problem in Docker, but in fact that I had introduced a bug whereby we were creating absolute path symlinks, which I would fully expect to fail when mounted in Docker images. 

This should fix the issue, and does in my local testing. Because this is creating the behavior we expected, I do not think we should need any docs updates, but please do confirm that you agree, and that this fixes the issue.

You should be able to run `./download-data.py --update-symlink` to fix any issues with your current `data/` directory.

### DIFF
--- a/download-data.py
+++ b/download-data.py
@@ -172,7 +172,7 @@ def update_symlink(data_dir: pathlib.Path, target: str, dryrun: bool = False) ->
         else:
             current_symlink = data_dir / "current"
             current_symlink.unlink(missing_ok=True)
-            current_symlink.symlink_to(target_path)
+            current_symlink.symlink_to(target) # the relative path
             print(f"Updated 'current' symlink to point to '{target}'.")
 
 


jaclyn-taroni (approved):
:+1: this fixed the issue I observed while developing in Docker

PR #740, opened 2024-08-30
Update docker images and GHA to use miniforge

jashapiro:
Closes #711 
Closes #713

In this PR, I changed the GitHub Actions for active modules and templates to use miniforge in place of miniconda.

I also updated the Docker images that use both conda and renv to be set up using miniforge (now mirroring the installation from the miniforge Docker image). I also updated the docs example for that instruction, but I did not do a larger docs update, as I will leave that for #712.

While I was in there, I cleaned a few little things in the Dockerfiles to make them more uniform.

PR #744, opened 2024-09-03
Test fix for docker actions failing after miniforge

jashapiro:
Looking at the changes made in #740, I think the most likely culprit for the failures in #743 is that I had removed adding conda.sh to `/etc/profile.d`. It seems possible (likely) that GHA isn't activating `.bashrc` when running its jobs, but hope it is the fix. I am just including the cell-type-ewings module for now. If this works after testing, I will add the same changes to all docker files and instructions.

### DIFF
--- a/.github/workflows/run_cell-type-ewings.yml
+++ b/.github/workflows/run_cell-type-ewings.yml
@@ -22,6 +22,7 @@ on:
       - main
     paths:
       - analyses/cell-type-ewings/**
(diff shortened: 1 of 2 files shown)


jaclyn-taroni (approved):
👍🏻  if it passes

jaclyn-taroni:
@jashapiro - the analysis module is what is failing?

Issue #676, opened 2024-07-31
Discussion: Module testing GHA path updates


sjspielman:
Two items I've noticed about triggers for module testing GHAs, using this for example:

https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/c6a0a243fdaf3d1268a0789d4664044a252655fa/.github/workflows/run_doublet-detection.yml#L25-L27

1) I wondered if we want to update `path` to _exclude_ `Dockerfile`? The reason for this is if a Dockerfile has been updated, then the environment is not up-to-date for the analysis to run successfully. That said, this _only_ is true if the GHA is actually using the Docker image. So, perhaps this something we'd do internally: When we update GHAs to pull Docker images, we'd also update it to exclude the Dockerfile path as a trigger. ([GHA path docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore))
2) At once point, I had added `.github/workflows/run_doublet-detection.yml` into the `paths` field. We might consider doing this more generally and/or adding it to the template module testing GHA

jashapiro:
Let's do this, yes! 

Note that we will need to revert the change noted in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/746#discussion_r1742470374

As for the Docker question, if changes are only in the `Dockerfile`/`.dockerignore`, then there is not much point in rerunning the workflow whether or not the action is using Docker, so we may as well do that too.

PR #746, opened 2024-09-03
Propagate miniforge Docker changes

### LINKED ISSUE #743: Analysis module failing in CI

jashapiro:
Closes #743

This PR propagates changes from #744 to all Dockerfiles where conda is installed separately through miniforge, and to the docs with miniforge installation instructions for Docker. I had checked that #744 was successful by running the [Run cell-type-ewings analysis module ](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/10687269539/job/29624484348) which is not yet complete, but is successfully running in the provided docker container. 

I also temporarily removed the trigger for `doublet-detection` to run on workflow changes, just because at the moment that would run using the old docker image. Once this PR is in and the new docker image has been pushed to ECR, this change can be reverted.

### DIFF
--- a/.github/workflows/run_doublet-detection.yml
+++ b/.github/workflows/run_doublet-detection.yml
@@ -24,7 +24,9 @@ on:
       - main
     paths:
       - analyses/doublet-detection/**
-      - .github/workflows/run_doublet-detection.yml
+      - "!analyses/doublet-detection/Dockerfile"
+      - "!analyses/doublet-detection/.dockerignore"
+      # - .github/workflows/run_doublet-detection.yml
 
 jobs:
   run-module:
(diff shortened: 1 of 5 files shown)


.github/workflows/run_doublet-detection.yml
```diff
@@ -24,7 +24,9 @@ on:
       - main
     paths:
       - analyses/doublet-detection/**
-      - .github/workflows/run_doublet-detection.yml
+      - "!analyses/doublet-detection/Dockerfile"
+      - "!analyses/doublet-detection/.dockerignore"
+      # - .github/workflows/run_doublet-detection.yml
```
sjspielman, line 29: I had actually added this line in at one point, see here for some context: https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/676

If you want to keep this line deleted for now and come back to this later, I'm fine with that!

jaclyn-taroni, line 29: I think you probably want to track reverting this in an issue.

sjspielman (approved):
LGTM, and looks like bots agree! 

I did wonder if you also wanted to add the `"!analyses/<module name>/<Docker & .dockerignore>"` lines to the `hello-`* module `run` workflow files? The only other module is `metacells` which isn't using Docker yet.

.github/workflows/run_doublet-detection.yml
jashapiro: Yes, I know why it is there, but it would make this PR never pass, as it would keep trying to use the current docker image, not the one defined by the changes to the Dockerfile.

sjspielman: Right, totally understand that, sorry if I wasn't clear!  - I'm _only_ saying if you want to wholesale delete, rather than comment out, that would be ok too.

jashapiro:
> I did wonder if you also wanted to add the `"!analyses/<module name>/<Docker & .dockerignore>"` lines to the `hello-`* module `run` workflow files? The only other module is `metacells` which isn't using Docker yet.

I thought about it, but those modules don't currently actually _use_ the Docker images, so I wasn't as worried about it, so I decided to keep the changes more limited. But I can see it going either way, really.

jaclyn-taroni (commented):
My one comment is let's not forget to undo the thing that is commented out.

jaclyn-taroni (approved):
Sorry, intended to approve!

.github/workflows/run_doublet-detection.yml
jashapiro: Added this as part of #676, if that works?

PR #747, opened 2024-09-03
Update documentation to remove `defaults` channel and suggest miniforge

### LINKED ISSUE #712: Update documentation to remove `defaults` channel and suggest miniforge
As part of https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/705, we will want to update all documentation to suggest using a miniforge installation rather than `miniconda`

A probably incomplete list of pages that need to be updated: 
- [ ] [README.md](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/README.md)
- [ ] [Conda setup](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/technical-setup/environment-setup/setup-conda.md)
  - Change recommendation to `miniforge` with associated instruction updates (no GUI installer)
  - Remove `conda config --add channels defaults` as we don't want the GUI!
  - Add a note about what to do if you already have a conda installation?
- [ ] [Docker images](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/docs/ensuring-repro/docker/docker-images.md?plain=1
  - Change suggested conda base image
  - Change template for R/conda images to follow #711

jashapiro:
Closes #712 

Here I updated the documentation to give instructions on how to install conda via miniforge, rather than with the graphical installer on Mac & command line elsewhere. The instructions should be the same for all installs, so I consolidated the instructions into a single block, with only a note about Windows/WSL. Does that seem reasonable?

I also removed the GitKraken terminal references, as we have seen errors with that (maybe miniforge is better, but I doubt it!).


docs/technical-setup/environment-setup/setup-conda.md
```diff
@@ -25,47 +25,66 @@ There are two main reasons we use conda for OpenScPCA:
...
+```
 
-To install Miniconda onto the WSL 2 side of your Windows machine, you will need to follow the [Quick command line install instructions for Linux](https://docs.anaconda.com/free/miniconda/#quick-command-line-install).
+!!! note
+    On Windows with WSL 2, you will need to install Miniforge on the WSL 2 side of your system, starting from the [Ubuntu terminal](../../getting-started/project-tools/using-the-terminal.md).
 
-- Copy and paste all commands, in order, into the [Ubuntu terminal](../../getting-started/project-tools/using-the-terminal.md).
-  - Be sure to also take the step to initialize conda for the Bash shell.
-- Then, as prompted, close your terminal and open a new terminal window to complete installation.
+You will be prompted first to view the license; press enter as instructed, read the license (you should be able to scroll with your arrow keys), then press `q` to return to the installer.
```
sjspielman, line 49: A little fiddly here...
When I tried this, I found if you just hit the down arrow to scroll, it ends up prompting you for yes/no and you don't ever need `q`. If otoh, you hit `q` without scrolling, it goes straight to yes/no.

docs/technical-setup/environment-setup/setup-conda.md
```diff
@@ -25,47 +25,66 @@ There are two main reasons we use conda for OpenScPCA:
...
+Accept the license by typing `yes`, then continue following the prompts to complete the installation.
+You will be asked about the installation location (the default should be fine).
```
sjspielman, line 51:
```suggestion
You will be asked about the installation location (you can press `enter` to accept the default, which should be fine).
```

docs/technical-setup/environment-setup/setup-conda.md
```diff
@@ -25,47 +25,66 @@ There are two main reasons we use conda for OpenScPCA:
...
-Miniconda is lightweight version of the full conda platform and includes the conda tool itself, Python, and a few other commonly-used packages.
+We recommend installing [Miniforge](https://github.com/conda-forge/miniforge?tab=readme-ov-file#miniforge) to obtain conda.
+Miniforge is lightweight version of the conda distribution that is free and open source.
+The installation includes the conda tool itself, Python, and a few other commonly-used packages.
 
 If you already have conda on your system, you do not need to re-install it.
 
 
-### Install Miniconda on macOS
+### Installing Miniforge
```
jashapiro, line 35: Maybe?
```suggestion
### Installing conda with Miniforge
```

sjspielman: +1

sjspielman (commented):
> The instructions should be the same for all installs, so I consolidated the instructions into a single block, with only a note about Windows/WSL. Does that seem reasonable?

This seems very reasonable to me! 

I recall briefly discussing at one point that we might have instructions "what if i already installed miniconda?" and I now think maybe it's fine to just deal with that on a case-by-case basis instead? Or, we could add a _by default collapsed_ admonition with code, for example, to removing `defaults` from the channels list and other small tasks

docs/technical-setup/environment-setup/setup-conda.md
```diff
@@ -25,47 +25,66 @@ There are two main reasons we use conda for OpenScPCA:
...
 
-To install Miniconda, [download the graphical installer for macOS](https://docs.anaconda.com/free/miniconda/miniconda-install/), and follow all instructions.
+To install Miniforge, follow the [installation instructions in the Miniforge repository](https://github.com/conda-forge/miniforge?tab=readme-ov-file#miniforge).
```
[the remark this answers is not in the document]
jashapiro: Did it hurt you? I don't think it is a big deal if people click the link and then go back.  I could also update the link to `https://github.com/conda-forge/miniforge?tab=readme-ov-file#install`, which is what I think I meant it to be. 
```suggestion
To install Miniforge, follow the [installation instructions in the Miniforge repository](https://github.com/conda-forge/miniforge?tab=readme-ov-file#install).
```

docs/technical-setup/environment-setup/setup-conda.md
jashapiro: Yeah, it can be a bit system dependent... if you hit down after you get to the end it goes back, and actually many other keys will take you back (space, for example). You can also hit return repeatedly to scroll. I just went with what I thought was likely to be most universal when you get dumped into a pager like that.

sjspielman: yeah, i think this is one just is what it is...

jashapiro:
> I recall briefly discussing at one point that we might have instructions "what if i already installed miniconda?" and I now think maybe it's fine to just deal with that on a case-by-case basis instead? Or, we could add a _by default collapsed_ admonition with code, for example, to removing `defaults` from the channels list and other small tasks

I think we can do this on a case-by-case basis, yes. If they use our templates `defaults` won't be in the channel list, so that is a start, but we should be keeping an eye out for it in conda environment files and conda-lock files. But in general it should be pretty easy to deal with, and I would rather keep the docs simpler until proven otherwise.

docs/technical-setup/environment-setup/setup-conda.md
```diff
@@ -25,47 +25,66 @@ There are two main reasons we use conda for OpenScPCA:
...
+    On Windows with WSL 2, you will need to install Miniforge on the WSL 2 side of your system, starting from the [Ubuntu terminal](../../getting-started/project-tools/using-the-terminal.md).
 
-- Copy and paste all commands, in order, into the [Ubuntu terminal](../../getting-started/project-tools/using-the-terminal.md).
-  - Be sure to also take the step to initialize conda for the Bash shell.
-- Then, as prompted, close your terminal and open a new terminal window to complete installation.
+You will be prompted first to view the license; press enter as instructed, read the license (you should be able to scroll with your arrow keys), then press `q` to return to the installer.
+Accept the license by typing `yes`, and then continue following the prompts to complete the installation.
+You will be asked about the installation location (you can press `enter` to accept the default, which should be fine).
+Finally, you will be asked whether you wish to update your shell profile to automatically initialize conda.
+We recommend selecting `yes`.
```
sjspielman, line 53: Bulletifying to make this easier to read
```suggestion
* Accept the license by typing `yes`, and then continue following the prompts to complete the installation.
* You will be asked about the installation location (you can press `enter` to accept the default, which should be fine).
* Finally, you will be asked whether you wish to update your shell profile to automatically initialize conda.
We recommend selecting `yes`.
```

sjspielman (approved):
> But in general it should be pretty easy to deal with, and I would rather keep the docs simpler until proven otherwise.

I don't disagree!

🚀

docs/technical-setup/environment-setup/setup-conda.md
jashapiro, line 53: All the steps are bullets.
```suggestion
* You will be prompted first to view the license; press enter as instructed, read the license (you should be able to scroll with your arrow keys), then press `q` to return to the installer.
* Accept the license by typing `yes`, and then continue following the prompts to complete the installation.
* You will be asked about the installation location (you can press `enter` to accept the default, which should be fine).
* Finally, you will be asked whether you wish to update your shell profile to automatically initialize conda.
We recommend selecting `yes`.
```

PR #749, opened 2024-09-04
Add function to perform graph-based clustering

### LINKED ISSUE #745: Add rOpenScPCA function to perform graph-based clustering

#727 & #731

This issue tracks adding a function to perform graph-based clustering to `rOpenScPCA`. 

- (EDIT!) The function should take a matrix _with row names_ where ~rows are assumed to be genes and columns are assumed to be cells.~ rows are assumed to be cells and columns are PCs(!)
- The function should take the following arguments:
  - algorithm (default louvain)
  - weighting (default jaccard)
  - k
  - number of threads
  - seed
  - optionally, a TSV file to export results to. If `NULL`, nothing will be exported. (see below for TSV contents...)
- The function should return a table of results with these columns (which would be the same as what is exported to TSV if specified): `barcode`, `cluster`, `algorithm`, `weighting`.

A function with associated docs & tests to perform graph-based clustering

N/A

N/A

Unless something else comes up, today!

sjspielman:

Closes #745 

This PR adds a function to perform clustering to the `rOpenScPCA` package. There are many file changes, because I've also merged in `main` to stay in sync, but review can focus only on code in `packages/rOpenScPCA`. 

I'm filing this as a draft to get some initial feedback, and then can implement the rest (including adding more tests, which I know are thin right now!) and mark for review. Here are the main questions I'm looking for feedback on, but of other high-level feedback is welcome :)

- As a starting point (as in, we could do more/less!), I have included three arguments for specifying "additional" clustering parameters:`resolution` (used by louvain and leiden), `objective_function` (used by leiden), and `cluster_args` (for anything else). How does this choice feel to you? Prefer to see fewer or more of the algorithm-specific options pulled out as individual arguments? 
  - Noting further I don't have any checks for these arguments as of yet, but sure could!
- Should the table the function returns contain additional columns, e.g. of other non-default parameters used for the clustering, beyond what I've coded up so far? 
  - At one point I wondered if a `DataFrame` would be better since it store a lot of that info as metadata, but then I felt it might get confusing for users. 
- Any thoughts on [this comment](https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/745#issuecomment-2327032143)? (noting that I edited the opening comment of #745 after I wrote this; the matrix quoted there is the wrong matrix! we want pcs, not expression...)

### Author checklists

_Check all those that apply._


packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -0,0 +1,86 @@
...
+      cluster.fun = algorithm,
+      cluster.args = cluster_args
+    )
+  )
+
+
+  # Transform results into a table and return
+  # TODO: Should this have _all_ (non-default/user-specified) parameters in cluster.args?
+  return(
+    tibble::tibble(
```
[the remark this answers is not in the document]
sjspielman: none whatsoever, this has just become muscle memory for me apparently. can `data.frame()`

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -0,0 +1,86 @@
...
+    "`cluster_args` must be a list." = class(cluster_args) == "list"
+  )
+
+  # Update cluster_args list with settings that users can directly provide
+  # clusterRows throws an error if this list has a param not used by the chosen algorithm
+  if (algorithm != "walktrap") {
+    cluster_args$resolution <- resolution
+  }
+  if (algorithm == "leiden") {
+    cluster_args$resolution <- objective_function
```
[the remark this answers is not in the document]
sjspielman: apparently times have changed! `resolution_parameter` has been superceded by `resolution` https://r.igraph.org/reference/cluster_leiden.html

sjspielman:
Should be ready for a real look now! I think the `expect_error()` checks are probably overkill, but I had a nice time writing them, so here they are for now. Sorry for the clutter..........!

jashapiro:
> Should be ready for a real look now! I think the `expect_error()` checks are probably overkill, but I had a nice time writing them, so here they are for now. Sorry for the clutter..........!

Will review after #752 goes in

jashapiro:
Did I say 752? I meant #753

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
```diff
@@ -0,0 +1,78 @@
+test_mat <- matrix(
+  runif(1000, 0.1, 10),
```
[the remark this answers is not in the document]
jashapiro: If this is a PCA, then we should make the data look like a bit like PCA?
```suggestion
  runif(1000, -3, 3),
```

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
```diff
@@ -0,0 +1,78 @@
...
+    all(cluster_df$resolution == 1)
+  )
+})
+
+
+test_that("calculate_clusters runs with additional cluster_args", {
+  cluster_df <- calculate_clusters(
+    test_mat,
+    algorithm = "leiden",
+    # the default is 2
```
jashapiro, line 41: Does it matter what the default is for this test? If the default is given, that should be fine too!

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
```diff
@@ -0,0 +1,78 @@
...
+    names(cluster_df),
+    c("cell_id", "cluster", "algorithm", "weighting", "nn", "resolution")
+  )
+  expect_equal(
+    cluster_df$cell_id,
+    as.character(1:100)
+  )
+  expect_true(
+    is.factor(cluster_df$cluster)
+  )
```
jashapiro, line 21:
```suggestion
  expect_type(cluster_df$cluster, "factor")
```

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
```diff
@@ -0,0 +1,78 @@
...
+
+
+
+test_that("calculate_clusters errors as expected", {
+  test_mat_nonames <- test_mat
+  rownames(test_mat_nonames) <- NULL
+
+  expect_error(
+    calculate_clusters(test_mat_nonames),
+    "The `mat` matrix must have row names representing cell ids, e.g. barcodes."
```
jashapiro, line 63: In general, I don't think it is worth including the specific error message, as that just means any changes have to be in both places. I also think that the test string is a regex, so periods aren't periods... But if you want to just use `"must have row names"` or some substring like that, I would probably be fine with it.

jashapiro (commented):
This looks good overall, but there are a couple places where there may be some logic updates. 

The biggest one is probably outputting the data frame of arguments for the possible case where you have a `cluster_arg` that is not just one thing; I can imagine a case where one of the arguments might take a vector or list, and we would have to be careful about putting than into a data frame properly. Or we can just check that all the args have length 1, and not support args that don't follow that pattern, as I don't think we expect to use anything like that...

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
```diff
@@ -0,0 +1,78 @@
...
+    cluster_args = list(n_iterations = 3)
+  )
+
+  expect_setequal(
+    names(cluster_df),
+    c("cell_id", "cluster", "algorithm", "weighting", "nn", "resolution", "objective_function", "n_iterations")
+  )
+  expect_true(
+    all(cluster_df$n_iterations == 3)
+  )
```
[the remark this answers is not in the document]
sjspielman: ha, this is funny, i had them all as  `expect_equal(unique(....` at first but switched them in a moment of indecision!

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -0,0 +1,101 @@
...
+#' # cluster using the leiden algorithm with a non-default of 3 iterations
+#' cluster_df <- calculate_clusters(
+#'   pca_matrix,
+#'   algorithm = "leiden",
+#'   cluster_args = list(n_iterations = 3)
+#' )
+#' }
+calculate_clusters <- function(
+    mat,
+    algorithm = c("louvain", "walktrap", "leiden"),
```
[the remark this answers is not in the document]
jashapiro: I don't have a very strong preference, aside from that if you change the defaults from the underlying method that should be made explicit in the docs. In particular, I would add a description after the roxygen heading explaining the motivation for this function and the way it differs from the bluster default/underlying implementation.

sjspielman:
>  Or we can just check that all the args have length 1, and not support args that don't follow that pattern, as I don't think we expect to use anything like that...

I haven't really been able to think of anything better than this... I guess we could also return the `cluster_args` list for users to do what they want with, but given that a minority of users (if any) are likely to be in this circumstance, I do think just the straight data frame return is cleaner.

sjspielman:
Ready for another look, and in particular interested in your feedback on my docs wording about parameter defaults, and overall wording for `cluster_args` length-1 restriction.

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -0,0 +1,115 @@
+#' Calculate graph-based clusters from a provided matrix
+#'
+#' @param mat Matrix, usually of PCs, where each row is a cell. Matrix must have rownames of cell ids (e.g., barcodes)
+#' @param algorithm Clustering algorithm to use. Must be one of "louvain" (default), "walktrap", or "leiden".
+#'  Be aware that the default of "louvain" is different from the bluster package default of "walktrap". This difference is
+#'  because louvain clustering is more commonly-used in scRNA-seq analysis.
```
jashapiro, line 6: I would actually leave off the explanation here. Instead, I was suggesting a paragraph of explanation before the `@params`, which will end up in the function description.

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -0,0 +1,115 @@
+#' Calculate graph-based clusters from a provided matrix
+#'
```
jashapiro, line 2: This is where I was suggestion you put the explanations of default & differences. Something like:

```suggestion
#' Calculate graph-based clusters from a provided matrix
#'
#' This function is provided to simplify application of some bluster functions for OpenScPCA data.
#' Note that defaults for some arguments may differ from the bluster defaults.
#' In particular, the clustering algorithm defaults to "louvain" and the distance measure to "jaccard" 
#' to align with common practice. 
```

jashapiro (commented):
I was actually suggesting something a bit different for the docs; consolidating the discussion of differences into the description. I wrote a bit of what I was thinking there, but you will probably want to rewrite it.

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -0,0 +1,115 @@
...
+
+  if (length(cluster_args)) {
+    stopifnot(
+      "`cluster_args` must be a named list." = is.list(cluster_args) && !("" %in% methods::allNames(cluster_args))
+    )
+  }
+
+  # Do not use any arguments whose length is >1
+  if (any(sapply(cluster_args, length) > 1)) {
+    warning("Only single-length values in 'cluster_args' will be used.")
```
jashapiro, line 70: I would actually error here. You don't want an output that is different from what people were trying to achieve. If they provide an unsupported value, the function should fail rather than discarding input.

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -0,0 +1,115 @@
...
+
+    cluster_args <- cluster_args |>
+      purrr::keep(
+        \(arg) {
+          length(arg) == 1
+        }
+      )
```
jashapiro, line 77: This should do it:
```suggestion
      "`cluster_args` must be a named list." = is.list(cluster_args) && !("" %in% methods::allNames(cluster_args)),
      "`cluster_args` elements must all have only a single value" = all(sapply(cluster_args, length) == 1))
    )
```

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -0,0 +1,106 @@
...
+#' @param mat Matrix, usually of PCs, where each row is a cell. Matrix must have rownames of cell ids (e.g., barcodes)
+#' @param algorithm Clustering algorithm to use. Must be one of "louvain" (default), "walktrap", or "leiden".
+#' @param weighting Weighting scheme to use. Must be one of "jaccard" (default), "rank", or "number"
+#' @param nn Number of nearest neighbors. Default is 10.
+#' @param resolution Resolution parameter used by louvain and leiden clustering only. Default is 1.
+#' @param objective_function Leiden-specific parameter for whether to use the Constant Potts Model ("CPM"; default) or "modularity"
+#' @param seed Random seed to set for clustering. Default is 2024.
+#' @param cluster_args List of additional arguments to pass to the chosen clustering function. Only single-length values will be used.
+#'   See igraph documentation for details on each clustering function: https://igraph.org/r/html/latest
+#' @return A data frame of cluster results with columns `cell_id` and `cluster`. Additional columns represent algorithm parameters
```
jashapiro, line 18: Just separating the return value
```suggestion
#'   See igraph documentation for details on each clustering function: https://igraph.org/r/html/latest
#'
#' @return A data frame of cluster results with columns `cell_id` and `cluster`. Additional columns represent algorithm parameters
```

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -0,0 +1,106 @@
...
+#' to align with common practice in scRNA-seq anaylsis.
+#'
+#' @param mat Matrix, usually of PCs, where each row is a cell. Matrix must have rownames of cell ids (e.g., barcodes)
+#' @param algorithm Clustering algorithm to use. Must be one of "louvain" (default), "walktrap", or "leiden".
+#' @param weighting Weighting scheme to use. Must be one of "jaccard" (default), "rank", or "number"
+#' @param nn Number of nearest neighbors. Default is 10.
+#' @param resolution Resolution parameter used by louvain and leiden clustering only. Default is 1.
+#' @param objective_function Leiden-specific parameter for whether to use the Constant Potts Model ("CPM"; default) or "modularity"
+#' @param seed Random seed to set for clustering. Default is 2024.
+#' @param cluster_args List of additional arguments to pass to the chosen clustering function. Only single-length values will be used.
```
jashapiro, line 16: Slight reword.
```suggestion
#' @param cluster_args List of additional arguments to pass to the chosen clustering function. 
#'   Only single values for each argument are supported (no vectors or lists).
```

jashapiro (approved):
Looks good! Just a couple wording/format changes. Which will require rebuilding docs, unfortunately.

packages/rOpenScPCA/R/calculate-clusters.R
sjspielman: well, "everything" is a vector, but i get it.

Issue #754, opened 2024-09-10
rOpenScpCA: Add function to extract PCA matrix


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/727 & https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/731

### Describe the goals of the changes to the analysis module.

This issue tracks adding a function to extract and return a PCA matrix, with barcode row names, from a given object, which could be either an SCE or Seurat object. I plan to have the function determine directly if the object is SCE or Seurat rather than using an additional argument. This function should be flexible enough for users to be able to specify the name of the PCA matrix in the given object, but we can still have a default.

As I alluded to in [this comment](https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/745#issuecomment-2327032143), I am currently thinking this should be a helper function (that may not be exported? though i suppose can't hurt?) that the `calculate_clusters()` function (currently being written in #749) can use. For that, we'd need to update `calculate_clusters()` to take an object instead of a matrix and call the helper function to convert it. 

A benefit to this alternative approach is users only would need to use `calculate_clusters()` rather than having to extract the matrix themselves, which may be more user-friendly. But, thinking towards the future, we'll also want to be wrapping the `calculate_clusters()` function to be able to sweep parameters (noting that we want to sweep ourselves, not with `bluster::clusterSweep()`, since we want to ensure the seed is set fresh for each parameter iteration), and we'd only like to convert _once_ for that. So, perhaps we allow for either a matrix or object to be passed into `calculate_clusters()` (I don't _love_ the lack of consistency there but I am not strictly opposed to it)? Or, just keep this function separate and users can extract themselves.

Tagging @jashapiro for any thoughts here!

### What will your pull request contain?

Function and tests.

### Will you require additional software beyond what is already in the analysis module?

N/A

### Will you require different computational resources beyond what the analysis module already uses?

N/A

### If known, when do you expect to file the pull request?

This sprint

PR #758, opened 2024-09-12
Add function to extract PC matrix (1/2)

sjspielman:
PR 1/2 for #754 

This PR adds a function and associated tests to extract a PC matrix from a given SCE or Seurat object, allowing for a custom PC name while trying to set reasonable defaults for either object type. Currently I don't export this function, but that decision was kind of arbitrary. I also removed the previous README so that `roxygen` could write to it, and I then cleaned up some imports. (I also learned that `stopifnot()` does not like `paste` or `glue::glue` in the error message, which I think I had previously known but forgot.)

I expect to write a second PR to close out #754 that updates the `calculate_clusters()` function to, as discussed, also be able to take a single cell object and call this function to extract the matrix before clustering when needed.


packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -6,6 +6,8 @@
 #' Specifically, the clustering algorithm defaults to "louvain" and the weighting scheme to "jaccard"
 #' to align with common practice in scRNA-seq analysis.
 #'
+#' @importFrom methods allNames
```
jashapiro, line 9: probably safe to just do the whole thing here
```suggestion
#' @import methods
```

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
```diff
@@ -1,9 +1,24 @@
+suppressPackageStartupMessages(library(SingleCellExperiment))
+suppressPackageStartupMessages(library(Seurat))
+
+sce <- scpcaTools:::sim_sce(n_cells = 100, n_genes = 10, n_empty = 0)
```
jashapiro, line 4: Not a big fan of using an internal function from a different package, even in testing. Maybe use `splatter::simpleSimulate()`?

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
```diff
@@ -1,9 +1,24 @@
...
 )
-rownames(test_mat) <- as.character(1:100)
+rownames(test_mat) <- colnames(sce)
+colnames(test_mat) <- paste0("PC_", 1:10) # quiet seurat warnings
+reducedDim(sce, "PCA") <- test_mat
+
+srat <- CreateSeuratObject(counts = counts(sce), assay = "RNA")
+srat[["pca"]] <- CreateDimReducObject(
+  embeddings = test_mat,
+  key = "PC_",
```
jashapiro, line 17: Is this the default? Or is it what we are creating when we convert?

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
```diff
@@ -1,9 +1,24 @@
+suppressPackageStartupMessages(library(SingleCellExperiment))
+suppressPackageStartupMessages(library(Seurat))
+
+sce <- scpcaTools:::sim_sce(n_cells = 100, n_genes = 10, n_empty = 0)
 test_mat <- matrix(
   runif(1000, -3, 3),
   nrow = 100,
   ncol = 10
 )
```
jashapiro, line 9: At this point you might as well get the real PCA matrix from the SCE?

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -106,3 +108,63 @@ calculate_clusters <- function(
...
+      pc_name
+    )
+  } else if (is(sc_object, "Seurat")) {
+    pc_name <- ifelse(is.null(pc_name), default_seurat, pc_name)
+    stopifnot(
+      "Could not find a PC matrix in the Seurat object." =
+        pc_name %in% names(sc_object@reductions)
+    )
+
+    pca_matrix <- Embeddings(
```
jashapiro, line 156: Is there any reason not to use `::` notation here and avoid importing the whole package?
```suggestion
    pca_matrix <- Seurat::Embeddings(
```

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -106,3 +108,63 @@ calculate_clusters <- function(
...
+    pc_name <- ifelse(is.null(pc_name), default_sce, pc_name)
+    stopifnot(
+      "Could not find a PC matrix in the SingleCellExperiment object." =
+        pc_name %in% reducedDimNames(sc_object)
+    )
+
+    pca_matrix <- reducedDim(
+      sc_object,
+      pc_name
+    )
```
jashapiro, line 148: I'd keep this on one line, as it is not particularly wide.
```suggestion
    pca_matrix <- reducedDim(sc_object, pc_name)
```

jashapiro (commented):
This looks good overall, but I had a few thoughts about imports, mostly.

In general, we probably want to avoid importing too much if we don't need to. There should not be any problem with continuing to use `::` in most cases (I'm thinking especially of Seurat here). On the other hand, in the case of `methods`, it seems reasonable to just import the whole thing.

Relatedly, `scpcaTools` is a pretty heavy requirement, even for just testing, so I would probably try to use something smaller, and with a public simulation function, even if it means we have a bit more code in the test package.

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
sjspielman: Ah, I think this should have be removed; it's not needed.

sjspielman: Ugh nope, it does need it, but it's not relevant to our usage of the Seurat object here. It will forcibly use this to update column names though...

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
sjspielman: I did this with a very quick 'n dirty normalization and `runPCA`, which does add `scater` to Suggests. Is this what you had in mind?

sjspielman:
> In general, we probably want to avoid importing too much if we don't need to

Seurat is out (as is scpcaTools), but I kept SingleCellExperiment as an import since I figured it's relative more lightweight; let me know if you prefer more `::` there

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
```diff
@@ -1,9 +1,20 @@
...
-  nrow = 100,
-  ncol = 10
+suppressPackageStartupMessages(library(SingleCellExperiment))
+
+set.seed(2024)
+sce <- splatter::simpleSimulate(nGenes = 1000)
+assay(sce, "logcounts") <- log1p(counts(sce))
+suppressWarnings({
+  sce <- scater::runPCA(sce)
+})
```
jashapiro, line 8: A bit simpler maybe, since you already have `scater`

```suggestion
set.seed(2024)
sce <- splatter::simpleSimulate(nGenes = 1000, verbose = FALSE) |>
  scater::logNormCounts() |>
  scater::runPCA(ncomponents = 10)
```

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -106,3 +108,59 @@ calculate_clusters <- function(
...
+        pc_name %in% reducedDimNames(sc_object)
+    )
+
+    pca_matrix <- reducedDim(sc_object, pc_name)
+  } else if (is(sc_object, "Seurat")) {
+    pc_name <- ifelse(is.null(pc_name), default_seurat, pc_name)
+    stopifnot(
+      "Could not find a PC matrix in the Seurat object." =
+        pc_name %in% names(sc_object@reductions)
+    )
```
jashapiro, line 150: Test whether Seurat is installed.
```suggestion
    stopifnot(
      "Seurat package must be installed to process a Seurat object" = 
        requireNamespace("Seurat", quietly = TRUE),
      "Could not find a PC matrix in the Seurat object." =
        pc_name %in% names(sc_object@reductions)
    )
```

packages/rOpenScPCA/DESCRIPTION
```diff
@@ -19,10 +19,14 @@ License: BSD_3_clause + file LICENSE
...
 Config/testthat/edition: 3
 RoxygenNote: 7.3.2
 Imports: 
-    bluster,
+    bluster (>= 1.14),
     dplyr,
-    methods
+    methods,
+    SingleCellExperiment,
+    Seurat
```
jashapiro, line 32:
```suggestion
    SingleCellExperiment
```

packages/rOpenScPCA/DESCRIPTION
```diff
@@ -19,10 +19,14 @@ License: BSD_3_clause + file LICENSE
 Encoding: UTF-8
 LazyData: true
 Suggests:
-    testthat (>= 3.0.0)
+    testthat (>= 3.0.0),
+    scater,
```
jashapiro, line 23:
```suggestion
    scater,
    Seurat,
```

jashapiro (commented):
Okay, this looks good, but I realized that I think it is probably worth going one step further and changing `Seurat` from a requirement to a suggestion. I suggested some code to bail if `Seurat` is not installed and a `Seurat` object is given

sjspielman:
Checks passing (locally), and no code styling needed from accepted suggestions which all made sense to me!

jashapiro (approved):
👍🏼

PR #759, opened 2024-09-12
Extract PC matrix within clustering function (2/2)

### LINKED ISSUE #754: rOpenScpCA: Add function to extract PCA matrix

sjspielman:
Closes #754 

This PR does the second half of #754 and allows `calculate_clusters()` to take a single-cell object instead of only a matrix.
I'm opening this as a draft because I couldn't pick between two ways of doing this..so I implemented them both. Which general approach should we go with (noting it might be a little bit "why not both"!)?

- First, in https://github.com/AlexsLemonade/OpenScPCA-analysis/commit/01d538a3f4ae689610e8096aad46819af58dfd0d, I primarily updated the `calculate_clusters()` function and left the extract pc function alone. While writing docs for it though, it felt odd strange to describe the default `pc_name` argument behavior in the clustering function when the extract function was actually defining those defaults. I then wondered if too much was getting duplicated, so I then...
- Implemented the most recent commit shown in the PR, where I moved a decent amount of the behavior out of the extract function and into the clustering function. Here, the extract function requires all arguments.


packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -110,42 +141,34 @@ calculate_clusters <- function(
...
+#' determined by the object_type argument.
 #'
 #' @import SingleCellExperiment
 #' @import methods
 #'
 #' @param sc_object Either a SingleCellExperiment or Seurat object
-#' @param pc_name Optionally, the name of the PC matrix in the object. If this is
-#' not provided, the name "PCA" is assumed for SingleCellExperiment objects, and
-#' "pca" for Seurat objects.
+#' @param object_type Whether the object is a SingleCellExperiment ("sce") or a Seurat ("seurat") object
```
jashapiro, line 154: I don't like this argument; you should check that the object is what was expected anyway, so I would leave the simpler form and check the actual object as you did before.

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -27,36 +33,61 @@
...
 
+  # check and prepare matrix, as needed
+  if (any(class(x) %in% c("matrix", "Matrix"))) {
+    stopifnot(
+      "The matrix must have row names representing cell ids, e.g. barcodes." = is.character(rownames(x))
+    )
+    pca_matrix <- x # redefine with better variable name
+  } else {
+    if (is(x, "SingleCellExperiment")) {
+      pc_name <- ifelse(is.null(pc_name), "PCA", pc_name)
```
jashapiro, line 77: Since the extract function already does this check/replacement, you can just pass it along as defined.

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -27,36 +33,61 @@
...
+      object_type <- "sce"
+    } else if (is(x, "Seurat")) {
+      pc_name <- ifelse(is.null(pc_name), "pca", pc_name)
+      object_type <- "seurat"
+    } else {
+      stop("The first argument should be one of: a SingleCellExperiment object, a Seurat object, or a matrix with row names.")
+    }
+    pca_matrix <- extract_pc_matrix(x, object_type, pc_name)
+  }
```
jashapiro, line 86: Is there any reason this would not work with the previous version of the function?
```suggestion
  } else if (is(x, "SingleCellExperiment") || is(x, "Seurat")) {
    pca_matrix <- extract_pc_matrix(x, pc_name)
  } else {
    stop("The first argument should be one of: a SingleCellExperiment object, a Seurat object, or a matrix with row names.")
  }
```

jashapiro (commented):
I think you may have overcomplicated things! The extract function you had already did all the conversions and checks, so let it handle that. I made a suggestion that I feel like ought to work with pretty minimal code changes, if you revert the extract function to its previous form.

sjspielman:
> I think you may have overcomplicated things! 

Me?!

jashapiro:
> * While writing docs for it though, it felt odd strange to describe the default `pc_name` argument behavior in the clustering function when the extract function was actually defining those defaults.

Just write something like "optional pc_name argument passed to extract_pc_matrix"

packages/rOpenScPCA/R/calculate-clusters.R
```diff
@@ -1,23 +1,29 @@
...
 #' @param algorithm Clustering algorithm to use. Must be one of "louvain" (default), "walktrap", or "leiden".
 #' @param weighting Weighting scheme to use. Must be one of "jaccard" (default), "rank", or "number"
 #' @param nn Number of nearest neighbors. Default is 10.
 #' @param resolution Resolution parameter used by louvain and leiden clustering only. Default is 1.
 #' @param objective_function Leiden-specific parameter for whether to use the Constant Potts Model ("CPM"; default) or "modularity"
-#' @param seed Random seed to set for clustering. Default is 2024.
 #' @param cluster_args List of additional arguments to pass to the chosen clustering function.
 #'   Only single values for each argument are supported (no vectors or lists).
 #'   See igraph documentation for details on each clustering function: https://igraph.org/r/html/latest
+#' @param seed Random seed to set for clustering. Default is 2024.
```
jashapiro, line 23: Just noticed this was not updated when the default was removed 
```suggestion
#' @param seed Random seed to set for clustering.
```

sjspielman:
Should be ready for a _real_ look now. Note that I ended up exporting `extract_pc_matrix` (so I also wrote some quick examples) since we refer to it in the `calculate_cluster` docs now.

packages/rOpenScPCA/tests/testthat/test-calculate-clusters.R
```diff
@@ -67,6 +67,27 @@ test_that("calculate_clusters runs with additional cluster_args", {
   )
 })
 
+test_that("calculate_clusters runs with an object, defaults", {
+  cluster_df_sce <- calculate_clusters(sce)
+  expect_equal(
+    names(cluster_df_sce),
```
jashapiro, line 73: I'd probably make this change everywhere you are comparing column names? (row names we do care about order for)
```suggestion
  expect_setequal(
    colnames(cluster_df_sce),
```

jashapiro (approved):
LGTM, with one little comment about column name checks.

PR #765, opened 2024-09-18
Add function to sweep clustering parameters

### LINKED ISSUE #755: rOpenScpCA: Add function to sweep clustering parameters

This issue tracks adding a function that can be used to sweep clustering parameters, e.g. nearest neighbor values. While there exists a function `bluster::clusterSweep()`, implementing this functionality ourselves will allow for improved reproducibility since we can ensure that the seed is set fresh at _each parameter iteration_.

Additional details will be filled in about this function once #749 & #754 have been implemented since some details may still be up in the air.

Function and tests.

N/A

N/A

_No response_

sjspielman:
Closes #755 

This PR adds a function and tests to perform clustering across a set of parameters. Implementation details:

- First, I am hardly enamored by my function and file naming. Do you agree and if so what thoughts do you have for me??
- The function ends up creating a data frame with an indicator column `cluster_set` to indicate which round of clustering the values pertain to, and for easier `split`ing in the future. Note that I could also leave this as a list and remove the `dplyr::bind_rows()` and allow users to take this step themselves if they prefer. A list might also be preferable since this function will end up getting used in a template Rmd to perform/evaluate clustering parameters, so for a lot of that we'd have to revert to a list anyways.
  - In part because of the need to keep these columns consistent, and also because it felt overkill/beyond the scope of this function's specific purpose, I do _not_ allow users to vary the algorithm. You get to pick 1 algorithm and vary parameters for that algorithm.
- I ended up adding a threads argument, since this can get time consuming with real SCEs
  - Speaking of real SCEs, how long does it take to extract a matrix from an SCE or Seurat object (a real big one, not simulated!)? It's essentially instantaneous. I still extract the matrix to begin with to save some time, but it's small potatoes either way.
- Any other tests we'd like to see?


packages/rOpenScPCA/R/sweep-parameters.R
```diff
@@ -0,0 +1,112 @@
...
+#' # performing louvain clustering, with jaccard and rank weighting, and
+#' # varying the nearest neighbor and resolution parameters.
+#' cluster_df <- calculate_clusters_sweep(
+#'   sce_object,
+#'   weighting = c("jaccard", "rank"),
+#'   nn = c(10, 15, 20, 25),
+#'   resolution = c(0.5, 1)
+#' )
+#' }
+calculate_clusters_sweep <- function(
```
jashapiro, line 58: maybe for naming?
```suggestion
sweep_clusters <- function(
```

jashapiro (commented):
Overall this looks good. 

I think my biggest comment is that I _would_ want to support multiple algorithms, but I can see your argument for not doing so. It definitely does introduce some complexity, but that will have to be handled somewhere anyway... I guess the stongest argument would be that sometimes parameter ranges should be different for different algorithms. But supporting it doesn't mean you have to use it!

Part of that reasoning is that I think you _do_ want to check the parameters against the algorithm anyway: if there are unused parameters given they should not be passed along to the sweep, as that will result in duplicate runs.

I think leaving the output as a list probably makes sense, as it makes it easier to combine different runs later. Once you put in an id column in the data frame, then you have to worry about it being unique.

I didn't evaluate testing yet, as I think it will change if the output is changing!

sjspielman:
Here's a question for you regarding allowing multiple algorithms: What is the right way to handle varying parameters that are shared across algorithms?

For example, here, 
- `objective_function` will apply to only `leiden` since it's algorithm-specific
- but `nn` would apply to _both_ louvain and leiden.
```
sweep_clusters(
  sce, 
  algorithm = c("louvain", "leiden"), 
  objective_function = "modularity", 
  nn = c(10, 15)
)
```

Alternatively, we could have some very fine control. Taking a stab at this conceptually but it feels a bit gnarly? 
```
sweep_clusters(
  sce, 
  list(
    "louvain" = list(nn = c(10, 15), 
    "leiden" = list(nn = c(20, 25), objective_function = "modularity")
  )
)
```

jashapiro:
> Here's a question for you regarding allowing multiple algorithms: What is the right way to handle varying parameters that are shared across algorithms?
> 
> For example, here,
> 
> * `objective_function` will apply to only `leiden` since it's algorithm-specific
> * but `nn` would apply to _both_ louvain and leiden.
> 
> ```
> sweep_clusters(
>   sce, 
>   algorithm = c("louvain", "leiden"), 
>   objective_function = "modularity", 
>   nn = c(10, 15)
> )
> ```
> 
> Alternatively, we could have some very fine control. Taking a stab at this conceptually but it feels a bit gnarly?
> 
> ```
> sweep_clusters(
>   sce, 
>   list(
>     "louvain" = list(nn = c(10, 15), 
>     "leiden" = list(nn = c(20, 25), objective_function = "modularity")
>   )
> )
> ```

I would do the former, but then handle it in the function with`mutate(objective_function = ifelse(algorithm == "leiden", objective_function, NA_character_)` after the `expand_grid`

sjspielman:
> I would do the former, but then handle it in the function with `mutate(objective_function = ifelse(algorithm == "leiden", objective_function, NA_character_)`

I had basically this exact this code in there yesterday before I locked down the algorithm! But in the end, I didn't think it was actually needed since `calculate_clusters` will ignore irrelevant parameters. I suspect we can just toss everything into the parameters list and the situation will sort itself out, but I look forward to finding out what I might be missing 😄

jashapiro:
> I had basically this exact this code in there yesterday before I locked down the algorithm! But in the end, I didn't think it was actually needed since `calculate_clusters` will ignore irrelevant parameters. I suspect we can just toss everything into the parameters list and the situation will sort itself out, but I look forward to finding out what I might be missing 😄

The reason we can't just let `calculate_clusters` ignore it is the multiple runs problem. The `mutate` function should be followed by a `distinct` to remove repeats.

packages/rOpenScPCA/R/sweep-clusters.R
```diff
@@ -77,21 +96,25 @@ sweep_clusters <- function(
...
     nn = unique(nn),
     resolution = unique(resolution),
     objective_function = unique(objective_function)
-  )
+  ) |>
+    # set unused parameters for the given algorithm to their defaults, since we can't provide NA
+    # to match.arg in calculate_clusters
+    dplyr::mutate(
+      resolution = ifelse(algorithm %in% c("louvain", "leiden"), resolution, 1),
+      objective_function = ifelse(algorithm == "leiden", objective_function, "CPM")
```
jashapiro, line 111: I don't really like this because I think we want these values filled in table as NA (or absent from the table).  Can we leave it these NA and deal with them in the function? I would be willing to give up `match.arg` for this, but we probably can work around it too. Maybe something like:

```
if(is.na(objective_function)){
  objective_function = NULL
} else {
  objective_function = match.arg(objective_function)
}
```

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,123 @@
...
+
+  expect_equal(
+    length(sweep_list),
+    4
+  )
+
+  # check colnames one at a time
+  sweep_list |>
+    purrr::walk(
+      \(df) {
```
jashapiro, line 24: Rather than nesting this, we could bind rows and then do the tests? I don't think it matters much.

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,123 @@
...
+
+        expect_equal(
+          df$cell_id,
+          colnames(sce)
+        )
+
+        expect_s3_class(
+          df$cluster,
+          "factor"
+        )
```
jashapiro, line 38: some of these tests are redundant with the individual tests, I think. I would skip these two at least.

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,123 @@
...
+        expect_equal(
+          unique(df$algorithm),
+          "louvain"
+        )
+        expect_equal(
+          unique(df$weighting),
+          "jaccard"
+        )
+        expect_true(
+          all(unique(df$nn) == 10) || all(unique(df$nn) == 15)
```
jashapiro, line 49: I don't think you need both all and unique here?
```suggestion
          all(df$nn == 10) || all(df$nn == 15)
```

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,123 @@
...
+        )
+        expect_true(
+          all(unique(df$resolution) == 0.5) || all(unique(df$resolution) == 1)
+        )
```
jashapiro, line 53: Same as above
```suggestion
        expect_true(
          all(df$resolution == 0.5) || all(df$resolution == 1)
        )
```

jashapiro:
> Speaking of `cluster_args` - this is a problem! We don't check that these are sane for the given algorithm, and I tend to think we indeed should not be in the business of doing so since it depends on `igraph`. That said, `bluster` will fail if an irrelevant parameter is passed in, and that's very much a possibility with the sweep function. Here's what I've thought of (noting I prefer the latter) -
> * We could cancel `cluster_args` for the sweep function
> * We could have users supply this argument a nested list per algorithm, e.g. `cluster_args = list(louvain = ..., walktrap = ....))`.

I would vote for no `cluster_args` in the sweep function. If you really need it, you can write your own sweep. And if there are particular cluster_args worth adding, we can add support as needed.

packages/rOpenScPCA/R/sweep-clusters.R
sjspielman: I confess, I have grown fond of `match.args`, so I'd rather have some ifs.

packages/rOpenScPCA/R/sweep-clusters.R
```diff
@@ -0,0 +1,127 @@
...
+    weighting = unique(weighting),
+    nn = unique(nn),
+    resolution = unique(resolution),
+    objective_function = unique(objective_function)
+  ) |>
+    # set unused parameters for the given algorithm to their defaults, since we can't provide NA
+    # to match.arg in calculate_clusters
+    dplyr::mutate(
+      resolution = ifelse(algorithm %in% c("louvain", "leiden"), resolution, 1),
+      objective_function = ifelse(algorithm == "leiden", objective_function, "CPM")
```
jashapiro: Okay, I looked at the function more closely, and it does appear that these unused parameters are fully not used and are not included in the table, so we should be fine.

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,123 @@
+suppressPackageStartupMessages(library(SingleCellExperiment))
+
+set.seed(2024)
+sce <- splatter::simpleSimulate(nGenes = 1000, verbose = FALSE) |>
+  scater::logNormCounts() |>
+  scater::runPCA(ncomponents = 10)
+
+
+test_that("sweep_clusters works as expected with defaults", {
```
jashapiro, line 9: These are not all defaults?

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,123 @@
...
+  sweep_list <- sweep_clusters(
+    sce,
+    nn = c(10, 15),
+    resolution = c(0.5, 1)
+  )
+
+  expect_equal(
+    length(sweep_list),
+    4
+  )
```
jashapiro, line 19:
```suggestion
  expect_length(sweep_list, 4)
```

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,123 @@
...
+      \(df) {
+        expect_setequal(
+          colnames(df),
+          c("cell_id", "cluster", "algorithm", "weighting", "nn", "resolution")
+        )
+
+        expect_equal(
+          df$cell_id,
+          colnames(sce)
+        )
```
jashapiro, line 33: stylewise: put this and ones like it on one line?

```suggestion
        expect_equal(df$cell_id, colnames(sce))

File: packages/rOpenScPCA/R/calculate-clusters.R [hunk 11]
```diff
@@ -81,12 +86,19 @@ calculate_clusters <- function(
...
   algorithm <- match.arg(algorithm)
   weighting <- match.arg(weighting)
-  objective_function <- match.arg(objective_function)
+
+  # this might be NA if it came from the sweep_clusters() function
+  if (is.na(objective_function)) {
+    objective_function <- NULL
+  } else {
+    objective_function <- match.arg(objective_function)
+  }
```
jashapiro, line 101: On reflection, since this parameter only shows up in the output table if it is not `NA`, the previous solution should solve this just fine, and you can revert this change! I thought it would end up included with the default value, but we already don't do that because of the way we (properly!) handle cluster_args. (If you do want to keep it though, you will need to wrap `is.na` with an `all()` or check the length, I realized.)

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,123 @@
...
+          expect_setequal(
+            colnames(df),
+            c("cell_id", "cluster", "algorithm", "weighting", "nn")
+          )
+        } else if (unique(df$algorithm) == "louvain") {
+          expect_setequal(
+            colnames(df),
+            c("cell_id", "cluster", "algorithm", "weighting", "nn", "resolution")
+          )
+        }
```
jashapiro, line 120: Need to catch if neither of the cases fit. Not sure this is the best way, or if you want to be more explicit.
```suggestion
        } else {
          fail("unexpected algorithm value")
        }
```

jashapiro (commented):
Mostly style comments here, but also I think you can revert the last change (setting and dealing with NA). I hadn't appreciated how well we throw out the default values, so I think we can leave your previous solution which did what I was worried it did not!

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
sjspielman: I actually had that originally for all tests _except_ column names since we definitely want to be sure about those on an individual-df level. I ended up moving everything into the `map` since I thought that was a smidge, but really only a smidge..., more precise

packages/rOpenScPCA/R/sweep-clusters.R
```diff
@@ -0,0 +1,126 @@
...
+
+  # Collect all specific inputs into a single list
+  sweep_params <- tidyr::expand_grid(
+    algorithm = unique(algorithm),
+    weighting = unique(weighting),
+    nn = unique(nn),
+    resolution = unique(resolution),
+    objective_function = unique(objective_function)
+  ) |>
+    # set unused parameters for the given algorithm to their defaults
```
jashapiro, line 102: I just want to clarify the reasoning here
```suggestion
    # set unused parameters for each algorithm to default; this will allow duplicates to be removed by distinct()
```

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,112 @@
...
+  )
+
+  # count algorithms
+  alg_counts <- sweep_list |>
+    purrr::map(
+      \(df) {
+        unique(df$algorithm)
+      }
+    ) |>
+    purrr::reduce(c)
```
jashapiro, line 89: I don't care too much, but for a really simple function, this seemed like a lot of blank space!
```suggestion
    purrr::map(\(df) unique(df$algorithm)) |>
    purrr::reduce(c)
```

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,112 @@
...
+
+set.seed(2024)
+sce <- splatter::simpleSimulate(nGenes = 1000, verbose = FALSE) |>
+  scater::logNormCounts() |>
+  scater::runPCA(ncomponents = 10)
+
+
+test_that("sweep_clusters works as expected with default algorithm & weighting", {
+  sweep_list <- sweep_clusters(
+    sce,
```
jashapiro, line 11: since there is logic in the sweep function for the conversion, it is probably worth including a quick test that this function works for matrices and Seurat objects as well. I don't think you need to do anything but check that those functions run and produce output, so using full defaults (one clustering) should be fine.

jashapiro (approved):
LGTM

I suggested two more quick tests (Seurat and matrix inputs), but I don't think I need to see this again.

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,159 @@
...
+  sweep_list |>
+    purrr::walk(
+      \(df) {
+        expect_setequal(
+          colnames(df),
+          c("cell_id", "cluster", "algorithm", "weighting", "nn", "resolution")
+        )
+      }
+    )
+})
```
jashapiro, line 69: Because everything else is already tested, I really meant it that this could be simple!
```suggestion
  sweep_list <- sweep_clusters(test_mat)

  expect_length(sweep_list, 1)
```

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
```diff
@@ -0,0 +1,159 @@
...
+    nn = c(10, 15),
+    # only used by louvain
+    resolution = c(0.5, 1)
+  )
+
+  # count algorithms
+  alg_counts <- sweep_list |>
+    purrr::map(\(df) unique(df$algorithm)) |>
+    purrr::reduce(c)
+  expect_equal(length(alg_counts), 6)
```
jashapiro, line 137: While I was looking one last time.
```suggestion
  expect_length(alg_counts, 6)
```

packages/rOpenScPCA/tests/testthat/test-sweep-clusters.R
sjspielman: ah!

PR #768, opened 2024-09-19
Syncing main into feature/ropenscpca

sjspielman:
Housekeeping!


jashapiro:
How did you create this branch? I would not expect it to be out of date with the base branch.

My method would be to branch from `upstream/feature/ropenscpca` and then merge in `upstream/main`... but it looks like you maybe just created a branch from `main`?

sjspielman:
> How did you create this branch? I would not expect it to be out of date with the base branch.
> My method would be to branch from upstream/feature/ropenscpca and then merge in upstream/main... but it looks like you maybe just created a branch from main?

Yeah, I created a branch from main to merge it into ropenscpca, but that makes more sense!

Issue #774, opened 2024-09-23
Annotation (label transfer) exploration for SCPCP000006


maud-p:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

This PR is following the discussion from the PR#750, especially:
https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/750#pullrequestreview-2310191830

### Describe the goals of the changes to the analysis module.

In this PR, I am adding one notebook in `/notebook/04_annotation_Across_Samples_exploration.Rmd` to explore the annotations and label transfers for all of the samples in SCPCP000006.

We integrated all the samples from SCPCP000006 to have a rapid and global view of label transfer. 
Please note that the integration is not the aim of this PR, this is just a way to display better genes and features. 

In order to explore the label transfer results, we look into some marker genes, table and percentages of cells in each annotation groups (from label transfers).

### What will your pull request contain?

One notebook that explores for all samples at once clustering and label transfer results.

### Will you require additional software beyond what is already in the analysis module?

no

### Will you require different computational resources beyond what the analysis module already uses?

no

### If known, when do you expect to file the pull request?

Today or tomorrow (European time)

sjspielman:
This notebook was completed.

PR #750, opened 2024-09-04
Wilms tumor 06- clustering exploration

maud-p:
#### Please link to the GitHub issue that this pull request addresses.

this is the following work on PR #704 taking into account the changes in PR #737

#### What is the goal of this pull request?

The aim here is to explore the clustering and label transfer from the 2 fetal references for each sample. 

#### Briefly describe the general approach you took to achieve this goal.

Here I started from the output of the notebook `[02b_label-transfer_fetal_kidney_reference_Stewart.Rmd](https://github.com/maud-p/OpenScPCA-analysis/blob/03_Look_into_clustering/analyses/cell-type-wilms-tumor-06/notebook_template/02b_label-transfer_fetal_kidney_reference_Stewart.Rmd)`  that contains:

- Normalisation with `SCTransform`, 
- dimensionality reduction `PCA`, `UMAP`, 
- clustering and 
- label transfer from the 2 fetal references

and explored the results looking at:

- marker genes (from a list of known marker genes and from differential expression analysis, as a more "discovery" approach)
- enriched pathways (from a list of known marker genes and from differential expression analysis, as a more "discovery" approach)

I compared the labels obtained from SingleR, CellAssign and the label transfer from the two fetal references (PR #737).

#### If known, do you anticipate filing additional pull requests to complete this analysis module?

Yes! More than one.  I think that from this analysis, I can find a way to annotate healthy cells such as "immune" and "endothelial cells". From here, I will be able to fill a new PR to include inferCNV and/or copyKAT to the template.

### Results
The notebook template produce a notebook per sample in `notebook/{sample_id}` folder. I have now uploaded the notebooks for the 2 first samples. Once we have discussed the analysis, I'll run for the 40 samples and add the notebooks!

#### What types of results does your code produce (e.g., table, figure)?
notebook
### Provide directions for reviewers

What do you think?

#### What are the software and computational requirements needed to be able to run the code in this PR?

I render the notebook from the `[00_run_workflow.R](https://github.com/maud-p/OpenScPCA-analysis/blob/03_Look_into_clustering/analyses/cell-type-wilms-tumor-06/00_run_workflow.R)` script. I open a new loop on purpose in order not to run everythink from PR #737 again, but I guess in a final step all the notebook will be ran i the same loop!

#### Is there anything that you want to discuss further?

I like to have your opinion on the best way to go to select normal cells as input for inferCNV. I am quite satisfyied by the labels from the fetal kidney reference `fetal_kidney_predicted.compartment` divided into:
- endothelial
[...]

PR #776, opened 2024-09-23
04_annotation_Across_Samples_exploration

maud-p:
https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/774

This PR is following the discussion from the PR#750, especially:
https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/750#pullrequestreview-2310191830

In this PR, I am adding one notebook in /notebook/04_annotation_Across_Samples_exploration.Rmd to explore the annotations and label transfers for all of the samples in SCPCP000006.

We integrated all the samples from SCPCP000006 to have a rapid and global view of label transfer.
Please note that the integration is not the aim of this PR, this is just a way to display better genes and features.

In order to explore the label transfer results, we look into some marker genes, table and percentages of cells in each annotation groups (from label transfers).

Yes, next step would be to run `copyKAT`

### Results

This PR do not contain any result, only a single `notebook`.

One notebook that explores for all samples at once clustering and label transfer results.

- 85.4925176 % of the cells are labeled as kidney cells (fetal full reference, looking at the `fetal_full_predicted.organ`. I think this is quite a nice result. From the `umap` and `barplot`, I think that most of the cells that are not labelled as kidney are endothelial or immune cells. (While writting this, I think it would be good to add a table of `fetal_full_predicted.organ`  and `fetal_kidney_predicted.compartment`, maybe in the next round after your review!)

- 0.8660364 % of the cells are labeled as immune cells and 0.8226098 % of the cells are labeled as endothelial cells. As Wilms tumor is known to be a cold tumor (immune excluded), and COG Wilms tumor samples are mostly not pre-treated, it is quite expected to have very few immune cells. If this is a problem for running copyKAT with very few cells as a reference, I have no idea to be honnest. 

I updated the renv.lock file, else no specific changes since the last PR :)

What about the next step, what do you think how we can run cpyKAT/inferCNV? I think I should start with copyKAT and try to run with and without a reference of normal cells, and try to evaluate what is the impact of the annotation of normal cells on the infered CNV. 


sjspielman (commented):
Hi @maud-p, thanks for getting going on the next PR here! I actually think we want to approach this notebook a bit differently. One reason is, I am not able to run it on my laptop since I run out of memory during the step where you `merge` all the individual Seurat objects. Since integration itself is not the main goal of this notebook, I suggest we remove that part and pursue it later with a dedicated analysis that we know will leverage more computational resources. 

Rather than looking at all samples as an integrated object, let's look at objects individually, just all in one place! I don't think I was clear about this when I said we should analyze samples together, so I'm very sorry for not being explicit! The primary goal of this notebook is to help us make decisions for next steps, and asking some summary questions about all samples in one notebook can make that a lot easier. We want to specifically dig into the `fetal_kidney_predicted.compartment` and `fetal_full_predicted.organ` annotations. A lot of the plots here are the same sort of exploratory analysis we did on individual samples but instead run on the integrated object. Instead, rather than being more exploratory with the same plots again, we want to answer specific questions about whether it seems safe to use these fetal references in next steps. 

To begin, I suggest we can make some tables of counts. For each sample, I'd like to see:
	- How many total cells?
	- What percentage _and count_ (looking at both is helpful!) of cells did the `organ` reference predict to be kidney?
	- What percentage _and count_ of cells did the `compartment` reference predict to be immune? Epithelial?
	
To help get this started, I've written some code here that can be used to read in and prepare this data for analysis in a way that will definitely fit in memory (we can always modify this later to grab more info out of the seurat objects if we need!)

This code will read in each seurat object and extract a data frame of the annotations of interest, and then finally combine it all into one data frame.

_That said_ - I have a question for you. When I tried to run this code myself, I get an error that the Seurt Rds result file for `SCPCS000196` is missing. Indeed, I don't see it in your bucket. Can you check to make sure all result files have been uploaded to the bucket? Thanks!

```
metadata <- readr::read_tsv(file.path(data_dir, "single_cell_metadata.tsv"))

# Create a data frames of all annotations
cell_type_df <- metadata$scpca_sample_id |>
  purrr::map(
    # For each sample_id, do the following:
    \(sample_id) {

      # Read in the Seurat object
      srat <- readRDS(
        file.path(result_dir, sample_id, paste0("02b-fetal_kidney_label-transfer_", sample_id, ".Rds"))
      )

      # Create and return a data frame from the Seurat object with relevant annotations
      # this data frame will have four columns: barcode, sample_id, compartment, organ
      data.frame(
        compartment = srat$fetal_kidney_predicted.compartment,
        organ = srat$fetal_full_predicted.organ
      ) |>
        tibble::rownames_to_column("barcode") |>
        dplyr::mutate(sample_id = sample_id)
    }
  ) |>
  # now combine all dataframes to make one big one
  dplyr::bind_rows()
```

maud-p:
Hi @sjspielman , Thank you for the rapid feedback 🏎️ ! 
I just reloaded the results to the s3 bucket, might have been a bug during the transfer to s3 because my local `results` directory does contain the 3 `rds` objects per samples.
I'll plot the different tables and come back to you, thanks for the few lines of codes to extract the metadata :)

maud-p:
@sjspielman I modified the notebook based on your suggestion, it is quite simple now but let me know if you thing to additional plots and checks that can be useful :) 

Thank you again for your help!

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
+---
+title: "Annotation exploration for `r params$scpca_project_id`"
+author: "Maud PLASCHKA"
+date: '2024-08-13'
+params:
+  scpca_project_id: "SCPCP000006"
+  sample_id: "SCPCS000176"
+  seed: 12345
```
sjspielman, line 8: We don't need any of these for this notebook actually, so they can be removed

```suggestion
```

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+                      warnings=FALSE
+                      )
+
+```
+
+
+## Introduction
+
+
+The aim is to explore the annotations and label transfers for all of the samples in `r params$scpca_project_id`.
```
sjspielman, line 31: For this module, it will always be the same project :) 
```suggestion
The aim is to explore the annotations and label transfers for all of the samples in SCPCP000006.
```

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+
+### Packages
+
+
+```{r packages, message=FALSE, warning=FALSE}
+library(Seurat)
+library(tidyverse)
+library(patchwork)
+
+set.seed(params$seed)
```
sjspielman, line 45:
```suggestion
```

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+The label transfer have been performed in the notebook `02a_fetal_full_reference_Cao_{sample_id}.Rmd`
+
+The fetal full reference (Cao et al.) provides three levels of annotations:
+
+- `fetal_full_predicted.organ` is one of the 15 organs used for building the reference. 
+We were nicely surprised that, for many of the samples, most of the cells from the Wilms tumor cohort mapped to fetal kidney cells. 
+
+- `fetal_full_predicted.annotation.l1` holds predicted cell type annotation
+
+- `fetal_full_predicted.annotation.l2` holds the combination of the predicted organ and predicted cell type annotation
```
sjspielman, line 213: Make this specific to the notebook:

```suggestion
This notebook looks at the `fetal_full_predicted.organ` annotation from this reference. 
```

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
+---
+title: "Annotation exploration for `r params$scpca_project_id`"
+author: "Maud PLASCHKA"
+date: '2024-08-13'
```
sjspielman, line 4: If you use this instead, it will automatically update when knit to html:
```suggestion
date: "`r Sys.Date()`"
```

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+
+```{r fig.width=25, fig.height=6,  results='asis'}
+tmp <- table(cell_type_df$sample_id, cell_type_df$organ)
+
+knitr::kable(tmp)
+```
+
+```{r fig.width=25, fig.height=6}
+do_Table_Heatmap(first_group = "sample_id", last_group = "organ")
+```
```
sjspielman, line 225: We can delete these chunks and replace with this code to make a table of counts for kidney vs not kidney. Then, style the data table however you usually like to do!

This table will have 3 columns in addition to the sample id: a total cell count, the percentage of cells labeled kidney, the percentage of cells labeled anything else that isn't kidney. This will make it easy for us to see which samples do really well, and which don't, for this reference.

```
kidney_count_df <- cell_type_df |>
  # make a new variable that tells us if it's kidney or not
  dplyr::mutate(organ_type = ifelse(organ == "Kidney", "kidney", "not_kidney"))  |>
  # count how many kidney or not
  dplyr::count(sample_id, organ_type) |>
  # make the data frame wide and use a value of 0 if a count is missing
  tidyr::pivot_wider(names_from = organ_type, values_from = n, values_fill=n) |>
  # add a column for total, and convert other counts to percentages
  dplyr::mutate(
    total = kidney + not_kidney, 
    kidney = round(kidney/total, 4)*100, 
    not_kidney = round(not_kidney/total, 4)*100) |>
  # arrange in descending order of percentage of kidney cells
  dplyr::arrange(desc(kidney)) 

DT::datatable(kidney_count_df)
```

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+    - `immune` subtypes: myeloid cells, lymphocytes, ...
+    
+    - `stroma` subtypes: mesenchymal stem cells, fibroblasts, ...
+    
+    - `fetal_nephron` subtypes: podocytes, uteric bud, mesenchymal cells (that appear to be cap mesenchyme cells) ...
+    
+    - `endothelial` cells
+
+
+We think that the first level of annotation `fetal_kidney_predicted.compartment` can be useful in identifying non-cancer cells, at least the endothelial and immune cells.
```
sjspielman, line 262: Can be removed for this notebook

```suggestion
```

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+```{r fig.width=25, fig.height=3, results='asis'}
+tmp <- table(cell_type_df$sample_id, cell_type_df$compartment)
+
+knitr::kable(tmp)
+```
+
+```{r fig.width=25, fig.height=3}
+
+do_Table_Heatmap(first_group = "sample_id", last_group = "compartment")
+```
```
sjspielman, line 274: Replace these with the following and again style the DT however you like!

This table will show counts for each of the cell types, and again will make it easy for us to see which samples seem to have a decent population of cells we think are normal.

```
compartment_df <- cell_type_df |>
  # count how many per compartment
  dplyr::count(sample_id, compartment) |>
  # make the table wide and use a value of 0 if a count is missing
  tidyr::pivot_wider(names_from = compartment, values_from = n, values_fill = 0) |>
  # order the columns how we want to show them
  dplyr::select(sample_id, fetal_nephron, stroma, endothelium, immune) 

DT::datatable(compartment_df)
```

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+
+### Output file
+
+The report will be saved in the `notebook` directory. 
+
+
+## Functions
+
+
+#### do_Table_Heatmap
```
sjspielman, line 126: You can delete this function, based on my other suggestions.

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+### Evaluate annotations with marker genes
+
+Here we evaluate with marker genes the identification of endothelial, immune cells sub-populations.
+
+
+##### Endothelial cells
+
+We look at the endothelial marker "ENSG00000110799" = "VWF"
+
+```{r fig.height=3, fig.width=15, message=FALSE, warning=FALSE, out.width='100%'}
```
sjspielman, line 287: This needs to be a lot taller to be able to see. I can't directly suggest on this line with GitHub, but try like 6 inches?

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+do_Feature_mean(  feature = "ENSG00000110799",
+                  group.by = "compartment")
+```
+
+##### Immune cells
+
+We look at the immune marker "ENSG00000081237" = "PTPRC" alias "CD45"
+
+
+```{r fig.height=3, fig.width=15, message=FALSE, warning=FALSE, out.width='100%'}
```
sjspielman, line 297: Same here, try 6 inches?

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+
+##### Endothelial cells
+
+We look at the endothelial marker "ENSG00000110799" = "VWF"
+
+```{r fig.height=3, fig.width=15, message=FALSE, warning=FALSE, out.width='100%'}
+do_Feature_mean(  feature = "ENSG00000110799",
+                  group.by = "compartment")
+```
+
```
sjspielman, line 291: Let's also try a boxplot and compare to see which kind of plot we think is more informative. What do you think? I wrote this code to start:

```
ggplot(cell_type_df, 
    aes(x = compartment, y = ENSG00000081237, fill = compartment)
  ) + 
  geom_boxplot(size = 0.5, size.outlier = 0.25) +
  facet_wrap(vars(sample_id), scale = "free_y") + 
  theme_bw() +
  theme(
    axis.text.x = element_text(size = rel(0.75), angle = 30), 
    legend.position = "none"
  )
```

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,310 @@
...
+#### do_Feature_mean
+
+`do_Feature_mean` shows heatmap of mean expression of a feature grouped by a metadata. 
+
+- `group.by` is the name of the metadata to group the cells
+- `feature` is the name of the gene to average the expression
+
+```{r fig.width=10, fig.height=4, out.width='100%'}
+do_Feature_mean <- function(group.by, feature ){
```
sjspielman, line 172: If we end up keeping this function, it should also definitely take the data frame (`cell_type_df`) as a formal argument and not just assume it's defined.

sjspielman (commented):
> it is quite simple now but let me know if you thing to additional plots and checks that can be useful :)

In this case simple is exactly what we're looking for, so this is looking great! Actually, I think we can simplify it further too. My hope is that we can use this notebook to pick a couple samples that we think we perform well to run with `copyKAT` and/or `inferCNV` and begin the next steps in earnest.

Speaking of simplifying further - I'm going to suggest a slightly different approach to making the tables because they are very long so hard to read, so it would be great to use the `DT` package. For this, I have suggested code to get these counts (or percentages) in a data frame form that can be passed into `DT`. As part of this, I think we can combine the two tables too so that is part of my code suggestion.

I don't think we need the heatmap of _counts_ since this is the same as what's in the tables, so I suggest to remove those.

For the other heatmaps with gene expression, these are too small to really see so they'll need to be bigger. Or, I've suggested that maybe we could try boxplots here as a different way to visualize the gene expression and identify samples with high gene expression in the expected "compartment".

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
maud-p: This is great thank you. I didn't know the function `pivot_wider`, that is great!
It is now running with `values_fill=0` instead of `values_fill=n`, if I understood correctly the help of the function and your code :)

maud-p:
Dear @sjspielman , thank you so much for the review and the suggestions, including codes! Was really useful, as usual! 

I should have made all suggested changes :D

One additional table that I would like to discuss is the sumary, for each `compartment`, of the percentage of cells that do match kidney annotation or not. 
image

The majority of fetal nephron cell (92%) has been predicted as kidney. However, the other `compartments` (stroma, endothelial and immune) do not really match to kidney cells. This is in my opinion not a concern and shouldn't be interpreted as a poor label transfer or annotation! I'll try to argue a bit why.

- For the stroma compartment: it is known that Wilms tumor stroma (sometimes) shows (unexpected) differentiation into cell types such as skeletal muscle cells, fat tissue, cartilage, bone and even glial cells [[1](https://academic.oup.com/jnci/article/89/15/1089/2526288)-[2](https://www.ncbi.nlm.nih.gov/books/NBK373364/)]. We also saw it on H&E staining of Wilms tumor biopsies. For that reason, I am not surprised that most stroma cells are not predicted as kidney cells. 
- For the immune compartment: it wouldn't be surprising that cancer cells, and/or treatment, modulate the immune microenvironment, via the attraction of immune cells that are not usually in the kidney and/or induction of a cancer-associated phenotype. 

Thank you again for your help!

sjspielman (commented):
Looking good, and I like the tables you added too! I think we're just about there so this probably the last little review :) 

I think we should just add a little more text:

- Let's just add a quick sentence for table that _are not_ sample-specific that clearly state that the table contains all samples pooled into one
- The explanation you commented about why the stroma annotations aren't necessarily concerning makes sense to me, and I think we should actually include it in the notebook! For this, I suggest we add a "Conclusions" section at the end of the document that has:
  - Your explanation about why we might expect that stroma cells don't get annotated as kidney. It would be helpful to include a reference from the literature to back this up, too.
  - Decide on 2-4 samples that you think are good candidates to proceed with and why (e.g., samples where most cells annotated as kidney and that have a decent amount of endo/immune cells). Looking across samples, these look like they might be good choices? SCPCS0000184, SCPCS0000208, SCPCS0000188, SCPCS0000179. What do you think of those?

Just a note about next steps after this PR - I think you should try running `copyKAT` on a few samples _both_ with and without providing a list of normal cells. That will be useful to us to see how consistent results are, and it will inform us about how we might analyze samples that aren't predicted to have that many (or sometimes any at all!) endo/immune cells.

analyses/cell-type-wilms-tumor-06/notebook/04_annotation_Across_Samples_exploration.Rmd
```diff
@@ -0,0 +1,363 @@
...
+              options = list(  dom = 'Bfrtip',
+                               buttons = c( 'csv', 'excel')))
+```
+
+
+
+
+### Evaluate annotations with marker genes
+
+Here we evaluate with marker genes the identification of endothelial, immune cells sub-populations.
```
sjspielman, line 315: Let's also add a bit more:

```suggestion
Here we evaluate with marker genes the identification of endothelial, immune cells sub-populations.
We expect that cells labeled as endothelial and immune cells will have higher expression of these markers compared to other cell types. 
```

sjspielman:
Ah, one more thing I forgot! 

We should add rendering this notebook to the `00_run_workflow.R` script. This step should be after/outside the for loop (since we only knit this once, not for each sample), and only run if we are not testing (since the input files for this notebook aren't generated in testing).

maud-p:
I ended up selecting the following samples: 

- sample SCPCS000194 has > 85 % of cells predicted as kidney and 234 + 83 endothelium and immune cells.
- sample SCPCS000179 has > 94 % of cells predicted as kidney and 25 + 111 endothelium and immune cells.
- sample SCPCS000184 has > 96 % of cells predicted as kidney and 39 + 70 endothelium and immune cells.
- sample SCPCS000205 has > 89 % of cells predicted as kidney and 92 + 76 endothelium and immune cells.
- sample SCPCS0000208 has > 95 % of cells predicted as kidney and 18 + 35 endothelium and immune cells.

I tried to enriched in samples having >100 normal cells, without decreasing too much the prediction of kidney cells. Would it be OK like this?

Thank you again for your help!!

analyses/cell-type-wilms-tumor-06/notebook/README.md
```diff
@@ -26,4 +26,12 @@ In brief, we used `Azimuth` to transfer labels from the `Azimuth` fetal full ref
...
\ No newline at end of file
+In brief, we explore the clustering results, we look into some marker genes, pathways enrichment and label transfer.
+
+
+## Global analysis
+
+At some step during the analysis, a global picture was required to interprete correctly the results and take the best decision for the next steps. 
+
+- `04_annotation_Across_Samples_exploration.html` is the output of the [`04_annotation_Across_Samples_exploration.Rmd`](../notebook/04_annotation_Across_Samples_exploration.Rmd) notebook. 
+In brief, we explored the label transfer results across all samples in the Wilms tumor dataset SCPCP000006.
```
sjspielman, line 37: I'm suggesting a small text change here to add a bit more context about why we are exploring these results:

```suggestion
The next step in analysis is to identify tumor vs. normal cells.

- `04_annotation_Across_Samples_exploration.html` is the output of the [`04_annotation_Across_Samples_exploration.Rmd`](../notebook/04_annotation_Across_Samples_exploration.Rmd) notebook. 
In brief, we explored the label transfer results across all samples in the Wilms tumor dataset SCPCP000006 in order to identify a few samples that we can begin next analysis steps with.
```

sjspielman (approved):
> I tried to enriched in samples having >100 normal cells, without decreasing too much the prediction of kidney cells. Would it be OK like this?

Yes, these look like great choices! 

Thanks for adding the notebook to the `00_run_workflow.R` script, and the `README` text is helpful too. I've made one small wording suggestion for the `README`, but this seems ready to go so I'm approving now!

After you accept the suggestion, I'll go ahead and get this merged in so you can start on next steps of running `copyKAT` to see whether it can help us identify tumor vs. normal in these samples. For that analysis, it would be great if you can write the code to run on only 1 sample at a time, similar to how you wrote the template notebooks. However, instead of a notebook to run `copyKAT`, you may wish to write this as an R script instead, and then have a separate notebook which explores those results. I am suggesting this because it will be helpful for eventually including your cell typing analysis in [our OpenScPCA-nf workflow](https://openscpca.readthedocs.io/en/latest/ensuring-repro/openscpca-nf/), which we use to generate the "official" OpenScPCA-analysis results for all completed modules.


(Just fyi- I am out of the office for the rest of this week (I will be back next Monday Oct 7), so depending when you file the next PR I may or may not be your reviewer.)

maud-p:
Great, thank you very much!

Thank you very much for letting me know about your days off. Then I might take the time to compare few different methods (copyKAT +/- reference, inferCNV +/- reference) on the few samples!

For the workflow, did I understood that it would be best to have:
- R scripts for steps that are building the final object
- notebooks for reports and results exporations?

Should I try in another PR to split the first 3 notebooks into scripts + notebook?

Thank you!

sjspielman:
> Should I try in another PR to split the first 3 notebooks into scripts + notebook?

Don't worry about this at all! The code you have written so far is completely fine. Again, not a strict rule :)

PR #780, opened 2024-09-24
Do not print messages about syncs completing when using --dryrun

jaclyn-taroni:

When using `scripts/sync-results.py` with `--dryrun`, I found that the printed messages could be misleading. In particular, I don't think we want to tell folks files have been downloaded or are now available on S3 if they use `--dryrun`. I could be persuaded that "Sync complete" should be outside of this logic.

### DIFF
--- a/scripts/sync-results.py
+++ b/scripts/sync-results.py
@@ -146,11 +146,12 @@ def main() -> None:
         )
         sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
LGTM. I don't have much of an opinion about "Sync complete"

Issue #782, opened 2024-09-26
Anchor transfer for Wilms tumor annotation (SCPCP000014) 


JingxuanChen7:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/628

### Describe the goals of the changes to the analysis module.

The goal here is to conduct anchor transfer on Wilm tumor dataset (SCPCP000014), so that we could have an overview of the cell type components in each sample.

More specifically, it will contain following parts:
- Prepare a Seurat object for fetal kidney atlas (https://www.kidneycellatlas.org/fetal-kidney-global), which includes two levels of annotation: `compartment` and `celltype`.
- Conduct anchor transfer at `compartment` level.
- Conduct anchor transfer at `celltype` level.

I will also make slight modifications on the preprocessing functions filed in my first PR (https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/764):
- Starting from raw count and adding the Seurat `logNormalize` function back during pre-processing. If the normalization method is not consistent between reference and query object, too few numbers of anchors would be identified.
- Remove genes with unknown gene symbols.

### What will your pull request contain?

This PR would contain code & doc to run anchor transfer, and documentation for corresponding results.

### Will you require additional software beyond what is already in the analysis module?

I will keep using R. Modification on packages would reflect in `renv.lock` file.

### Will you require different computational resources beyond what the analysis module already uses?

 Standard-4XL virtual computer (i.e, 16 cpus and 32 GB RAM) on Lightsail.

### If known, when do you expect to file the pull request?

~Early October

jaclyn-taroni:
Hi @JingxuanChen7, I just wanted to point out that the other Wilms tumor cell typing module ([`cell-type-wilms-tumor-06`](https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-wilms-tumor-06)) is also using fetal atlases for transfer in case taking a look at that analysis would be helpful to you!

JingxuanChen7:
Dear @jaclyn-taroni , thanks for pointing out this similar analysis! 

[...]

PR #784, opened 2024-09-27
Anchor transfer (with Seurat) for Wilms tumor annotation (SCPCP000014)

JingxuanChen7:

The goal for this PR is to conduct anchor transfer on Wilm tumor dataset (SCPCP000014), so that we could have an overview of the cell type components in each sample.

- Issue: https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/782
- Discussion: https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/628

By applying the Seurat anchor transfer strategy, we could have an overview for the cell type composition, which is helpful to guide our next steps.

Seurat anchor transfer workflow (`FindTransferAnchors` and `TransferData`) was applied to pre-processed Seurat objects. The reference database I used here was from [Human Kidney Atlas (fetal)](https://www.kidneycellatlas.org/fetal-kidney-global) .

Yes, regarding the same issue. As [cell-type-wilms-tumor-06](https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-wilms-tumor-06) module applied a similar approach but different workflows (`SCTransform` + `Azimuth`), it may be worthwhile to add `SCTransform` and/or `Azimuth` as functions to this module. However, since the rationale and reference database are same, I wouldn't expect major differences in the result. Any suggestions?

`s3://researcher-009160072044-us-east-2/cell-type-wilms-tumor-14/`

* Results are uploaded to `s3://researcher-009160072044-us-east-2/cell-type-wilms-tumor-14/results/01_anchor_transfer_seurat`
* Each sample would have 4 result files, 2 for cell type level annotation, and 2 for compartment level. 
* `.csv` files are label transfer result tables including cell ID, predicted cell type, along with predicted scores.
* `.pdf` files are visualization plots.
* Note that prediction score < 0.5 would be labeled as "Unknown" in this analysis.
```
.
├── [sample_id]_celltype.csv
├── [sample_id]_celltype.pdf
├── [sample_id]_compartment.csv
└── [sample_id]_compartment.pdf
```

- Since this dataset only contains 10 samples, I generated results for all 10 samples here. 
- At the compartment level, majority of the cells are labeled as fetal nephron, which is the only compartment found in several samples (`SCPCS000515`, `SCPCS000519`, `SCPCS000748?`).
- At the cell type level, `Cap mesenchyme`, `S Shaped body`, `Renal vesicle` and their proliferating counterparts took the major proportion. 
- With the current approach, we could separate stroma cells, endothelium (although very few) pretty well. We could separate part of immune cells. We could mostly separate proliferating cells from their counterparts. However, `Cap mesenchyme`, `S Shaped body`, `Renal vesicle` and other nephron cells are often mixed.

- Analysis could be executed on a virtual computer ([Standard-4XL](https://openscpca.readthedocs.io/en/latest/aws/lsfr/creating-vcs/)) via AWS Lightsail for Research.
- Required R packages checked in `renv.lock` file.

After getting these results, I have a few concerns about the next step:
- Cell types/quality may vary across samples. At some point we need to treat each sample differently (and manually?). Do you think it's fine to have sample-specific codes in this project?
- I had the plan to run inferCNV/CopyKat for tumor cell identification, which would require a reference level. However, for samples with only nephrons, it's hard to determine the reference/normal cell level. Do you think we should still pursue tumor cell annotation based on CNV?
- Since most Seurat clusters have mixed cell types, it may not be a good idea to perform cluster-based or cluster-level annotation (such as scType) on this dataset.
- I noticed that one sample (`SCPCS000747`) ended up with a weird UMAP and cell type annotation result. So I went back to the inital QC report provided by the portal, it turned out that this sample had a lower number of detected genes in general. Any suggestions on this sample?

<details>

<summary>miQC plot for SCPCS000747</summary>

image

</details>

Thank you so much in advance for any suggestions/feedback!

### Author checklists

_Check all those that apply._


analyses/cell-type-wilms-tumor-14/scripts/utils/01_anchor_transfer_seurat_functions.R
```diff
@@ -0,0 +1,131 @@
...
+  
+  p1 <- Seurat::DimPlot(sample_obj, reduction = "umap", group.by = "predicted.id", label = F, cols = color) +
+    ggtitle(paste0(sample))
+  p2 <- Seurat::DimPlot(sample_obj, reduction = "umap", label = T)
+  df <- sample_obj@meta.data %>%
+    dplyr::group_by(seurat_clusters, predicted.id) %>%
+    dplyr::count(name = "sum")
+  p3 <- ggplot(df, aes(x = seurat_clusters, y = sum, fill =  predicted.id)) +
+    geom_bar(width = 0.5, stat = "identity", position = "fill") +
+    scale_fill_manual(values = color)
```
jashapiro, line 108: I think you can do this more simply, as `ggplot::geom_bar` will already calculate this grouped count by default
```suggestion
  p3 <- ggplot(sample_obj@meta.data, aes(x = seurat_clusters, fill =  predicted.id)) +
    geom_bar(width = 0.5, position = "fill") +
    scale_fill_manual(values = color) +
    labs(y = "Proportion of cells")
```

jashapiro (commented):
Hi @JingxuanChen7, and thank you for this recent contribution! Looking at the overall process, it seems like you have a good set of procedures for this stage of the analysis. Before we get too much further though, I would like to propose some restructuring to allow both for easier running and testing of the module. 

The main change I would like to propose is to add a `run_cell-type-wilms.sh` script at the top level of the module, which would include all steps of the analysis as we add them.  That script might look like the following:

```bash
#!/bin/bash

# This script runs scripts associated with the the cell-type-wilms-tumor-14 module 

# set error options
set -euo pipefail

# Set the working directory to the directory of this file
cd "$(dirname "${BASH_SOURCE[0]}")"

scratch_dir="scratch"
results_dir="results"
data_dir="../../data/current/SCPCP000014"

meta_path="${data_dir}/single_cell_metadata.tsv"

scratch_anchor_dir="${scratch_dir}/01_anchor_transfer_seurat"
results_anchor_dir="${results_dir}/01_anchor_transfer_seurat"

# Download and process reference data
ref_h5ad="${scratch_dir}/Fetal_full_v3.h5ad" 
ref_seurat="${scratch_dir}/Fetal_full_v3.h5ad" 

ref_url="https://cellgeni.cog.sanger.ac.uk/kidneycellatlas/Fetal_full_v3.h5ad"
curl -o $ref_h5ad $ref_url

# This script does not exist yet, but I would make one or one like it
Rscript scripts/00_preprocess_reference.R --h5ad "$ref_path" --seurat "$ref_seurat"

# Preprocess data
Rscript scripts/00_preprocessing_rds.R

# Assign anchors (assumes adding named arguments)
Rscript scripts/01_anchor_transer_seurat.R \
  --reference "$ref_seurat" \
  --metadata "$meta_path" \
  --library SCPCL000846,SCPCL000847 
```

Note that I would suggest using named arguments to pass file paths into scripts, and I generally think this is a good idea generally; it is often easier to adjust the outer shell script to accommodate changes to paths rather than requiring rewriting the R scripts that are being called, since more than one would often need to be modified. 

(I did not include code to prevent redownloads and rerunning here; but those could be added if needed)

There are a few other advantages to the bash script approach:
1. It is sometimes easier to see a working example of the full workflow in script form than as a series of steps in the readme
2. We can test that everything works on some by running this one script on each pull request as part of a GitHub Action. I am going to add to this PR some changes to the workflow action to start that automatic testing, as I think we want to have that going sooner rather than later. Note that this will likely mean that you will see some warnings about failures for a bit: this is expected until we start get that part of the process running smoothly.
3. Splitting the steps to build a workflow becomes a bit easier, especially if each step includes references to the files it is using. 

### Your questions

I also want to address some of your questions, but I will note that some of these might be best served by posting a new [discussion thread](https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/new?category=modify-an-existing-analysis) for your next steps, as that will let us separate that discussion from the changes we want to make to the code in this PR. Hopefully that can also allow for broader discussion of your questions! In particular, I think the discussions of copy number and cluster-level annotation may be something that would benefit from such posts. I do have some thoughts there (e.g., are the Seurat default clusters ideal here?), but I think getting others involved would be worthwhile. 

Some of them, however, I will address here:
- I think it can be fine to have some sample-specific code here, though if we can we would like to find if there are any patterns we can apply so that we do not necessarily have separate procedures for each sample. But we do expect that there may well be some tweaks that apply to specific samples, or groups of samples. 
- If there is a sample that does not have many genes detected, it may be that this sample can't really be reliably cell typed due to the quality of the sequencing. 


### Other comments

I made a number of line-level comments, but given the structural changes that I am suggesting above, many of those may not be fully relevant. I would suggest starting with any restructuring, and then I can read through again to see which of the specific comments are still relevant. 

One other thought I had while reading though was that it might be worth separating the plotting from the anchor transfer step. I think there are likely to be a lot of changes to the plots, so it would be nice to have a separate function that does the plotting on a single Seurat object to allow easier iteration. That might end up called from within the anchor transfer function, but keeping the logic separate and reusable may have later benefits as well.

.github/workflows/run_cell-type-wilms-tumor-14.yml
```diff
@@ -47,16 +47,33 @@ jobs:
...
+          ./download-results.py --test-data --projects SCPCP000014 --modules doublet-detection
 
       - name: Run analysis module
         run: |
           cd ${MODULE_PATH}
-          # run module script(s) here
+          # run individual scripts for now, until we have a main script
+          Rscript scripts/00_preprocessing_rds.R
+          # anchor transfer currently fails with a small number of cells
+          # Rscript scripts/01_anchor_transfer_seurat.R
```
jashapiro, line 79: I updated this test workflow file to run the current versions of your scripts, but the anchor transfer script is currently failing with the following error:

```
Error in `map()`:
ℹ In index: 1.
Caused by error in `FindWeights()`:
! Number of anchor cells is less than k.weight. Consider lowering k.weight to less than 37 or increase k.anchor.
Backtrace:
     ▆
  1. ├─purrr::walk(...)
  2. │ └─purrr::map(.x, .f, ..., .progress = .progress)
  3. │   └─purrr:::map_("list", .x, .f, ..., .progress = .progress)
  4. │     ├─purrr:::with_indexed_errors(...)
  5. │     │ └─base::withCallingHandlers(...)
  6. │     ├─purrr:::call_with_cleanup(...)
  7. │     └─global .f(.x[[i]], ...)
  8. │       └─global run_anchorTrans(...)
  9. │         └─Seurat::TransferData(...)
 10. │           └─Seurat:::FindWeights(...)
 11. │             └─base::stop(...)
 12. └─base::.handleSimpleError(...)
 13.   └─purrr (local) h(simpleError(msg, call))
 14.     └─cli::cli_abort(...)
 15.       └─rlang::abort(...)
Execution halted
```

I suspect that the error here is due to the fact that our test data only has 100 cells, but it will take some time to test the best way to get around this error.

Separately, the current tests take a long time because of the need to download and process the large reference file. Do you know if there might be another small reference file that we could use for testing only? If not, we might consider creating a subset of the reference for testing. 

For the moment, I have commented out the anchor transfer step, but we will probably want to try to solve this as part of this PR

analyses/cell-type-wilms-tumor-14/scripts/utils/01_anchor_transfer_seurat_functions.R
```diff
@@ -0,0 +1,131 @@
...
+    download.file('https://cellgeni.cog.sanger.ac.uk/kidneycellatlas/Fetal_full_v3.h5ad', 
+                  destfile = path_h5ad, 
+                  method = "wget")
+  }
+  
+  sce <- zellkonverter::readH5AD(path_h5ad)
+  rownames(sce) <- SingleCellExperiment::rowData(sce)$ID
+  seurat_obj <- SeuratObject::CreateSeuratObject(counts = SingleCellExperiment::counts(sce),
+                                                 assay = "RNA",
+                                                 project = "kidneyatlas")
```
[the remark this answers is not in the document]
JingxuanChen7: Thanks for suggesting `schard::h5ad2seurat()` here. I will keep using `zellkonverter` in this specific case, but switch to `schard` when feature renaming is not needed next time.

.github/workflows/run_cell-type-wilms-tumor-14.yml
jaclyn-taroni: I have reproduced (and solved!) this locally, so I will add comments to relevant spots of files in a review shortly.

jaclyn-taroni (commented):
We encountered a similar problem in testing `cell-type-wilms-tumor-06`, but that module uses `RunAzimuth()`, which seemed to not respect certain arguments (or we didn't find the right combination). 

However, `TransferData()` appears to use `k.weight` as documented. So, we should be able to use a lower value when running on test data. I left inline comments about how I would approach that. I didn't test everything I'm suggesting locally, but I did make sure I could set `k.weight = 15` and run successfully on test data.

analyses/cell-type-wilms-tumor-14/scripts/01_anchor_transfer_seurat.R
```diff
@@ -0,0 +1,65 @@
...
+  # if specify the number of samples for test
+  nsample <- as.numeric(args[1])
+  path_repo <- rprojroot::find_root(rprojroot::is_git_root)
+} else if (length(args) == 2) {
+  # if specify the number of samples & followd by repo path 
+  nsample <- as.numeric(args[1])
+  path_repo <- args[2]
+}else {
+  stop("Usage: Rscript 01_anchor_transfer_seurat.R <optional_num_of_test_samples> <opt_path_repo>", call.=FALSE)
+}
```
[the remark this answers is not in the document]
JingxuanChen7: Thanks both for suggesting `optparse`! I have applied this modification in my latest commit https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/784/commits/cbc5e89c83300755b917365b733941d9d88d20b5

analyses/cell-type-wilms-tumor-14/renv.lock
jaclyn-taroni, line 1: Hi @JingxuanChen7, I updated the workflow file now that we should be able to test the `01_anchor_transfer_seurat.R` script in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/784/commits/a9717eed51bbcfd85b543d6f2d018aa9bd0c5a58. 

The most recent failure is because `optparse` is not installed: https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/11131324686/job/30932638269

Can you rerun `renv::snapshot()` and commit the results please?

JingxuanChen7: My bad! I have updated `renv.lock` in my latest commit.

.github/workflows/run_cell-type-wilms-tumor-14.yml
```diff
@@ -47,16 +47,33 @@ jobs:
...
+          ./download-results.py --test-data --projects SCPCP000014 --modules doublet-detection
 
       - name: Run analysis module
         run: |
           cd ${MODULE_PATH}
-          # run module script(s) here
+          # run individual scripts for now, until we have a main script
+          Rscript scripts/00_preprocessing_rds.R
+          # anchor transfer
+          Rscript scripts/01_anchor_transfer_seurat.R --testing
```
jashapiro, line 79: Updating to use the main module script that you wrote.
```suggestion
          # run module script
          bash run_cell-type-wilms-14.sh
```

.github/workflows/run_cell-type-wilms-tumor-14.yml
```diff
@@ -47,16 +47,31 @@ jobs:
...
+        run: |
+          ./download-data.py --test-data --projects SCPCP000014 --format SCE
+          ./download-results.py --test-data --projects SCPCP000014 --modules doublet-detection
 
       - name: Run analysis module
         run: |
           cd ${MODULE_PATH}
-          # run module script(s) here
+          # run module script
+          bash run_cell-type-wilms-14.sh
```
jashapiro, line 77: Add an environment variable for testing
```suggestion
          CI_TESTING=1 bash run_cell-type-wilms-14.sh
```

analyses/cell-type-wilms-tumor-14/run_cell-type-wilms-14.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+# This script runs scripts associated with the the cell-type-wilms-tumor-14 module 
+
+# set error options
+set -euo pipefail
+
+# Set the working directory to the directory of this file
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
```
jashapiro, line 10: This will read the testing variable from the environment. 
```suggestion
# set CI variable if unset
CI_TESTING=${CI_TESTING:-0}

```

analyses/cell-type-wilms-tumor-14/run_cell-type-wilms-14.sh
```diff
@@ -0,0 +1,43 @@
...
+    --out_fetal_atlas "${ref_seurat}"
+
+## Preprocess data
+Rscript scripts/00_preprocessing_rds.R
+
+## Assign anchors 
+Rscript scripts/01_anchor_transfer_seurat.R \
+  --reference "${ref_seurat}" \
+  --metadata "${meta_path}" \
+  --samples SCPCS000514,SCPCS000515
```
jashapiro, line 43: Use the environment variable to selectively set the testing flag for the script
```suggestion
if [ "$CI_TESTING" -eq 0 ]; then
  TEST_FLAG=""
else
  TEST_FLAG="--testing"
fi

Rscript scripts/01_anchor_transfer_seurat.R \
  --reference "${ref_seurat}" \
  --metadata "${meta_path}" \
  --samples SCPCS000514,SCPCS000515 \
  $TEST_FLAG

```

jashapiro (commented):
In these three comments I am adding a bit of logic so the bash script will use the testing flag that was set up in the test environment.

analyses/cell-type-wilms-tumor-14/scripts/utils/01_anchor_transfer_seurat_functions.R
```diff
@@ -0,0 +1,107 @@
...
+                                                              TRUE ~ predicted.id))
+  sample_obj <- AddMetaData(object = sample_obj, metadata = predictions)
+  
+  # save prediction table
+  write.csv(predictions, file = filename_csv)
+  
+  # anchor transfer plots
+  plot_anchorTrans(path_anal = path_anal, 
+                   scratch_out_dir = scratch_out_dir, 
+                   results_out_dir = results_out_dir,
```
jashapiro, line 57: Do you maybe want to put these in the plots directory? I know that to do this you would need to add a line or lines to `.gitignore`, as they are too large to commit, but that might be worth doing for organization.

jashapiro (commented):
Thank you for your updates @JingxuanChen7; it is nice to have this running in the CI system we have set up, which will be a nice check going forward. 

Thank you for breaking out the plotting code; I think that will make it easier to read and modify. At the moment, I have not made any real comments on the plots, as I think those are mostly diagnostic, so the precise formatting might not matter.  

I made a few minor suggestions at this point, but the one I would like to get to before this goes in is to transition from sample ids to library ids for the unit of analysis. While for these samples there is a 1:1 relationship, that will not always be the case, so the library id is preferred for this kind of analysis. 

After those small changes and the line comments below, I think this can be nearly ready to be merged in. 

My other main comment, which I think I made earlier, is that it might be worth looking next (not in this PR) more into the relationship between clustering and cell types with different clustering methods. The default for Seurat appears fairly coarse, and it might be worth exploring whether other clustering methods might match the cell typing more closely. There is of course some potential circularity there, which we would want to be careful of, but in general I would not be too concerned that the current clusters might not match the cell type results well, and that there are some unseparated clusters. Clustering is quite difficult, and cell typing is often looking at only a subset of the data. You might want to do some more quantitative calculations about the clusters, such as looking at the purity with respect to the cell types.

I will also note that we are working on some tools for evaluating clustering at the moment, but they may not be quite ready in time for your use.

analyses/cell-type-wilms-tumor-14/scripts/utils/01_anchor_transfer_seurat_functions.R
```diff
@@ -0,0 +1,107 @@
...
+  #                            compartment == "immune" ~ "immune",
+  #                            compartment == "endothelium" ~ "endothelium",
+  #                            TRUE ~ celltype))
+  if (level == "compartment") {
+    ref_obj@meta.data <- ref_obj@meta.data %>%
+      mutate(annot = compartment)
+  } else {
+    ref_obj@meta.data <- ref_obj@meta.data %>%
+      mutate(annot = celltype)
+  }
```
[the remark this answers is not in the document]
JingxuanChen7: I had to change back to my original version, since `compartment` and `celltype` columns are stored as factor.

JingxuanChen7:
Hi @jashapiro , thank you so much for all the comments & efforts on setting up the CI system!

> After those small changes and the line comments below, I think this can be nearly ready to be merged in.
I have applied all the changes, including converting to library ID instead of sample ID, in my latest commit.

> My other main comment, which I think I made earlier, is that it might be worth looking next (not in this PR) more into the relationship between clustering and cell types with different clustering methods. The default for Seurat appears fairly coarse, and it might be worth exploring whether other clustering methods might match the cell typing more closely. There is of course some potential circularity there, which we would want to be careful of, but in general I would not be too concerned that the current clusters might not match the cell type results well, and that there are some unseparated clusters. Clustering is quite difficult, and cell typing is often looking at only a subset of the data. 

I totally agree with you that the default Seurat clustering workflow is quite rough. My anchor transfer results are not dependent on the clustering, so I think I can work on improving clustering in parallel. 

My preliminary analysis before has already shown simply modifying parameters for the Seurat workflow would not make a great difference on clustering. Therefore, I can try external tools for feature selection, clustering algorithm, and/or dimensional reduction. My colleague was able to get better clustering using SAM algorithm. I could probably try that as well? 

> You might want to do some more quantitative calculations about the clusters, such as looking at the purity with respect to the cell types.

I did some simple bar plots in this PR to check proportion of cell types within each cluster. But I will look into some other more robust ways to quantify cluster purity later.

Thanks again for all the help!

analyses/cell-type-wilms-tumor-14/scripts/01_anchor_transfer_seurat.R
```diff
@@ -17,10 +17,10 @@ option_list <- list(
     help = "Path to cohort metadata"
   ),
   make_option(
-    opt_str = c("--samples"),
+    opt_str = c("--libraries"),
     type = "character",
     default = "all",
-    help = "list of samples to run (starts with ), seperated by comma. Default is processing all samples in this cohort."
+    help = "list of libraries to run (starts with ), seperated by comma. Default is processing all libraries in this cohort."
```
jashapiro, line 23:
```suggestion
    help = "list of libraries to run, comma separated. Default is processing all libraries in this cohort."
```

analyses/cell-type-wilms-tumor-14/run_cell-type-wilms-14.sh
```diff
@@ -0,0 +1,64 @@
...
+  TEST_FLAG=""
+else
+  TEST_FLAG="--testing"
+fi
+
+Rscript scripts/01_anchor_transfer_seurat.R \
+  --reference "${ref_seurat}" \
+  --metadata "${meta_path}" \
+  --samples SCPCS000514,SCPCS000515 \
+  $TEST_FLAG
```
jashapiro, line 52: I think this should have been removed? (It is what is causing the failure of the CI testing, I think)
```suggestion
```

analyses/cell-type-wilms-tumor-14/scripts/00_preprocess_reference.R
```diff
@@ -0,0 +1,86 @@
...
+  seurat_obj <- Seurat::RunUMAP(seurat_obj, dims = 1:ndims)
+  # Seurat::DimPlot(seurat_obj, reduction = "umap", label = T, group.by = "compartment")
+  SeuratObject::SaveSeuratRds(seurat_obj, file = out_fetal_atlas)
+  
+}
+
+########### Prepare reference seurat obj ##########
+prepare_fetal_atlas(in_fetal_atlas = opt$in_fetal_atlas,
+                    out_fetal_atlas = opt$out_fetal_atlas,
+                    use_exist = T)
```
jashapiro, line 85:
```suggestion
                    use_exist = TRUE)
```

analyses/cell-type-wilms-tumor-14/scripts/00_preprocess_reference.R
```diff
@@ -0,0 +1,86 @@
...
+########### function for reference preparation ##########
+library(dplyr)
+library(Seurat)
+library(ggpubr)
+library(zellkonverter)
+library(SingleCellExperiment)
+
+prepare_fetal_atlas <- function(in_fetal_atlas = in_fetal_atlas,
+                                out_fetal_atlas = out_fetal_atlas,
+                                use_exist = T){
```
jashapiro, line 38:
```suggestion
                                use_exist = TRUE){
```

analyses/cell-type-wilms-tumor-14/.gitignore
```diff
@@ -5,3 +5,6 @@
 # Ignore the scratch directory (but keep it present)
 /scratch/*
 !/scratch/.gitkeep
+
+# Plots for certain steps should not be committed
+/plots/01_anchor_transfer_seurat/*
```
jashapiro, line 10: Only exclude pdfs to allow other plots to be included
```suggestion
/plots/01_anchor_transfer_seurat/**.pdf
```

jashapiro (commented):
Thank you for these updates. I think this is just about ready to go in. I had a few minor comments at the moment, one of which will hopefully fix the current CI failure. 

After talking with my colleague, I think we also want to try to include at least some of the results plots that you have made in the repository itself. The PDFs that you are making are a bit large, but it might be worth choosing some of the subplots to output for inclusion in the repository and outputting those as PNGs. Perhaps just the first page of the pdfs?

> My preliminary analysis before has already shown simply modifying parameters for the Seurat workflow would not make a great difference on clustering. Therefore, I can try external tools for feature selection, clustering algorithm, and/or dimensional reduction. My colleague was able to get better clustering using SAM algorithm. I could probably try that as well?

I agree that exploring the clustering can be done in parallel. We have found that the parameters of the leiden algorithm can make significant differences, especially modifications of the `objective_function` parameter used by `igraph`. We tend to implement clustering with the `bluster` package, as it provides a bit more flexibility and transparency; to use it you first extract the PCA (or other) matrix that you want to use for the clustering, then use the `clusterRows()` function: Some code using that approach can be found in a branch of this repo: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/feature/ropenscpca/packages/rOpenScPCA/R/calculate-clusters.R

> I did some simple bar plots in this PR to check proportion of cell types within each cluster. But I will look into some other more robust ways to quantify cluster purity later.

Yes, these bar plots were helpful! But if you were to evaluate clustering more, I would expect you would want to use a measure that was easier to compare across methods and parameters.

analyses/cell-type-wilms-tumor-14/run_cell-type-wilms-14.sh
JingxuanChen7: sorry for that. I forgot to pull before making modifications during last commit, and didn't notice this outdated command when resolving conflicts.

analyses/cell-type-wilms-tumor-14/scripts/utils/01_anchor_transfer_seurat_functions.R
```diff
@@ -106,4 +107,10 @@ plot_anchorTrans <- function(path_anal, scratch_out_dir, results_out_dir, plots_
                    filename = filename,
                    width = ifelse(length( pred_ids ) < 15,9,12), 
                    height = ifelse(length( pred_ids ) < 15,6,9))
+  
+  # save pngs
+  ggsave(p, file = filename_core,
+         device = "png", dpi = 72,
```
jashapiro, line 113: 72 dpi is pretty low these days, even for just on screen viewing. We could stay with the default 300dpi, but 200 is probably a good compromise to shrink the files a bit.
```suggestion
         device = "png", dpi = 200,
```

jashapiro (approved):
Thank you for making these updates! 

I wanted to make one more comment as you continue, which returns to the question of individual processing: It does seem that SCPCL001109 is behaving poorly, which is actually consistent with the fact that this library had a low number of detected genes per cell. For that reason alone I am not surprised to learn that the anchor transfer methods are not successful, but I am also not surprised that the UMAPs looks strange. I would expect that this sample could be filtered out and be left with unassigned cell types due to the low quality data.

Interestingly enough, our included UMAP looks more "standard", so it may partly depend on the normalization procedure. I don't think that is a reason on its own to adjust anything, as I don't think this is a good library overall, but it does illustrate some of the effects that early preprocessing can have.

PR #792, opened 2024-10-07
Add PNGs to large file exclusion

jaclyn-taroni:

Adding PNGs to the file extensions that get checked with the larger of the two file size pre-commit checks. This is because we have large PNGs we would like to add to the repository.


jashapiro (approved):
lgtm

PR #795, opened 2024-10-07
Merge rOpenScPCA into main

sjspielman:
This PR merges rOpenScPCA into `main`. 

I also made two changes in to the table in `packages/README.md` (let me know if you prefer I do that directly in `feature/ropenscpca` first? I figure if/when we use that branch again it would need to be synced with `main` anyways, so just did it here.).
- Change `=` -> `-` Since apparently github markdown doesn't like `=` for table rendering (see here: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/feature/ropenscpca/packages/README.md)
- Pop in the package version


jashapiro (approved):
LGTM, with the only comment being that I assume the docs failure is transient.

Issue #796, opened 2024-10-07
Add notebook to demonstrate clustering with rOpenScPCA


sjspielman:
Related to #791 

This issue tracks adding a notebook in the `hello-clusters` module that will perform clustering with `rOpenScPCA::calculate_clusters()`. The notebook should explain the general usage of this function and its arguments, demonstrate its use with a couple different parameter settings, and show how to add the cluster results into an SCE and/or Seurat object.

Edit: As part of this, also activate the `run_hello-clusters.yml` workflow to use the Docker image, which will have been pushed.

sjspielman:
Now that I'm thinking a bit more, I think it might make sense to also show clustering evaluation here with the functions:
- `calculate_silhouette()`
- `calculate_purity()`
- `calculate_stability()`

Having this in the same notebook will emphasize that you shouldn't just calculate clusters - you should check them clusters! Again, the concepts behind these evaluation metrics will be explained, and some associated viz for the metrics.

sjspielman:
While here, fix some spelling: https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/809#issuecomment-2413672442

PR #805, opened 2024-10-09
Initiate hello-clusters

### LINKED ISSUE #793: Establish module to demonstate clustering with rOpenScPCA
Related to #791 

This issue tracks establishing an analysis module (which should use `renv`) which will be used to show the clustering functionality in `rOpenScPCA`. Notably, before establishing the module, the `feature/ropenscpca` branch should be merged into `main`.

sjspielman:
Closes #793 

This PR establishes the `hello-clusters` module:

- I didn't include the analysis template directories (plots, scratch, & results) since I don't think they are necessary for this module, which I anticipate will just have some example notebooks at the top level. Do you foresee something I don't?
- I added an initial description to the README, including a brief overview of the functions this module will present and how to install `rOpenScPCA`. As notebooks get added, the README will be expanded accordingly.
  - I put in an instruction for `force=TRUE`, but I don't think I like that I did that? I imagine when there are relevant-enough changes we'll bump the package version itself and `renv::update()` maybe should be used instead?
- I added an initial (mostly empty) notebook which will get filled in as part of #796. I added this to be able to load libraries that allowed me to build up the `renv.lock` file. I also added an, essentially empty for now, script to run the module which will likely just be used for testing 
- The Dockerfile is ready to go as well, but I didn't activate the build docker workflow yet. But, if we want to actually just get the run module workflow up and running with a Docker environment from the get-go, maybe now is actually the time so the next PRs can use it!

### DIFF
--- a/.github/workflows/docker_hello-clusters.yml
+++ b/.github/workflows/docker_hello-clusters.yml
@@ -13,22 +13,22 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  # pull_request:
-  #   branches:
-  #     - main
-  #   paths:
-  #     - "analyses/hello-clusters/Dockerfile"
-  #     - "analyses/hello-clusters/.dockerignore"
-  #     - "analyses/hello-clusters/renv.lock"
-  #     - "analyses/hello-clusters/conda-lock.yml"
-  # push:
-  #   branches:
-  #     - main
-  #   paths:
-  #     - "analyses/hello-clusters/Dockerfile"
-  #     - "analyses/hello-clusters/.dockerignore"
-  #     - "analyses/hello-clusters/renv.lock"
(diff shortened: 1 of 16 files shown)


jashapiro (approved):
This looks good for the start of the repo. One typo fix I saw, and the comments below (tl;dr: do add back the standard directories).

I had some interesting trouble installing at first with your directions and a slightly nonstandard renv setup, but I think the fix for that can go elsewhere. 

> * I didn't include the analysis template directories (plots, scratch, & results) since I don't think they are necessary for this module, which I anticipate will just have some example notebooks at the top level. Do you foresee something I don't?

I would include these, as they don't hurt at all. You can also have the notebooks save some results (multiple clustering results as a table) and plots. I think it is good to demonstrate this even in a simple module like this.

Issue #810, opened 2024-10-15
Update cell-type-wilms-tumor-06 to not use Azimuth for label transfer


sjspielman:
Currently, much of the `cell-type-wilms-tumor-06` module is not run in CI due to issues we experienced getting the [`Azimuth::RunAzimuth()`](https://satijalab.github.io/azimuth/reference/RunAzimuth.html) function to run on our test data. This issue tracks updating the label transfer steps in this module to no longer use this function directly. Instead, we've "reverse-engineered" `Azimuth::RunAzimuth()`so that we can use its functionality without actually using the function directly.

For context, there are two references which are used for label transfer:
- The `fetusref` reference (Cao). Currently, this reference is directly obtained from Azimuth, and then used for label transfer, in `notebook_template/02a_label-transfer_fetal_full_reference_Cao.Rmd`.
- The fetal kidney reference (Stewart). Currently, this reference is prepared for use with Azimuth in `scripts/download-and-create-fetal-kidney-ref.R` and used in  `notebook_template/02b_label-transfer_fetal_full_reference_Stewart..Rmd`.

Here's how we will update the code (and associated documentation):
- `scripts/download-and-create-fetal-kidney-ref.R` will be updated to _also_ prepare the `fetusref` reference for use in label transfer, borrowing some strategies from `Azimuth::RunAzimuth()`. The prepared reference will be exported as an RDS file to ~`scratch`~ `results` for subsequent use in `notebook_template/02a_label-transfer_fetal_full_reference_Cao.Rmd`. This object will be a list containing a couple different fields of information that can be pulled out when it comes time for label transfer.
- A new utils file will be added to `notebook_template` with functions used for label transfer. These functions contain code derived from `Azimuth::RunAzimuth()` .There will be two functions:
	- `prepare_query()` will do a bit of formatting on Seurat objects to run label transfer on
	- `transfer_labels()` will perform the label transferring directly 
- The code in `notebook_template/02a_label-transfer_fetal_full_reference_Cao.Rmd` and `notebook_template/02b_label-transfer_fetal_full_reference_Stewart.Rmd` will be updated to use these new functions to perform the label transfer. The goal is to modify as little code as possible to achieve the new functionality. We will also need to add Rmd params to indicate whether we're running in CI to set certain parameters appropriately (e.g., a small `k.weight`  value will be needed for test data in CI.) 
  - As part of this, we'll want to manually look at some results to ensure that changes are minimal-to-none 🤞  
[...]

PR #814, opened 2024-10-15
Wilms 06 update: script to prepare fetal references

sjspielman:
Towards #810 

This PR takes the first step of transitioning the label transfer code in the `cell-type-wilms-tumor-06` module. Here, I replaced the script `download-and-create-fetal-kidney-ref.R` with the new script `prepare-fetal-references.R`, which contains a function that does more things to a given Azimuth reference to get all information we'll need in place for label transfer. This script does three main things:
- Prepares and exports the fetal kidney reference for label transfer. This is reference the script originally prepared for Azimuth. Now, it still prepares it for Azimuth, but then also further tweaks it for our adapted label transfer approach.
- Prepares and exports the full fetus reference for label transfer. 
- Downloads a `homologs.rds` file from Azimuth which is used during label transfer as part of gene mapping

Locally, I tested the prepared references to ensure they will work for next steps, and they seem good!

I also updated relevant documentation throughout the module, as well as the module's run workflow script. There are a good amount of space changes here since the pre-commit hook had not previously styled these files, so I recommend *turning off whitespace during review*! 

Finally, I updated the workflow for this module so that it will run on PRs to `feature/*` branches, as `feature/wilms-tumor-06-azimuth` is our base branch for this endeavor.


analyses/cell-type-wilms-tumor-06/scripts/prepare-fetal-references.R
```diff
@@ -0,0 +1,215 @@
...
+# Two references are prepared:
+# - A fetal kidney atlas (Stewart) obtained from CELLxGENE. This is saved to <output_dir>/stewart_formatted_ref.rds
+# - A fetal organ reference (Cao) obtained from Azimuth. This is saved to <output_dir>/cao_formatted_ref.rds
+#
+# The script also downloads and saves an ID mapping file from Azimuth which will be used to convert ensembl IDs to
+#  gene names during label transfer. This is saved to <output_dir>/homologs.rds
+#
+# USAGE:
+# Rscript prepare-fetal-references.R \
+#   --kidney_url https://datasets.cellxgene.cziscience.com/40ebb8e4-1a25-4a33-b8ff-02d1156e4e9b.rds \
```
jashapiro, line 16: For a few reasons (transparency, easier translation to a workflow, etc.), I think it would be better to separate downloads out of this script and instead have the downloads done in a wrapper script (that could also check if the downloads have already happened).

analyses/cell-type-wilms-tumor-06/scripts/prepare-fetal-references.R
```diff
@@ -0,0 +1,215 @@
...
+
+# format for label transfer
+cao_ref_list <- prepare_azimuth_reference(fetus_ref, cao_annotation_levels)
+
+# export formatted reference
+saveRDS(cao_ref_list, cao_ref_file)
+
+
+# Finally, download the ID mapping file from Seurat ---------------------
+download.file(url = opts$homologs_url, destfile = homologs_file)
```
jashapiro, line 215: Without knowing exactly how this will be used, I'm not sure we want to do this? We already have all of the relevant mappings stored within the SCE objects, and I don't think we know what versions this particular file is/will be based on, so it might well be mismatched. If we do need this particular structure, we can generate it?

Also, for transparency and easier later updates, I would not perform this step within this script, but as an external step.

jashapiro (commented):
I didn't have a full look yet, but I had a couple initial thoughts...

jaclyn-taroni (approved):
Looks good, and I don't need to see this again, but I do want to see how the output directory is handled in `prepare-fetal-references.R` get updated before this goes in.

analyses/cell-type-wilms-tumor-06/scripts/prepare-fetal-references.R
sjspielman: We do need this particular structure - this is specifically adapted from Azimuth's code which downloads/uses this file directly. But, indeed, no idea how versioning will play out, and..yeah.

We could generate something like this on our own for mapping, though:
```
> head(h)
  Gene.stable.ID.human Transcript.stable.ID.human Gene.name.human
1      ENSG00000204479            ENST00000376098        PRAMEF17
2      ENSG00000204479            ENST00000376098        PRAMEF17
3      ENSG00000204479            ENST00000376098        PRAMEF17
4      ENSG00000204479            ENST00000376098        PRAMEF17
5      ENSG00000204479            ENST00000376098        PRAMEF17
6      ENSG00000204479            ENST00000376098        PRAMEF17
  Transcript.name.human Gene.stable.ID.mouse Transcript.stable.ID.mouse
1          PRAMEF17-201   ENSMUSG00000094741         ENSMUST00000178291
2          PRAMEF17-201   ENSMUSG00000074720         ENSMUST00000099245
3          PRAMEF17-201   ENSMUSG00000094337         ENSMUST00000179140
4          PRAMEF17-201   ENSMUSG00000094337         ENSMUST00000164489
5          PRAMEF17-201   ENSMUSG00000095570         ENSMUST00000179972
6          PRAMEF17-201   ENSMUSG00000096728         ENSMUST00000178296
  Gene.name.mouse Transcript.name.mouse
1      AC140325.3        AC140325.3-201
2      AC140325.1        AC140325.1-201
3      AC140325.2        AC140325.2-201
4      AC140325.2        AC140325.2-202
5      AC140325.4        AC140325.4-201
6      AC140325.5        AC140325.5-201
```

(FWIW, I have all the code written on my end for label transfer working on my end, and it's slowing debuting in PRs 😄 )

jashapiro: We can wait to resolve this until the file is being used. I feel like we should be able to bypass Azimuth's conversion by converting the feature names from `gene_id` to `gene_symbol` using the `rowData` before/during conversion to Seurat objects, which will allow us to be explicit and consistent.

sjspielman: Noting from separate discussion that this is definitely a "nice to have" for the module, but not an urgent change to do now since we have some deadlines approaching. We may circle back to this in the future!

analyses/cell-type-wilms-tumor-06/00_run_workflow.R
```diff
@@ -35,6 +35,7 @@ metadata <- read.table(sample_metadata_file, sep = "\t", header = TRUE)
 module_base <- file.path(root_dir, "analyses", "cell-type-wilms-tumor-06")
 
 # Download and create the references for label transfer, and download the homolog file for ID mapping ----------
+system(command = glue::glue("Rscript ", file.path(module_base, "scripts", "download-reference-files.R")))
```
jashapiro, line 38: If you do the downloads here rather than in a separate script, you can include the file paths as the arguments to the later commands, making any later changes easier. 

```suggestion
# The fetal kidney atlas (Stewart) RDS file from CELLxGENE.
kidney_url <- "https://datasets.cellxgene.cziscience.com/40ebb8e4-1a25-4a33-b8ff-02d1156e4e9b.rds"
kidney_file <- file.path(root_dir, "scratch", "fetal_kidney.rds")
if (!file.exists(kidney_file)) {
  download.file(url = kidney_url, destfile = kidney_file)
}

# The homologs.rds file from Azimuth
homologs_url <- "https://seurat.nygenome.org/azimuth/references/homologs.rds"
homologs_file <- file.path( root_dir, "scratch", "homologs.rds")
if (!file.exists(homologs_file)) {
  download.file(url = homologs_url, destfile = homologs_file)
}
```

sjspielman:
So, this PR began failing CI after it passed because I accidentally changed a file name (dealt with in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/814/commits/63830e6c470e93ad63004d829c2a76ae2d234a7c; noting this diff is big because the Rmd got styled so it's mostly whitespace). While addressing this bug, I came across a different bug that did _not_ previously trigger a CI failure but should have. There was an error preparing the built-in Azimuth reference, because I did not run `InstallData` before running `LoadData`. Even though that script being called failed, the overall workflow script did not, and hence CI did not fail; it just silently threw an error for that line (`system(Rscript prepare-kidney-references.R)`) and kept going onto the next line, which is not great for CI testing! So we've learned something about using `system()`...

Either we'll want to change `00_run_workflow.R` somehow. We could...
- switch all the `system()` calls to instead be `source()`. That said, while I know `source()` would address this issue, that's not something that plays great with optparse. This option might be DOA.
- figure out a flag that can be used with either `system` or  `Rscript` to cause the whole script to die in the event of any error; that said, I can't find such a flag/argument, so this too might be DOA....
- change the script to be a shell script 
- another idea??

I think we probably want to do this sooner rather than later to prevent bugs from piling up without our knowledge, so I'd either tack it into this PR or handle in a different PR before merging in this one. CC @jashapiro @jaclyn-taroni

jashapiro:
> So, this PR began failing CI after it passed because I accidentally changed a file name (dealt with in [63830e6](https://github.com/AlexsLemonade/OpenScPCA-analysis/commit/63830e6c470e93ad63004d829c2a76ae2d234a7c); noting this diff is big because the Rmd got styled so it's mostly whitespace). While addressing this bug, I came across a different bug that did _not_ previously trigger a CI failure but should have. There was an error preparing the built-in Azimuth reference, because I did not run `InstallData` before running `LoadData`. Even though that script being called failed, the overall workflow script did not, and hence CI did not fail; it just silently threw an error for that line (`system(Rscript prepare-kidney-references.R)`) and kept going onto the next line, which is not great for CI testing! So we've learned something about using `system()`...
> 
> Either we'll want to change `00_run_workflow.R` somehow. We could...
> 
> * switch all the `system()` calls to instead be `source()`. That said, while I know `source()` would address this issue, that's not something that plays great with optparse. This option might be DOA.
> * figure out a flag that can be used with either `system` or  `Rscript` to cause the whole script to die in the event of any error; that said, I can't find such a flag/argument, so this too might be DOA....
> * change the script to be a shell script
> * another idea??
> 
> I think we probably want to do this sooner rather than later to prevent bugs from piling up without our knowledge, so I'd either tack it into this PR or handle in a different PR before merging in this one. CC @jashapiro @jaclyn-taroni

`system()` returns the exit code, so you can do something to capture that and fail if it isn't zero, but you have to do that with every call, so you'd probably want a little wrapper function like:
```
system_check <- function(command){
  return_code <- system(command)
  if(return_code != 0){stop("Failed command: ", command)}
}
```

The better solution is probably to convert to a shell script, though, as that will be more portable and easily adapted.

sjspielman:
> The better solution is probably to convert to a shell script, though, as that will be more portable and easily adapted.

This was done in #817 and has been merged in here.

sjspielman:
@jashapiro this is ready for another look now - specifically the updates to `00_run_workflow.sh`. The pre-commit check is hanging presumably because of other changes that happened in parallel, so perhaps we just need to get #823 in first and update this branch accordingly

analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
```diff
@@ -22,15 +22,30 @@ data_dir="../../data/current"
...
+# Create the fetal references for label transfer (Stewart et al and Cao et al)
+Rscript scripts/prepare-fetal-references.R --kidney_ref_file "${kidney_ref_file}"
 
 # Characterize the fetal kidney reference (Stewart et al.)
 Rscript -e "rmarkdown::render('${notebook_template_dir}/00b_characterize_fetal_kidney_reference_Stewart.Rmd',
     output_format = 'html_document',
     output_file = '00b_characterization_fetal_kidney_reference_Stewart.html',
-    output_dir = '${notebook_output_dir}/00-reference')"
+    output_dir = '${notebook_output_dir}/00-reference',
+    params = list(fetal_kidney_path = '../${kidney_ref_file}'))" # path relative to notebook
```
jashapiro, line 48: Maybe a bit less brittle? Or you could define the `kidney_ref_file="${PWD}/scratch/fetal_kidney.rds"` at the start?
```suggestion
    params = list(fetal_kidney_path = '${PWD}/${kidney_ref_file}'))" # absolute path
```

jashapiro (approved):
Once tests are passing with merging in the precommit workflow, this should be fine.

Issue #816, opened 2024-10-16
Use shell script for cell-type-wilms-tumor-06 workflow


sjspielman:
This issue is based on this discussion: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/814#issuecomment-2415199088

Briefly, we have discovered that calling `system()` from within an Rscript is actually not a robust strategy for running the module workflow, since individual steps that fail don't cause the entire workflow to error. Therefore, it would be best to instead use a shell script to run the workflow, which for consistency we'll call `00_run_workflow.sh`.

I'm also opening this issue so @maud-p is aware - for now, you can feel free to please continue working with your existing `00_run_workflow.R` script just as you have been! While I'm working in parallel, I will not modify `00_run_workflow.R`. Once all code related to #810 is merged into `main`, you can file a PR to formally remove `00_run_workflow.R` (I can also help to remove this file, but I think it's best if you do it so that you have no confusion when it's time to stop using that script!). I will make sure to let you know when code has been merged and we're ready for that next step, and I can assist as needed!

Noting we should not close this issue out until the workflow is formally using the new shell script from `main`.

sjspielman:
With #883, this new script was merged into main, and the old R script was removed.

PR #817, opened 2024-10-16
Create shell script for wilms-06 workflow

sjspielman:
Towards #816 

This PR adds a new shell script to run the `cell-type-wilms-tumor-06` workflow. This does NOT contain any of the active changes in #814, since I wanted to make sure a directly comparable version of this script was in place before changing up the steps. As such, this is basically a "shell version" of `00_run_workflow.R`.

Other changes include:
- updating the workflow yaml file to use this script, and run on this branch
- updating the module README to reference this script instead of the old one. Note this file got got by precommit, hence whitespace diffs.

EDIT: Link to workflow R script for comparison during review: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/feature/wilms-tumor-06-azimuth/analyses/cell-type-wilms-tumor-06/00_run_workflow.R

### DIFF
--- a/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
+++ b/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
@@ -1,0 +1,105 @@
+#!/bin/bash
+
+# This script runs the module workflow.
+#
+# USAGE:
+# bash 00_run_workflow.sh
+#
+# USAGE in CI:
+# TESTING=1 bash 00_run_workflow.sh
+
+set -euo pipefail
+
+IS_CI=${TESTING:-0}
+project_id="SCPCP000006"
+
+# Ensure script is being run from its directory
+module_dir=$(dirname "${BASH_SOURCE[0]}")
+cd ${module_dir}
+
+# Define directories
+data_dir="../../data/current"
+notebook_template_dir="notebook_template"
+notebook_output_dir="notebook"
+
+
+# Download and create the fetal kidney reference (Stewart et al)
+Rscript scripts/download-and-create-fetal-kidney-ref.R
+
+# Build the gene position file reference for infercnv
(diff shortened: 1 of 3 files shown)


analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
```diff
@@ -0,0 +1,98 @@
...
+data_dir="../../data/current"
+notebook_template_dir="notebook_template"
+notebook_output_dir="notebook"
+
+
+# Download and create the fetal kidney reference (Stewart et al)
+Rscript scripts/download-and-create-fetal-kidney-ref.R
+
+# Build the gene position file reference for infercnv
+Rscript scripts/06a_build-geneposition.R
```
jashapiro, line 30: It is confusing to have this out of numerical order. Do we want to rename the file or move this step?

analyses/cell-type-wilms-tumor-06/README.md
```diff
@@ -137,11 +137,11 @@ We gather from the litterature marker genes and specific genomic alterations tha
 
 ## Clustering and label transfer from fetal references
 
-R Script to be rendered : `00_run_workflow.R`
+R Script to be rendered : `00_run_workflow.sh`
```
jashapiro, line 140:
```suggestion
Bash script to be run : `00_run_workflow.sh`
```

jashapiro (approved):
Looks good as a general translation of the R script.

analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
sjspielman: I agree. I moved it at one point, and then in the spirit of keeping things as they were, I unmoved it. But you see it too, so yes let's move it.

PR #823, opened 2024-10-16
Update the feature/wilms-tumor-azimuth-06 branch

sjspielman:
This PR updates the `feature/wilms-tumor-azimuth-06` branch with code from `main`. Along the way, I ran into an issue with `.gitleaks` getting a bit too eager.

It detected in the new notebooks (same line/issue for all 5 new ones):
```
Finding:     ...i--;break;case V.ui.keyCode.END:REDACTED;break;case V.ui.keyC...
Secret:      REDACTED
RuleID:      generic-api-key
Entropy:     3.882045
File:        analyses/cell-type-wilms-tumor-06/notebook/SCPCS000184/05_copykat_exploration_SCPCS000184.html
Line:        75
Fingerprint: analyses/cell-type-wilms-tumor-06/notebook/SCPCS000184/05_copykat_exploration_SCPCS000184.html:generic-api-key:75
```

Further investigation revealed this culprit (which I expected to be more interesting, frankly), where `=this.anchors.length-1` was REDACTED.
```
i--;break;case V.ui.keyCode.END:i=this.anchors.length-1;break;case V.ui.keyC
```

So, I basically just popped this pattern into the `.gitleaks.toml` file using `.+` instead of `REDACTED`. Should I modify this further? I wasn't sure how specific or general to make this, but hey it seems to work. Any feedback there appreciated! I attached a text file with L75 as needed.
[L75.txt](https://github.com/user-attachments/files/17400309/L75.txt)

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -5,4 +5,7 @@ useDefault = true
 [allowlist]
 # skip base64 encoded images, which might have substrings that look like tokens
 regexTarget = "line"
-regexes = ['''(?i)<img src="data:image\/.+;base64,.+?".+\/>''']
+regexes = [
+    '''(?i)<img src="data:image\/.+;base64,.+?".+\/>''',
+    '''i--;break;case V\.ui\.keyCode\.END:.+;break;case V\.ui\.keyC'''
+]
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -19,6 +19,7 @@ It is not meant to be run as a standalone action.
 
 ## Repository maintenance
 
+* `pre-commit.yml` runs pre-commit checks defined by the `.pre-commit-config.yaml` file.
 * `spellcheck.yml` performs spell checking of markdown and R Markdown documents.
 * `code-styling.yml` performs code styling using `styler` and `ruff`.
   * This workflow also consumes `modules_code-styling.txt` to determine which modules to style.
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,18 @@
+name: Pre-commit check
+on:
+  pull_request:
+    branches:
+      - main
+      - feature/*
+
+jobs:
+  pre-commit:
+    if: github.repository_owner == 'AlexsLemonade'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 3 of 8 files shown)


.gitleaks.toml
```diff
@@ -5,4 +5,7 @@ useDefault = true
 [allowlist]
 # skip base64 encoded images, which might have substrings that look like tokens
 regexTarget = "line"
-regexes = ['''(?i)<img src="data:image\/.+;base64,.+?".+\/>''']
+regexes = [
+    '''(?i)<img src="data:image\/.+;base64,.+?".+\/>''',
+    '''i--;break;case V\.ui\.keyCode\.END:.+;break;case V\.ui\.keyC'''
```
jashapiro, line 10: This line is the jQuery inclusion (which probably has changed between versions of rmarkdown, hence a new error), so to be a bit more general/robust to future changes that could look like secrets, we might want to match the start of the line with something like:

```suggestion
    # skip jQuery definition function
    '''^!function\(.+jQuery'''
```

I think the comment should be okay within the array, so I would move the comment on line 6 to preface the first regex.

.gitleaks.toml
```diff
@@ -5,4 +5,8 @@ useDefault = true
 [allowlist]
 # skip base64 encoded images, which might have substrings that look like tokens
```
jashapiro, line 6:
```suggestion
```

.gitleaks.toml
```diff
@@ -5,4 +5,8 @@ useDefault = true
...
 regexTarget = "line"
-regexes = ['''(?i)<img src="data:image\/.+;base64,.+?".+\/>''']
+regexes = [
+    '''(?i)<img src="data:image\/.+;base64,.+?".+\/>''',
```
jashapiro, line 9:
```suggestion
regexes = [
    # skip base64 encoded images, which might have substrings that look like tokens
    '''(?i)<img src="data:image\/.+;base64,.+?".+\/>''',
```

jashapiro (approved):
LGTM, with one change to comments. I think we can merge this without waiting for the full check to complete, but up to you.

sjspielman:
> I think we can merge this without waiting for the full check to complete, but up to you.

This workflow has been thoroughly checked, and the changes in this PR won't affect it. So I'm sure on board to merge without waiting for the full check.

Issue #822, opened 2024-10-16
cell type/tumor annotation for ETP T-ALL (SCPCP000003)


UTSouthwesternDSSR:
### Please link to the GitHub Discussion for this proposed analysis.

#629 

### Describe the goals of this analysis module.

The goal is to annotate cell type and tumor/normal status for each ETP T-ALL samples (n=31) in SCPCP000003. I followed the same analysis module, as proposed in non-ETP T-ALL (Issues: #761 and #787):

- using [SAM](https://github.com/atarashansky/self-assembling-manifold) algorithm to separate the cells
- [ScType](https://github.com/IanevskiAleksandr/sc-type) for cell type annotation using the same marker list
- [CopyKat](https://github.com/navinlabcode/copykat) for identification of malignant cells based on the profile of genome-wide aneuploidy

### What software will you require?

- R (4.2.3)
- Main packages: `Seurat`, `reticulate`, `SAM`, `ScType`, `CopyKat`

### What will your first pull request contain?

- Creating a module skeleton
`./create-analysis-module.py cell-type-ETP-ALL-03 --use-r  --use-renv --use-conda --conda-file-only`
- Adding analysis code for `00-01_processing_rds.R`, `02-03_annotation.R`, and `multipanel_plot.R`

### What computational resources will you require?

I am using a standard-4XL virtual machine via AWS Lightsail for Research.

### If known, when do you expect to file the first pull request?

~Today or tomorrow

PR #826, opened 2024-10-16
cell type/tumor annotation for ETP T-ALL (SCPCP000003)

UTSouthwesternDSSR:

To perform cell type/tumor annotation for ETP T-ALL samples (n=31) in SCPCP000003

#822 

To perform cell type/tumor annotation for ETP T-ALL samples (n=31) in SCPCP000003

The same approach is followed as proposed in the module for non-ETP T-ALL (SCPCP000003):

- using [SAM](https://github.com/atarashansky/self-assembling-manifold) algorithm to separate the cells
- [ScType](https://github.com/IanevskiAleksandr/sc-type) for cell type annotation using the same marker list
- [CopyKat](https://github.com/navinlabcode/copykat) for identification of malignant cells based on the profile of genome-wide aneuploidy

The only difference is that there are more than one cluster identified as `B cell` in 4 samples (`SCPCL000055`, `SCPCL000066`, `SCPCL000696`, and `SCPCL000709`). I check the location of B cells on the umap and also compare with the `BFeatures1` (average expression of B marker genes) on the dotplot. (I also check with the expression of adt_CD19 in these 4 samples. Higher expression is shown on the separated B island.) 
Screenshot 2024-10-16 at 12 35 15 PM

I believe that only those that are completely separated (an island on its own, rather than attached to the other clusters) can be confidently used as the normal cells for running CopyKat, although the results is not super promising (shown later).

- `rds` objects: `s3://researcher-650251722463-us-east-2/cell-type-ETP-ALL-03/results/rds`
- metadata and ScType results: `s3://researcher-650251722463-us-east-2/cell-type-ETP-ALL-03/results/`
- umap and dot plots: `s3://researcher-650251722463-us-east-2/cell-type-ETP-ALL-03/plots` 

- `rds` objects
- two text files for each sample: `_metadata.txt` (cell ID, leiden clusters, cell type annotation, low confidence cell type annotation, CopyKat prediction, and new CopyKat prediction based on the "selected" B cells [for the 4 samples]) and `_sctype_top10_celltypes_perCluster.txt` (top 10 possible cell types with their respective sctype score in each cluster)
- `multipanels_` umap plots showing leiden clustering, cell type, and copyKat prediction respectively (for the 4 samples, I am showing the new CopyKat prediction based on "selected" B cells).
- `dot` plots showing the average expression of group of markers for each cell type using `AddModuleScore()`

With the default threshold of having sctype score > 25% of ncells in a cluster (`sctype_classification`), there are a large number of cells being annotated as "Unknown" in each sample, ranging from 0 to 61%, with the median ~25%.
Screenshot 2024-10-16 at 2 19 05 PM
If we were to use 10% threshold (`lowConfidence_annot`), the percentage of `Unknown` is now capped at 28% (instead of 61%).

Every sample has B cells annotated. Thus, I ran CopyKat on all of them, but as mentioned above, I selected some B cells as the normal for these 4 samples (`SCPCL000055`, `SCPCL000066`, `SCPCL000696`, and `SCPCL000709`). Here are the comparison between using all B cells (`copykat.pred`) vs particular B cells (`new_copykat.pred`). The results seem to make sense for `SCPCL000055`, since there are now much more aneuploid cells, and Late Eryth has become diploid. It works for `SCPCL000066` too, but not really for `SCPCL000709`, as the number of `aneuploid` cells decrease. `SCPCL000696` shows very little changes. 
Screenshot 2024-10-16 at 2 28 23 PM
Screenshot 2024-10-16 at 2 32 31 PM

Overall, there are very few `not.defined` cells from CopyKat prediction results, capping at 6% of total cells in a sample.

- The packages are installed and updated in `renv.lock` and `conda.lock.`
- Analysis could be executed on a Standard-4XL virtual machine via AWS Lightsail for Research

- Please let me know what do you think about selecting some B cells for these 4 samples, instead of providing all B cells as the normal.
- I am not sure if you still want the CopyKat results like what we discussed in the non-ETP module, as the total size for their output is about 101GB, and it seems like we may try using inferCNV instead.

### Author checklists

_Check all those that apply._


jaclyn-taroni (approved):
Thanks for this contribution, @UTSouthwesternDSSR! Since this is similar to the non-ETP ALL module, we can merge it.

Having reviewed the results, the B cell assignments from ScType are reasonably convincing for *some libraries*. 

In terms of next steps, I would recommend picking one set of samples (non-ETP ALL *or* ETP ALL) to focus on and see if you can get the best quality B cell assignments possible (you might want to pull in the automatic assignments we already have from CellAssign and SingleR for comparison, too) and get inferCNV up and running.

Thank you again, and please let us know if there's anything you want to discuss!

UTSouthwesternDSSR:
Sure, I would start out with the non-ETP samples, and do what you have suggested (making sure that the B cells called are indeed solid, and then used them for running inferCNV) in the other pull request, and then test them with the ETP samples. Thank you so much for the suggestion!

PR #827, opened 2024-10-16
Update cell-type-wilms-tumor-06 workflow 

### LINKED ISSUE #824: Update cell-type-wilms-tumor-06 workflow file
We'd like to update this module's workflow in `run_ cell-type-wilms-tumor-06.yml` as follows:

- Start using the Docker image instead of building the environment with `renv` from scratch.
- Similar to how we limited the number of test datasets run through _on a PR trigger_ [in the `doublet-detection` module](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/a5c36235eff73530cf496c4d82715e33e79fc90e/.github/workflows/run_doublet-detection.yml#L44-L55), we'd like something similar in this module. For this, we can use a flag to specify which datasets to download; either the full project if the trigger is not PR, or a subset of samples if the trigger is a PR.

sjspielman:
Closes #824 

This PR updates the `cell-type-wilms-tumor-06` workflow file to:
- Use Docker (note that I did also confirm just to be sure that the system dependencies needed are indeed already in the Docker container, thanks bioc base image!)
- Use a subset of samples only when testing on PRs. I chose these samples as the ones that are specifically run through copyKAT & inferCNV, since the workflow currently evaluates these methods only on a subset of samples. This way, those samples will be present! There is nothing else particularly special about them.


sjspielman:
Ah, so this is the first workflow we've used Docker in where the container doesn't have a conda environment, so `aws` bells and whistles for downloading data aren't installed. I think the move is probably to add a step in the workflow to build the `openscpca` conda environment, vs adding it to Docker. @jashapiro, any other opinions on this?

jashapiro:
Yes, I would rather leave the docker image alone. I don't think you should need to install conda; just a step to install `aws` directly should be sufficient: 
```
curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
unzip awscliv2.zip
sudo ./aws/install
```

.github/workflows/run_cell-type-wilms-tumor-06.yml
```diff
@@ -34,32 +34,28 @@ jobs:
...
-        uses: r-lib/actions/setup-r@v2
-        with:
-          r-version: 4.4.0
-          use-public-rspm: true
-
-      - name: Set up pandoc
-        uses: r-lib/actions/setup-pandoc@v2
-
-      - name: Install system dependencies
+      - name: Install aws-cli to support the data download
```
jashapiro, line 46:
```suggestion
      - name: Install aws-cli
```

.github/workflows/run_cell-type-wilms-tumor-06.yml
```diff
@@ -34,32 +34,28 @@ jobs:
...
-          working-directory: ${{ env.MODULE_PATH }}
-
-      # Update this step as needed to download the desired data
-      - name: Download test data
-        run: ./download-data.py --test-data --format SCE --projects SCPCP000006
+          curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+          unzip awscliv2.zip
+          sudo ./aws/install
+
+      - name: Download test data in SCE format
```
jashapiro, line 52: My general feeling is that these names are meant for tagging the steps in the display, not so much as comments, so leaving them shorter (and consistent) is probably better. (and `--format SCE` is pretty clear?)
```suggestion
      - name: Download test data
```

jashapiro (approved):
This looks good with just a couple minor naming/style comments.

A separate thought is if we want to track the `fetusref.SeuratData` package in `renv` and not install it as part of a script as we currently do. This would obviously make the docker image bigger, but save repeated downloads. (Other reference files could also potentially be installed as part of the Docker image, but this one is the easiest given current infrastructure.)

jashapiro:
The changes to code that will affect the docker image should really be done in a separate PR... I'm sorry I was not clear about that!

sjspielman:
> The changes to code that will affect the docker image should really be done in a separate PR... I'm sorry I was not clear about that!

Yeah, I realized this seconds promptly after pushing but took the dog for a walk first! Going to revert here :)

PR #828, opened 2024-10-16
06 explore infercnv for Wilms tumor -06

maud-p:

This PR is following the discussion from the PR#776.
It explores the `infercnv` results generated in PR#801.

Please link to the GitHub issue that this pull request addresses.

https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/790

We create a notebook template to explore the `infercnv` results generated using the script `06_infercnv.R`

infercnv was run using the 06_inferCNV.R script with and without a normal reference. We thus tested the impact of the subselection of normal cells using either immune, and/or endothelial cells as healthy reference.

In this notebook, we want to compare the heatmaps of CNV profiles, and evaluate how comparable are the methods and how sensible they are to key parameters such as selection of healthy reference.

We also explore tools to better visualize and interprete CNV results.

Yes, I think the next one will be to test to run `infercnv` with a HMM i3 model, check the running time (costs) and explore the results, in a similar way than in this PR. I would like to try to stick to the use of both endothelial and immune cells as normal cells for reference. As we don't know the behavior of the model in extreme case with only one-five normal cells, I would suggest to try for one sample to graduatly use 1, 5, 10, 50 nornal cells as reference and compare the result. I guess it would at least avoid to take the mean of expression as reference and smooth out CNV present in the majority of cells. 

Opinion very welcome 😃 🚀 

I think the present notebook could maybe be improved with the calculation of a general CNV score and see if this single score could help us identifying normal cells. 
I am however not sure what would be the best way of calculating this score (as simple as the sum of all loss/dupli in each chromosome?)

The script to generated the `infercnv` result required a lot of resource, but the notebook to explore the result do not require much 😄 

- [ x] This analysis module [uses the analysis template and has the expected directory structure](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/analysis-modules/).


sjspielman (commented):
I left a few quick comments from skimming the code, but before I do a larger review here (though fyi, I think this will be fairly quick!), I wanted to ask about your next steps with `i3`. Specifically, it seems to me that you should be able to re-use your existing infercnv script as well as this template notebook, but with a few more arguments, for both i3 and i6 models?. In that case, we'd want to tell the script which model to run as an input option to handle with `optparse`, and we'd want to tell the notebook which model was run with a `param`. I am less familiar with running infercnv than you are, so please tell me if you think this doesn't make sense! 

If it does make sense, then as part of this PR, we should also take the opportunity to the code ready for that next step, including:
- Updating the infercnv script to take an argument for whether you're running i3 or i6, and update the output file names to reflect which model was run. You can choose how you want to handle that organization, as long as the file name indicates if it was i3 or i6.
- Renaming the current result files to have that i6 name in them - you don't need to re-run the script, just rename the files and sync back up with results. This will save a lot of time!!
- Adding a parameter to the rmarkdown file which will help determine which infercnv result file to read in, and make sure that the knitted HTML output file name will contain i3 or i6 too. You'd also want to state clearly in the notebook introduction if it's exploring i3 or i6 results.

I recommend making these changes now since it will affect code in this notebook, and it makes sense to do the script changes simultaneously. It will make this PR a bit larger, but I'm totally ok with that for this one :) Then, your next PR can be running the code with i3 and committing result notebooks. Again let me know if you think this plan doesn't actually make sense for how infercnv works!

maud-p:
Thank you  @sjspielman for the review! 
Actually I already went a bit ahead and wrote another script `07_infercnv_HMM-i3.R` that generate the `Seurat` object `07_infercnv_HMM-i3_{sample_id}_reference-both.rds` that I wanted to submit in the next PR. But I could already push it in this PR or adapt the script `06_infercnv.R` as you suggested (might indeed looks cleaner).

The good news is that it took 2 hours to run `infercnv` with the HMM i3 model for the 5 samples, so it is faster and tmh reasonable (would be 16h for the 40 samples, which is reasonable over a weekend).

maud-p:
so I will now
- [ ] adapt the `06_infercnv.R` and rename manually my results
- [ ] add one section in the notebook to explore HMM i3 results
- [ ] add your changes and re-run the notebook template
- [ ] update the results s3 bucket, and README.md file
- [ ] commit the changes

Hopefully done by tomorrow 😃

maud-p:
Hi @sjspielman ,

So I adapted the `06_infercnv.R` script to have the info if a HMM model have been run and which type (i3 versus i6).
I re-run the script for all reference condition and no HMM model, and for the HMM-i3 model, to make sure paths are OK. 
I simply rename the final `Seurat` object containing `infercnv HMM_type = i6` so the notebook_template can run, but didn't re-run the code. As the script is running for `HHM_type=i3`, I don't see any reason why it wouldn't work for i6 🤞 

Results are uploaded to the s3 bucket researcher-008971640512-us-east-2.

For the next step, after discussing with team members, I would like to try for the same five samples to:
- [ ] select and merge all endothelium and immune cells from these 5 samples
- [ ] spike in this subet of cells in each sample and use it as the healthy reference

If the results are comparable with the one obtained with their own normal cells ech, we could use this spike-in strategy for every sample in the dataset, as we need imho a normal reference, and it should be mixed as possible of cell types to avoid artifact due to cell type specific regions on some chromosome (17 for immune cells for example).

Thank you 😃

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
```diff
@@ -0,0 +1,445 @@
...
+- `infercnv_obj` is the `infercnv` object
+
+- `group.by` is the metadata used for grouping the violin plots
+
+```{r}
+
+Do_CNV_heatmap <- function(object, infercnv_obj, group.by){
+  
+
+  chromosome_locations = SCpubr::human_chr_locations
```
sjspielman, line 83: We can just this directly
```suggestion
```

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
```diff
@@ -0,0 +1,445 @@
...
+
+  out <- SCpubr::do_CopyNumberVariantPlot(sample = object,
+                                        infercnv_object = infercnv_obj,
+                                        using_metacells = FALSE,
+                                        chromosome_locations = chromosome_locations,
```
sjspielman, line 88:
```suggestion
                                        chromosome_locations = SCpubr::human_chr_locations,
```

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
```diff
@@ -0,0 +1,445 @@
...
+```
+
+#### Calculate a global CNV score per cell to check the general distribution
+
+For a `Seurat` object an `infercnv` object created with the script `06_infercnv.R` using `reference_value` as a reference, the function `Do_CNV_score` load the `infercnv` object and calculate a CNV score per cell.
+The score is calculated based on the [biostar discussion](https://www.biostars.org/p/9573777/).
+It 
+
+- `reference_value` is the selection of normal cells used for `infercnv`
+
```
sjspielman, line 108: I think this part still needs to be written?

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
```diff
@@ -0,0 +1,445 @@
...
+  
+  return(out)
+}
+
+```
+
+#### Calculate a global CNV score per cell to check the general distribution
+
+For a `Seurat` object an `infercnv` object created with the script `06_infercnv.R` using `reference_value` as a reference, the function `Do_CNV_score` load the `infercnv` object and calculate a CNV score per cell.
+The score is calculated based on the [biostar discussion](https://www.biostars.org/p/9573777/).
```
sjspielman, line 104: Is there a particular reason why you chose boundaries of 0.9 and 1.1, since 0.95 and 1.05 is presented in the discussion? It's worth explaining that reasoning here

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
```diff
@@ -0,0 +1,445 @@
...
+
+```
+
+## Analysis
+
+### Heatmap of infercnv results
+
+Here we plot the output of `infercnv` as heatmaps of CNV.
+We first look at the png file generated by the `infercnv` function.
+We then used the `infercnv object` to look at mean CNV value across compartments (immune, endothelial, stroma and fetal nephron).
```
sjspielman, line 266: At first I was going to leave a review here that it's worth noting these results are generated without HMM. But then I wondered - are these results identical regardless of HMM, and specifying an HMM just gives you _additional_ information on top of what running without would give you? Is that your experience with infercnv? If yes, then this simplifies how you can run `06_infercnv.R`. You would never need to run it without an HMM, since that would just duplicate the main CNV results.  If no, and results are different, then why don't we show the heatmaps here for HMM results?

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
```diff
@@ -0,0 +1,445 @@
...
+  dplyr::pull(subdiagnosis)
+
+```
+
+This notebook explores using [`infercnv`](https://github.com/broadinstitute/inferCNV/wiki) to estimate tumor and normal cells in `r params$sample_id` from SCPCP000006. 
+This sample has a(n) `r subdiagnosis` subdiagnosis.
+
+`infercnv` was run using the `06_inferCNV.R` script with and without a normal reference. We also tested the impact of the subselection of normal cells using either immune, and/or endothelial cells as healthy reference.
+
+In this notebook, we just want to compare the heatmaps of CNV profiles, and evaluate how comparable are the methods and how sensible they are to key parameters such as selection of healthy reference.
```
sjspielman, line 47: I would also mention that you are exploring the use of the HMM, as well, and briefly explain the difference among "no", "i6",and "i3" conditions.

analyses/cell-type-wilms-tumor-06/scripts/explore-cnv-methods.R
```diff
@@ -16,39 +16,69 @@ for (sample_id in c("SCPCS000179",
...
+  
+  system(command = glue::glue("Rscript ", file.path(module_base,"scripts", "06_infercnv.R"), " --sample_id ", sample_id, " --reference ", "both", " --HMM ", "no"))
+  
+  # We run and explore infercnv using both endothelial and immune cells as reference and i3 HMM model
+  
+   system(command = glue::glue("Rscript ", file.path(module_base,"scripts", "06_infercnv.R"), " --sample_id ", sample_id, " --reference ", "both", " --HMM ", "i3"))
+
+  # We run and explore infercnv using both endothelial and immune cells as reference and i6 HMM model
+  
+ #  system(command = glue::glue("Rscript ", file.path(module_base,"scripts", "06_infercnv.R"), " --sample_id ", sample_id, " --reference ", "both", " --HMM ", "i6"))
```
sjspielman, line 68:
```suggestion
   system(command = glue::glue("Rscript ", file.path(module_base,"scripts", "06_infercnv.R"), " --sample_id ", sample_id, " --reference ", "both", " --HMM ", "i6")) 
```

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
```diff
@@ -0,0 +1,445 @@
...
+
+```
+
+#### Plot UMAP of known Wilms tumor related alteration
+
+As a reminder, we plot the label transfer from the fetal kidney atlas. 
+
+```{r fig.width=20, fig.height=8, out.width='100%'}
+
+visualize_metadata(srat, "seurat_clusters", "fetal_kidney_predicted.compartment")
```
sjspielman, line 331: All the colors in this plot are very hard to see. Maybe it would make sense to flip the clusters and compartment labels, so that clusters are on the x axis and compartment is the colors?

sjspielman (commented):
The i3/i6 code looks good and results are clearly generated, so that's good to see! I left a few small comments. As far as your next steps go, I have a few thoughts:

- First, which findings in particular to you hope to be able to see when combining "normal" cells across references, specifically in terms of _validating_ this approach? Having some hypotheses here will be a helpful guide: What will results look like if it's a good strategy to combine normal cells across patients? What will the results look like if it's not a good strategy?
- Second, I'm not entirely sure how to think of the influence batch effects & patient variability for this analysis. I might expect different profiles of "normal" cells across patients just driven by biological variability, and it might not provide a good baseline and/or be very influenced by batch effects. That said, we probably don't want to be correct for batch effects data here since that will smooth out biological signal which I assume is needed for this kind of analysis (right?). This is another reason why it would be good to have a clear hypothesis in mind - if you have a particular trend you're looking for already, it will help interpret whether the merging strategy itself is reliable or if it's producing artifacts.
- You wrote in your notebook that you wonder if some endothelial cells are misclassified. Indeed, that's always a concern! Something we did not do is look in-depth at the scores for some of these classifications. I wonder if exploring the scores would be helpful for you to pick which cells to use as normal. I don't really know how these scores are calculated, but I assume that we might be more confident in cell labels if their scores are higher above some threshold? This might be worth checking - if many of the endothelial cells, for example, have low scores, perhaps that suggests they are not a reliable reference? This might also be something to discuss with your team.

Also, anytime you want, you can always feel free to open a Discussion to get feedback more generally for your next steps so other Data Lab team members and/or contributors have a chance to weigh in too :)

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
maud-p: hm right, actually I first used 0.95 and 1.05 and the results were not really convincing (no higher scores for normal cells) so I decided to be more strict with the thresholds. But as you see it hasn't improved... not sure what to do with this score at that stage, might be to remove. But in the exploratory notebook might also be to keep and explain that it is not what we would have expect.

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
maud-p: From my understanding, HMM prediction is an _additional_ step happening _after_ the [standard inferCNV processing routines](https://github.com/broadinstitute/infercnv/wiki/Running-InferCNV). 
I manually checked the heatmaps `06_infercnv_{sample_id}_reference-both_HMM-no_heatmap.png` and `06_infercnv_{sample_id}_reference-both_HMM-i3_heatmap.png`, they look indeed identical. 

The reason why I first ran `infercnv` without the additional HMM CNV prediction model was to _quickly_ compare the choice of the _normal reference_. I think the HMM model is not required to look at the variability between the CNV profiles when taking no, or immune and/or endothelial cells as normal reference. 

From thee 4 heatmaps, I conclude that **we need to use normal cells as reference**. Else, the mean of expression is taking as the normal and the more abundant cells will thus look normal, as illustrate with the example in sample -179:
image
Here, fetal nephron, which we _know/think_ contains cancer cells, looks the more _normal_. Also chr17 seems to be gained in immune and endothelial cells, which is an artefact due to the fact that chr17 encodes lot of chemokines secreted by the 2 cells types. 

Then, taking the endothelial cells as reference, it seems that _all_ compartments have a chr3 and chr5 loss. 
image

This is however an artefact due to the fact that key endothelial genes are encoded in chr3 and chr5. This artefact is corrected when using both immune and endothelial cells in the reference. 
image

maud-p: > you would never need to run it without an HMM, since that would just duplicate the main CNV results.

I agree for the condition `-- reference "both"`, I could only run it with the HMM model. 
For the other `--reference` conditions, I would however not spend too much time and resources calculating the HMM model.  

Is that what you mean?

maud-p:
> *  I'm not entirely sure how to think of the influence batch effects & patient variability for this analysis. I might expect different profiles of "normal" cells across patients just driven by biological variability, and it might not provide a good baseline and/or be very influenced by batch effects. That said, we probably don't want to be correct for batch effects data here since that will smooth out biological signal which I assume is needed for this kind of analysis (right?). This is another reason why it would be good to have a clear hypothesis in mind - if you have a particular trend you're looking for already, it will help interpret whether the merging strategy itself is reliable or if it's producing artifacts.

I 💯 % agree with your concerned regarding the batch effect and patient variability and therefore I don't know if this approach will work. But this is the only strategy I can imagine so far for the **patient with very few normal cells**, as we showed that in the Wilms tumor dataset, `infercnv` cannot be ran without a reference and that using multiple cell types (immune _and_ endothelial) seem avoiding most of the cell types induced CNV artifacts (_~FALSE positive CNV?_). See my related [comment](https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/828#discussion_r1807011862) above.

I hope (🤞) that as the reference is then the mean of expression of all reference cells, batch and patient effect induced by merging the immune and endothelial cells from multiple patients will be smoothed while taking the _mean_. (would that make sense?)

> * First, which findings in particular to you hope to be able to see when combining "normal" cells across references, specifically in terms of _validating_ this approach? Having some hypotheses here will be a helpful guide: What will results look like if it's a good strategy to combine normal cells across patients? What will the results look like if it's not a good strategy?

I was thinking to the following strategy:
- [ ] select for all _non treated_ patients immune and endothelial cells (as I am not sure if and how chemotherapies can impact the CNV profile of these normal cells) and create a `Seurat` object with these cells. _I would adapt the names of the cells with a TAG like `spike_{sample_id}` to ensure that I can easily track these cells after spike in._
- [ ] modify the `06_infercnv.R` script to, for each sample, merge the spike-in `Seurat` object with the {sample_id} object, define the normal cells as all spike-in cells and run `infercnv` for the 5 selected samples that do have `enough` normal cells. 
- [ ] I will compare the CNV profiles obtained with _patient-dependent normal cells_ (what we already ran) and with the _inter-patient reference of normal cells_.

> _validating_ this approach

We were thinking that independently of the approach, the best validation would be to have a ground truth (SNP array?), or at least some information for some samples. I think looking at 1q in clinic for Wilms tumor patient should be a standard in the COG protocol. Getting this 1q status for some patients would allow us to check (even if not completely) our `infercnv` results. What do you think?

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
sjspielman: > The reason why I first ran infercnv without the additional HMM CNV prediction model was to quickly compare the choice of the normal reference.

This makes total sense to me as the right place to start! 

> Is that what you mean?

Yes, this basically what I mean! At this point, we know that the results are the same with and without HMM. I also looked a bit through inferCNV's code, and it indeed seems that the HMM doesn't affect the other code. There is just an `if` statement to add on the HMM, but it doesn't change the other calculations. Therefore, you never need to run without HMM. Even if this only saves a small amount of time and resources, especially for running in CI.

sjspielman:
Hi @maud-p, just a quick update - we're out of town at a conference this week, so I will be in and out of being able to review. I left some comments on your other PR looking at scores which mostly focus on the need to have more of a justification for why you have chosen 0.85. Once we have a threshold identified in that PR, we can follow up here on actually using that threshold. I've briefly looked at the changes to your inferCNV script, and looks mostly fine (but I haven't run it yet). I do recommend adding a bit more documentation for what "pull" means, since it is not obvious from any comments in the code yet.

In terms of the external validation you mention, it seems potentially promising but I imagine it would be quite a bit of additional work that would not be able to happen before the cell type submission deadline. This might be something to follow up on after you generate your first cell type annotation table as a way to refine the analysis in the future - these modules can certainly be "living" analyses well into the future!

maud-p:
Hi @sjspielman, thank you!

I'll add few lines about the `pull` parameter then, just for you here it indicates that we take as normal reference for `infercnv` a _pull_ of normal cells identified in Wilms tumor patient (only non-treated Wilms tumor samples, endothelial and immune cells with a `mapping.score > threshold (0.85)`.

Regarding the deadline, I wanted to let you know that _my_ deadline is the 29/10 as I will be traveling after this and won't be able to work/focus on the analysis the last 3 days... 😕  However, I think we are on a good track to have within the next weeks solid annotations and I would definitely like to continue working on it after the deadline. Also to then use the annotations to answer more specific questions 🚀

sjspielman:
Hi @maud-p, I don't see any new commits to clarify the `pull` parameter here. Maybe not all of the code changes you made were pushed?

sjspielman (commented):
I've returned to this review by trying to run through the notebook code, and I encountered quite a few bugs running the code from a fresh R session so I was not able to run it. Here is some of what I observed:
- It appears you need the `ggdist` dependency for some of your plotting. To add this, you'll need to add `library(ggdist)` to the file `components/dependencies.R`, and then re-run `renv::snapshot()` to save it to the lockfile.
- The `visualize_feature()` function is not working properly. It looks like you are using the `features` argument as two separate variables - as both a character vector of features to plot, and the plot title. When you use this function, the argument you are passing in corresponds to the plot title, not a vector of features to plot. Something here needs to be fixed, probably both with the function itself and how you call it.
- I wasn't able to make most of the plots for the Seurat object from the `pull` reference condition, but I'm not 100% sure what's going on there - maybe something with how you constructed the Seurat object itself in the `06b_build-normal_reference.R` is not consistent with the other Seurat objects, so code cannot be generalized?

Generally speaking, this notebook does have quite a bit of repeated code where you explore the same plots for each of the inferCNV results. It would be nice to firm this up to avoid future bugs. For this, you'd want to make a wrapper function to perform all the tasks you need for each result. All of the test you have explaining the plots can be added in a separate section that provides the overall context. On one hand, doing this may actually help you fix existing bugs! But on the other hand, I recognize you are short for time right now, so instead it might be best to file an issue to circle back to to improve the molecularity of this notebook. This way we'll have this fully tracked for later. But again, this may help you debug in the first place, so it could be worth doing now. Let me know if you want to chat about this more!!

Separate from the notebook, I'm confused about the result files you have here specifically. In the `results` directory, you have files, e.g. for sample `SCPCS000184`, which are _not saved_ in the `06_infercnv` directory. 
- `06_infercnv_HMM-i3_SCPCS000184_reference-both.rds`
- `06_infercnv_HMM-i3_SCPCS000184_reference-pull.rds`
- `06_infercnv_HMM-i6_SCPCS000184_reference-both.rds`

It isn't immediately clear to me how these files differ from files in the `06_infercnv` folder, and further which ones are meant to be used when. More documentation will help here! It would be really helpful to document this in `results/README`, and more importantly, in the `06_infercnv.R`script itself. At the top of that script, please explain which files it exports. 

Let me know where I can clarify or help here!

maud-p:
> Hi @maud-p, I don't see any new commits to clarify the `pull` parameter here. Maybe not all of the code changes you made were pushed?

Hi @sjspielman , could be that I haven't pushed correctly all the modifications. Tbh I have a bit of troubles working on different branches. I'll try to correct this!

In the meantime, I realized that I was actually meaning _pool_ and not _pull_ .... sorry for the confusion, my English is not the best... 😕

sjspielman:
> Tbh I have a bit of troubles working on different branches. I'll try to correct this!

Let me know if I can help in any way! Happy to hop on Slack and chat, feel free to DM me there! Still a bit of time today and tomorrow :) 


> In the meantime, I realized that I was actually meaning pool and not pull ...

Ah this makes sense, but it's no problem! I was thinking it meant to "pull" annotations from one dataset to another, so it made sense to me this way too!

maud-p:
> Let me know if I can help in any way! Happy to hop on Slack and chat, feel free to DM me there! Still a bit of time today and tomorrow :)


Thank you! I try to take your comments one by one and keep going, I'll let you know if I get stuck! Thank you 😃

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
```diff
@@ -0,0 +1,765 @@
...
+### HMM-i3 CNV prediction model with an inter-patient reference of immune cells (endothelial and immune)
+
+We lastly explore `infercnv` results generated with a pull of immune and endothelial cells from different Wilms tumor samples as reference, using a HMM-i3 prediction model.
+
+The selection of normal cells to be spiked-in as a normal reference prior running `infercnv` has been performed in `06a_build-normal-cell_reference.R.
+We selected samples that haven't been pre-treated with chemotherapies. Indeed, even if normal cells shouldn't be affected by chemotherapy in terms of CNV, we are not sure how chemotherapies can affect the CNV profile of normal cells.
+We thus decided to only take endothelial and immune cells from non-treated samples to build the reference of normal cells.
+
+
+#### Load the `Seurat` object
```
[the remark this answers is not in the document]
maud-p: @sjspielman I think you are right, might e an `Error` in the construction if the `srat` object but I can't figure it out... I am not able to reproduce the `Error` 🤔 

What about removing the exploration of the `infercnv with `_pull_ = _pool_ of inter-patient normal cells as reference for now, as it can be quite controversial on its own and I am able to run `infercnv` on 39/40 samples with a intra-sample normal reference? 

Might be something that we investigate then in a later PR?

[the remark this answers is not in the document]
maud-p: I tried to adapt the `Seurat` object construction to make it safer and hopefully avoir `Error`. Let me know if it helps, it is running from my side 🤞

sjspielman:
> @sjspielman I am not sure to understand the issue here. As I am only using the visualize_feature() for a single feature, the feature can be a character and do not need to be a vector. This might not be very elegant but should work? Or is there something that I don't get? Thank you!

Ah, I understand what you mean here. This was an idea that I had for why the function wasn't working for me when I tried to run it, but it sounds like that was not the bug and there is a different reason it doesn't work for me. Can you confirm if it is currently working for you? I can try to do some more thorough debugging on my end!

maud-p:
> > @sjspielman I am not sure to understand the issue here. As I am only using the visualize_feature() for a single feature, the feature can be a character and do not need to be a vector. This might not be very elegant but should work? Or is there something that I don't get? Thank you!
> 
> Ah, I understand what you mean here. This was an idea that I had for why the function wasn't working for me when I tried to run it, but it sounds like that was not the bug and there is a different reason it doesn't work for me. Can you confirm if it is currently working for you? I can try to do some more thorough debugging on my end!

it is working for me :)
I changed a bit the notebook by the way, I should have simplify the transfer of `infercnv metadata` to the seurat object, to reduce Èrror`. Still not working on your side like this?

sjspielman:
> Still not working on your side like this?

I will pull your current code and test it again shortly!

sjspielman:
It still doesn't work for me a lot of the time, but I still can't tell why. I get the error `Error in .subset2(x, i, exact = exact) : subscript out of bounds` _some of the time_ that I use it which implies that something is not present in the object as expected, but everything does seem to be there. Sometimes, I seem to be able to use `Seurat::FeaturePlot()` in place of `SCpubr::do_FeaturePlot()` which does work, but not always. I will keep looking into this, but in the meantime, can you make sure all the correct versions of the results files have been pushed to your bucket?

maud-p:
strange... is that for every sample or one in particular? for which of the `feature` is that failing or working?
Thank you for your help, I'll re-load the s3 bucket. Some of the object are missing, I need to look more into details why. I guess at least the one with only one normal cell (177).

sjspielman:
So far I have only tested with the default sample in the notebook but I am currently downloading results for SCPCS000194 to try that one too.

sjspielman:
It ran through for me with SCPCS000194! I'm going to redownload SCPCS000184 and try again with that sample, too.

sjspielman:
I figured out the issue! I had regenerated label transfer results locally with the current data release, but it appears that you used an older data release version for this analysis, and similarly the results in the bucket were probably generated a while ago. When I re-download the label transfer results (namely, the 02b Rds files) from the results bucket, the code works. Given that code works consistently with files generated from the same data release, this code should all fine. We in the Data Lab will address any issues on our end as we continue to work on updating the Azimuth code, but you are fine to continue here!!

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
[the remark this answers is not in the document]
sjspielman: Noting we found the issue - we were using input 02b files generated from different data releases.

maud-p:
> I figured out the issue! I had regenerated label transfer results locally with the current data release, but it appears that you used an older data release version for this analysis, and similarly the results in the bucket were probably generated a while ago. When I re-download the label transfer results (namely, the 02b Rds files) from the results bucket, the code works. Given that code works consistently with files generated from the same data release, this code should all fine. We in the Data Lab will address any issues on our end as we continue to work on updating the Azimuth code, but you are fine to continue here!!

Thank you so much for your efforts!!!

analyses/cell-type-wilms-tumor-06/scripts/06_infercnv.R
```diff
@@ -8,6 +8,11 @@
 #   --sample_id SCPCS000179
 #   --reference one of none, immune, endothelium, both
 
+# OUTPUT :
+# For every condition, the `infercnv` object and final `infercnv` heatmap are saved in the corresponding subfolder
```
sjspielman, line 12:
```suggestion
# For every condition, the `infercnv` object and final `infercnv` heatmap are saved in the corresponding {sample_id}/06_infercnv results subfolder
```

analyses/cell-type-wilms-tumor-06/scripts/06_infercnv.R
```diff
@@ -8,6 +8,11 @@
...
+# When running the HMM- CNV prediction, the proportion of CNV per chromosome is saved as a metadata for each cells.
+# The updated `Seurat` object is additionally saved under the {sample_id} subfolder for easy use downstream.
```
sjspielman, line 14:
```suggestion
# The updated `Seurat` object is additionally saved under the {sample_id} results subfolder for easy use downstream.
```

analyses/cell-type-wilms-tumor-06/scripts/06_infercnv.R
```diff
@@ -26,7 +31,13 @@ option_list <- list(
     opt_str = c("-r", "--reference"),
     type = "character",
     default = "both",
-    help = "Reference cells to use as normal cells, either none, immune, endothelium or both"
+    help = "Reference cells to use as normal cells, either none, immune, endothelium, both or pull"
```
sjspielman, line 34:
```suggestion
    help = "Reference cells to use as normal cells, either none, immune, endothelium, both or pull. The 'pull` option will pull in normal cells from other samples to include in the reference."
```

analyses/cell-type-wilms-tumor-06/notebook_template/06_cnv_infercnv_exploration.Rmd
```diff
@@ -0,0 +1,569 @@
...
+  meta$has_cnv_score <- rowSums(meta[,grepl("has_cnv_chr", colnames(meta))])
+  meta$has_cnv_score[meta$has_cnv_score > cnv_threshold] <- "CNV"
+
+  # we know that some immune cells might have false positive CNV 
+  meta$has_cnv_score[meta$has_cnv_score %in% c(0:cnv_threshold)] <- "no CNV"
+  infercnv_obj <- AddMetaData(infercnv_obj, meta$has_cnv_score, col.name =  "has_cnv_score") 
+  p[["infercnv_obj"]] <- infercnv_obj
+
+  p[["cnv_seurat"]] <- visualize_metadata(infercnv_obj, "has_cnv_score", "seurat_clusters")
+  p[["cnv_compartment"]] <- visualize_metadata(infercnv_obj, "has_cnv_score", "fetal_kidney_predicted.compartment")
```
sjspielman, line 307: Adding some spaces to make this a bit easier to read

```suggestion

  #### Plot UMAP of known Wilms tumor related alteration
  p[["compartment"]] <- visualize_metadata(infercnv_obj, "fetal_kidney_predicted.compartment", "seurat_clusters")
  
  #### Feature plot and repartition of the CNV per chromosome
  for(i in 1:22){
    tmp1 <- visualize_feature(infercnv_obj, feature=glue::glue("proportion_cnv_chr", i), group.by = "fetal_kidney_predicted.compartment")
    tmp2 <- visualize_density(infercnv_obj, features=glue::glue("proportion_cnv_chr", i), group.by = "fetal_kidney_predicted.compartment")
    p[[glue::glue("proportion_cnv_chr", i)]] <- tmp1 + tmp2 + plot_layout(ncol = 3, widths = c(1,2,2))
  }
  
  #### DotPlot
  p[["Dotplot"]] <- DotPlot(infercnv_obj, features =  colnames(infercnv_obj@meta.data)[grepl("has_cnv_chr", colnames(infercnv_obj@meta.data))], group.by = "fetal_kidney_predicted.compartment", assay = "SCT") + 
  theme(axis.text.x = element_text(angle = 90, vjust = 0.5, hjust=1))
  
  #### CNV score
  meta <- infercnv_obj@meta.data
  meta$has_cnv_score <- rowSums(meta[,grepl("has_cnv_chr", colnames(meta))])
  meta$has_cnv_score[meta$has_cnv_score > cnv_threshold] <- "CNV"

  # we know that some immune cells might have false positive CNV 
  meta$has_cnv_score[meta$has_cnv_score %in% c(0:cnv_threshold)] <- "no CNV"
  infercnv_obj <- AddMetaData(infercnv_obj, meta$has_cnv_score, col.name =  "has_cnv_score") 
  p[["infercnv_obj"]] <- infercnv_obj

  p[["cnv_seurat"]] <- visualize_metadata(infercnv_obj, "has_cnv_score", "seurat_clusters")
  p[["cnv_compartment"]] <- visualize_metadata(infercnv_obj, "has_cnv_score", "fetal_kidney_predicted.compartment")
```

sjspielman:
@maud-p Before I post my review here - please _pull_ this code to update locally, since I merged in main to fix the conflict. Let me know if you have any issues after doing that!

sjspielman (commented):
Thanks as always for all your hard work on this PR!! I have left a final round of comments, all very small changes. Before accepting any suggestions, please remember to pull first and make sure everything is ok. Then, you'll want to re-render the notebooks so they have the spacing changes, and after that I expect to approve :)

maud-p:
I encounter some issue to _pull_, I have some conflicts. So I am cloning in a _new_ folder the repo locally and will go from here, just take a bit of time ;)

maud-p:
just to have it tracked somewhere, I did : `git checkout -b 06_updated_branch d7f4cd801b4ec43382cf2669c027ac7d9682437b` which should be your last merge

maud-p:
> Please feel free to DM me on Slack if I can help resolve the conflicts!

should be running :)

sjspielman:
Looks like conflicts are resolved! But, please still re-render the notebooks since there was a change there, and re-request my review when they are pushed :) Then, I expect this will be ready to merge!

sjspielman (approved):
Looks good!!!

maud-p:
@sjspielman thank you so much for the very productive day today!!
I just wanted to let you know, I should open the last PR tonight to draft the first annotations. I'll be traveling tomorrow and Thursday, but I still hope to find some time to answer your comments 😃 🤞

sjspielman:
Code in this PR does not (currently) run in CI, so we can merge this before the final check is complete since we know that `renv` builds.

PR #829, opened 2024-10-16
Install fetusref package and track in renv

sjspielman:
This PR updates `cell-type-wilms-tumor-06` to track `fetusref.SeuratData` with renv instead of installing it in a script. I also removed Seurat and Azimuth from `dependencies.R` since I'm confident at least they don't need to be there. `renv` apparently decided to add more BioC remotes too, which seems fine.

Noting this should go in _after_ #827.

### DIFF
--- a/analyses/cell-type-wilms-tumor-06/components/dependencies.R
+++ b/analyses/cell-type-wilms-tumor-06/components/dependencies.R
@@ -3,9 +3,8 @@ library(tidyverse)
 library(assertthat)
 
 # Single-cell packages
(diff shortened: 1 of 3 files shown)


sjspielman:
Hm actually, the script update to use the installed package in renv needs to be separate too. So, I'll do this with a fresh brain tomorrow instead...

jashapiro:
I know this is closed, but you should direct the renv PR to `main` so that the changes get pushed to the docker repository. Once that is in, then you can send all the code changes to your feature branch.

Issue #830, opened 2024-10-17
Ensure fetusref.SeuratData is versioned in cell-type-wilms-tumor-06


sjspielman:
Currently, this module is installing `fetusref.SeuratData` from a script with `SeuratData::InstallData()`. This may not be stable in the long-term, so we'd like a different approach to downloading this reference in a more robsut way. 

First, we've observed that the data itself is an R package. We might like to track this more formally with `renv`, since the dataset itself is a versioned package that lives at `http://seurat.nygenome.org/src/contrib/fetusref.SeuratData_1.0.0.tar.gz`. For this, we'd need to:
- [ ] First, test whether and how exactly to track this in `renv`. In some local testing, there may be system dependencies that this is installing when you restore a `renv` environment where this package is tracked. Mac isn't granting permissions (reasonably) to install those deps, so we'll want to test whether an environment can indeed be restored on a LSfR instance, whether other dependencies are needed, and further how exactly (if it's not intuitive) to load the package to be able to access the data.
- [ ] Update renv to track this package. This should be in a PR by itself since the Docker image will need to be built
- [ ] Update code to no longer install this package

But, step 1 may not pan out, and/or it may be a rabbit hole. Alternatives are therefore to download the reference directly from one of these two spots, save it in `scratch`, and then load it up with `Azimuth::LoadReference()`. 
- `curl https://seurat.nygenome.org/src/contrib/fetusref.SeuratData_1.0.0.tar.gz` and manually pull out the dataset for use. The reference data lives in `inst/azimuth`.
- Download the reference from where it permanently lives in Zenodo: https://zenodo.org/records/4738021#.YJIW4C2ZNQI. **This is probably my preference**, since I am most confident the Zenodo DOI will continue to exist indefinitely, but the Seurat links may change in the future.

jashapiro:
> [ ] First, test whether and how exactly to track this in `renv`. In some local testing, there may be system dependencies that this is installing when you restore a `renv` environment where this package is tracked. Mac isn't granting permissions (reasonably) to install those deps, so we'll want to test whether an environment can indeed be restored on a LSfR instance, whether other dependencies are needed, and further how exactly (if it's not intuitive) to load the package to be able to access the data.

Did you try installing this package with `renv::install("https://seurat.nygenome.org/src/contrib/fetusref.SeuratData_1.0.0.tar.gz")` ?

[...]

PR #831, opened 2024-10-17
Add fetusref.SeuratData to renv

jashapiro:
Part of #830

Here I am adding the `fetusref.SeuratData` file to the `renv` for  `cell-type-wilms-tumor-06`. I also switched the `Seurat` install from GitHub to CRAN, which should make installations faster. 

To test that things are working as expected, I also removed the `InstallData` step from the notebook that was using it. 

Both local installs and Docker installation seem to work fine, and I confirmed that the command `SeuratData::LoadData("fetusref", "azimuth")` worked as expected in both (with `Seurat` loaded).

I also added a `.Rproj` file for convenience in development.


sjspielman:
Just to update - I'm testing a fresh `renv::restore()` for this on LSfR, and it's very much taking it's time... Stay tuned!
With any luck, posting this comment will speed it along 😃

sjspielman (approved):
> With any luck, posting this comment will speed it along 😃

🍀 

I was able to restore the environment and load the data as you describe 🚀

jashapiro:
I'm not merging this quite yet, because I discovered an odd bug: if you open a terminal in RStudio on lightsail from within a project, it for some reason adds `/usr/bin` to the path, which is no good. I have some ideas, but I want to be able to test them a bit.

PR #842, opened 2024-10-28
Use docker image for wilms-14 CI

jashapiro:
This PR activates using the Docker image for testing the `cell-type-wilms-tumor-14` module. Hopefully this will speed up testing!

Filing as a draft for testing first


jashapiro:
Not sure why this was failing CI for a while (with an error 6?) but it seems to be going fine now. I had added a few echo statements in the shell script to try to figure out where the failure really was, but I'll leave them there because it was kind of nice to have.

analyses/cell-type-wilms-tumor-14/run_cell-type-wilms-14.sh
```diff
@@ -23,6 +23,8 @@ scratch_dir="scratch"
 results_dir="results"
 data_dir="../../data/current/SCPCP000014"
 
+ls -l ${data_dir}
```
sjspielman, line 26: Did you want to keep this too, or just the echos?

.github/workflows/run_cell-type-wilms-tumor-14.yml
```diff
@@ -33,41 +33,22 @@ jobs:
...
-            jags \
-            libcurl4-openssl-dev \
-            libfribidi-dev \
-            libglpk40 \
-            libharfbuzz-dev \
-            libhdf5-dev \
-            libmagick++-dev \
-            libtiff5-dev
+          apt-get update
+          apt-get install -y git
```
sjspielman, line 42: What is git used for in this module, and why not add it to Dockerfile directly? I don't see anything particular in the scripts directory that needs the executable, so mostly I am just curious what's up here.

jashapiro: git is used to check out the repository. If it is not there, then checkout step uses the API, which is less efficient. It might have even failed once, but I'm not sure I remember exactly.

analyses/cell-type-wilms-tumor-14/run_cell-type-wilms-14.sh
jashapiro: No, I can remove that
```suggestion
```

.github/workflows/run_cell-type-wilms-tumor-14.yml
sjspielman: > If it is not there, then checkout step uses the API, which is less efficient. 

Ahhhh this makes sense. As the repo continues to grow, maybe we consider adding this to more workflows then

jashapiro: Looked at the failure: yes, it failed because when using the API for checkout, no `.git` directory is created, which means that we can't use that to find the root directory.

sjspielman: oh that certainly _does_ make sense!

analyses/cell-type-wilms-tumor-14/run_cell-type-wilms-14.sh
```diff
@@ -31,22 +32,26 @@ meta_path="${data_dir}/single_cell_metadata.tsv"
...
+
 # Download and process reference data
 ref_h5ad="${scratch_dir_step}/Fetal_full_v3.h5ad"
 ref_seurat="${scratch_dir_step}/kidneyatlas.rdsSeurat"
 ref_seurat_sct="${scratch_dir_step}/kidneyatlas_SCT.rdsSeurat"
 
+echo "Downloading reference data"
 if [[ ! -e ${ref_h5ad} ]]; then
     ref_url="https://cellgeni.cog.sanger.ac.uk/kidneycellatlas/Fetal_full_v3.h5ad"
     curl -s -o ${ref_h5ad} ${ref_url}
```
jashapiro, line 45: The error 6 seems to be coming here, so I am going to add a few options to try to make this more reliable and add the error report
```suggestion
    curl -s --show-error -L -o ${ref_h5ad} ${ref_url}
```

sjspielman: I wonder if this would be a more stable link?
https://datasets.cellxgene.cziscience.com/fcb6225a-b329-4ac9-8a3f-559ca9bac50e.h5ad

From https://cellxgene.cziscience.com/collections/120e86b4-1195-48c5-845b-b98054105eec, which should be [the same](https://www.kidneycellatlas.org/). This is where we download from for wilms-06, which I'm realizing probably also needs to be changed to an anndata link...

jashapiro: Unfortunately, these are not the same file. I can't tell you what the difference is, but they are not the same size, so I don't want to substitute for the purposes of this PR

This seems to be a DNS issue; I never have trouble downloading the file locally.

jashapiro:
@sjspielman I can not guarantee that the forcing of ipv4 was sufficient to fix CI here, but the download does seem to be working for at least the last two attempts.

sjspielman:
> but the download does seem to be working for at least the last two attempts.

As good as we'll get, at least for now! If it continues to fail we can revisit it.

sjspielman (approved):

Issue #852, opened 2024-11-04
Update the wilms 06 workflow script and GHA


sjspielman:
_Broken out of #810_

We'll need to update the `cell-type-wilms-tumor-06/00_run_workflow.sh`script, along with a few updates to the module workflow. Specifically, we want to:

- Add a flag to the workflow script for whether to run exploratory steps which do _not_ contribute to the cell type annotation itself. We should not run them in CI, but they should be present to run the full workflow. The exploratory steps include:
  - Rendering the `03_clustering_exploration.Rmd` notebook
  - Rendering the `04_annotation_Across_Samples_exploration.Rmd` notebook
  - Running the `explore-cnv-methods.R` script
- Add another sample (SCPCS000190) to run in the test workflow, since this sample is treated differently and we want that condition covered https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/7511ae09c5fb5606574c68f9657dd0be56b13440/.github/workflows/run_cell-type-wilms-tumor-06.yml#L57 _Edit: I realize this may not be the case when we reprocess label transfer, so keep tabs!_
- Run the workflow!!
  - We can expect some debugging upon discovery here, which may need to be broken up into additional issues

sjspielman:
> Run the workflow!! 

Noting that these results may best be committed in separate PR(s), since there are quite a few notebooks we'll need to re-render:
- Per sample:
  - `01_seurat_processing_{sample id}.html`
  - `02a_fetal_all_reference_Cao_{sample id}.html`
  - `02b_fetal_kidney_reference_Stewart_{sample id}.html`
  - `03_clustering_exploration_{sample id}.html`
- `04_annotation_Across_Samples_exploration_predicted.score_threshold_{threshold}.html`
  - where `threshold \in {0.5, 0.75, 0.85, 0.95}`
- `07_combined_annotation_across_samples_exploration.html`

jashapiro:
> * Add a flag to the workflow script for whether to run exploratory steps which do _not_ contribute to the cell type annotation itself. We should not run them in CI, but they should be present to run the full workflow.

Why not just put exploratory analyses in a separate workflow script?

sjspielman:
> Why not just put exploratory analyses in a separate workflow script?

Mostly because these aren't standalone steps - they depend on output from previous steps that are not exploratory. I suppose this would work with an assumption that the other steps/"main script" was already run, if this is an assumption we're comfortable with.

[...]

PR #859, opened 2024-11-06
1/n: Update wilms 06 module

sjspielman:
Towards #850 
Part of #852 

This PR begins the next stages of updates for the Wilms 06 module:

- For #852, I've updated the GHA workflow to run a few more samples through, which will (soon!) be treated differently by `inferCNV` so we'd like to include them in testing
- I re-rendered all the 01 (seurat processing), 02a (first label transfer), and 02b (second label transfer) notebooks

The next PR will include updated 03 (clustering) notebooks, as well as code updates to allow that notebook to run in CI if desired. That said, it generally won't be (happy to hear a different opinion here though!) since it is strictly exploratory and doesn't contribute to the final annotations, as noted in #852.


jaclyn-taroni (approved):
Approving on the basis of the workflow changes and text changes to the one template here

PR #860, opened 2024-11-06
2/n: Update wilms 06 module

sjspielman:
Towards #850 
Part of #852 

This PR begins the next stages of updates for the Wilms 06 module:

- For #852, I added the `RUN_EXPLORATORY` flag to the module's workflow, and I use it (so far!) to skip the 00b reference exploration notebook, and the 03 clustering notebook. 
- Even so, I also updated the 03 clustering notebook to be able to run in CI - for example, we might want such steps to run on pulls but not PRs? Any thoughts there?
- I re-rendered all the sample-specific clustering notebooks, which is the MAJORITY of changes here!

_Edit_: Fun times with GitHub interface with this one... I could split up the notebooks into a few batches across PRs? Please let me know if that would be preferable!

### DIFF
--- a/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
+++ b/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
@@ -1,12 +1,23 @@
 #!/bin/bash
 
 # This script runs the module workflow.
+# There are two variables this script will take:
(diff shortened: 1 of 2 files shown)


sjspielman:
Closing in favor of opening a new smaller PR, since this one is still _too big_!

PR #862, opened 2024-11-06
Updated wilms clustering notebooks 1/3

sjspielman:
Trying again for https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/860!

---

Towards #850 
Part of #852 

This PR continues updates for the Wilms 06 module:

- For #852, I added the `RUN_EXPLORATORY` flag to the module's workflow, and I use it (so far!) to skip the 00b reference exploration notebook, and the 03 clustering notebook. 
- Even so, I also updated the 03 clustering notebook to be able to run in CI - for example, we might want such steps to run on pulls but not PRs? Any thoughts there?
  - Highly recommend to review with hidden whitespace, since this file also got styled 
- This PR also contains re-rendered notebooks for samples 168-179. Assuming this number of HTMLs is ok for GitHub's UI (which it seems to be 🤞 !) I will submit 2 more PRs with the remaining updated notebooks.

### DIFF
--- a/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
+++ b/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
@@ -1,12 +1,23 @@
 #!/bin/bash
 
 # This script runs the module workflow.
+# There are two variables this script will take:
+# 1. The TESTING variable controls whether test data should be used.
+#  Setting TESTING=1 will use the test data.
+#  This variable is 0 by default.
+# 2. The RUN_EXPLORATORY variable controls whether optional exploratory
+#  steps that do not directly contribute to final cell type annotations
+#  should be run. Setting RUN_EXPLORATORY=1 will run those steps.
+#  This variable is 0 by default.
 #
-# USAGE:
-# bash 00_run_workflow.sh
+# Default usage:
+# ./00_run_workflow.sh
 #
-# USAGE in CI:
-# TESTING=1 bash 00_run_workflow.sh
+# Usage in CI:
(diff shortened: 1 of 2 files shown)


analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
```diff
@@ -1,16 +1,28 @@
 #!/bin/bash
 
 # This script runs the module workflow.
+# There are two variables this script will take:
+# 1. The TESTING variable controls whether test data should be used.
+#  Setting TESTING=1 will use the test data.
+#  This variable is 0 by default.
```
jaclyn-taroni, line 7: This doesn't seem accurate – it controls whether or not particular _steps are run._ What _data_ is being used is entirely controlled  by what happens to be symlinked at `data/current`, or am I missing something?

jaclyn-taroni (dismissed):
If the check passes, this seems mostly good 👍🏻  However, please update the description of `TESTING` in the shell script before merging.

analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
sjspielman: No, you're not missing something - this description needs tweaking.

sjspielman:
Moving right along!

- testing variable description fixed
- textual bug fixed (it needed a `unique` 🫠 ), and while I was there, I fixed a bunch of typos too before re-rendering
- notebooks re-rendered

jaclyn-taroni (approved):
👍🏻

Issue #611, opened 2024-07-15
Exploratory analysis notebook for SEACells


jashapiro:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#362

### Describe the goals of the changes to the analysis module.

Following #610, we will want to look at one or more metacell results to see how well the method is performing on our data sets. To do this, we will create a notebook (Jupyter most likely) that will include calculating statistics such as compactness and separation to look at how well the metacells cover the original cell types. This will also include a visualization of the metacells overlaid on the original cells in a UMAP. 

This notebook will also include a summary computational requirements that were required for calculating the metacells.

The ultimate goal here is to allow comparisons among different metacell methods, but at this stage we will still be restricted to just the one, though perhaps with a range of input parameters.

### What will your pull request contain?

A Jupyter notebook, as well as a script to allow rendering the notebook with different input AnnData objects. 

### Will you require additional software beyond what is already in the analysis module?

Not expected

### Will you require different computational resources beyond what the analysis module already uses?

Unknown at this time.

### If known, when do you expect to file the pull request?

By July 26

PR #868, opened 2024-11-07
Add initial evaluation notebook to metacells module

jashapiro:
Here I am adding a small notebook with some initial plots to evaluate the SEACells output. The main purpose here is not so much to give a full accounting of performance, but to get the infrastructure in place for further updates. 

The notebook is written in a mix of Python and R so that I can use tools from both to make plots, depending which seems more convenient. So far it seems to be working pretty well locally, but I do have some concerns about the test environment and how well this will work for others, which is part of why I am submitting it in this state. 

Future updates will likely include the randomization noted in the final paragraphs, as well as additional plots of compactness and separation, as discussed in #611 a while ago!


jashapiro:
I can't seem to solve the issue we are seeing in CI. There is some strange interaction between matplotlib and reticulate that seems to only be coming up in conda environments, and I don't quite know what to do about it. I get various _different_ errors depending on whether I am running with `Rscript -e "rmarkdown::render('notebook.rmd')"` or if I am interactively running the notebook, and while in interactive mode I seem to be able to proceed while ignoring the errors, I can't do that in a script-based rendering. 

I think I am going to have to abandon the idea of doing this in an R/Python hybrid environment and instead use either a plain Jupyter notebook, or perhaps try a python-only quarto notebook. I'm leaving this up as a draft, but I will work on a different approach next.

jaclyn-taroni:
@jashapiro - I do not plan to review this while it is a draft. If you would like a review at this stage and/or when you're ready for a review, please mention me. Thank you!

PR #869, opened 2024-11-08
Add dockerfile for metacell analysis

jashapiro:
Since something seems to be going wrong with the environment in #868, I thought I would take an alternative approach and also work on the dockerfile for the metacell module. 

Here I am bringing over the environment files from that branch and activating building the dockerfile. When this goes in, I can see whether the dockerfile helps solve the issue in #868, though I might also like to solve it within the GHA context, or at least verify that the environments work on a linux system.


sjspielman (approved):
👍

PR #871, opened 2024-11-08
Update calculate_stability usage

### LINKED ISSUE #870: Update rOpenScPCA::calculate_stability usage
Following https://github.com/AlexsLemonade/rOpenScPCA/pull/4 , we'll need to update these lines from `clusID.df$cluster` to just `clusID.df`

https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/410e06a13e40928ba359f21553e9bbab3f42471f/analyses/cell-type-ETP-ALL-03/scripts/05_cluster_evaluation.R#L19

https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/410e06a13e40928ba359f21553e9bbab3f42471f/analyses/cell-type-nonETP-ALL-03/scripts/05_cluster_evaluation.R#L20

sjspielman:
Closes #870 
Pretty much does what it says, following https://github.com/AlexsLemonade/rOpenScPCA/pull/4!
Files were styled by pre-commit, too.

### DIFF
--- a/analyses/cell-type-ETP-ALL-03/renv.lock
+++ b/analyses/cell-type-ETP-ALL-03/renv.lock
@@ -2372,7 +2372,7 @@
       "RemoteHost": "api.github.com",
       "RemoteUsername": "AlexsLemonade",
       "RemoteRepo": "rOpenScPCA",
-      "RemoteRef": "f71a8191130fa543e6506c73c7b62ffa55e8ba3f",
+      "RemoteRef": "main",
       "RemoteSha": "f71a8191130fa543e6506c73c7b62ffa55e8ba3f",
       "Requirements": [
         "BiocParallel",
--- a/analyses/cell-type-ETP-ALL-03/renv.lock
+++ b/analyses/cell-type-ETP-ALL-03/renv.lock
@@ -211,6 +211,13 @@
       ],
       "Hash": "ed5a67b4c7ae01edd46809f1387adbee"
     },
+    "HiddenMarkov": {
+      "Package": "HiddenMarkov",
+      "Version": "1.8-13",
+      "Source": "Repository",
+      "Repository": "RSPM",
+      "Hash": "26d4e106f5429773167ccf8f55581cdd"
+    },
     "IRanges": {
       "Package": "IRanges",
       "Version": "2.38.1",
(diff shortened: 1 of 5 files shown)


sjspielman:
I did retrigger this run after the `rOpenScPCA` update went in, but maybe the environment was cached.

sjspielman:
Environment indeed seems to be cached. Let's just re-trigger after the weekend!

sjspielman:
Still mostly cached, but there was this line which makes me think `rOpenScPCA` was not cached, in which case I do not really understand why the updated `rOpenScPCA` wasn't retrieved. @jashapiro, do you have any ideas?

> Installing rOpenScPCA ...      OK [built from source and cached in 20s]

sjspielman:
> Did you update renv.lock? It does record the commit hash, so I would expect that it is using that to retrieve the correct version.

Ah, I had assumed since we didn't do release/tag (though, we probably need to adopt a `development` branch over there...) that wouldn't change. Perhaps bad assumption..

jashapiro:
> we probably need to adopt a `development` branch over there...  

I don't think that we do, at least not at this time.

sjspielman:
When I snapshot'ed `analyses/cell-type-ETP-ALL-03/renv.lock`, a _bunch_ of new packages got added. I guess it just had missed a snapshot in there at some point during development, but not enough to affect CI...?

analyses/cell-type-ETP-ALL-03/renv.lock
```diff
@@ -2076,8 +2372,8 @@
       "RemoteHost": "api.github.com",
       "RemoteUsername": "AlexsLemonade",
       "RemoteRepo": "rOpenScPCA",
-      "RemoteRef": "main",
-      "RemoteSha": "fc784446f8d86b072e6f7f67287adfff598f4911",
+      "RemoteRef": "f71a8191130fa543e6506c73c7b62ffa55e8ba3f",
```
jashapiro, line 2375: Did you make this update manually, installing the commit directly? I would have expected the `RemoteRef` to stay as `main`, which will enable future updates with just `renv::update("rOpenScPCA")`. Otherwise we have to keep track of the hashes much more manually, which is annoying.

sjspielman: I did `renv::install("rOpenScPCA@......")`. Will go again with `update`

jashapiro: You should be able to just change these lines manually.
```suggestion
      "RemoteRef": "main",
```

jashapiro (approved):
LGTM

PR #872, opened 2024-11-11
Wilms-06 update CNV methods: copykat (1/3)

sjspielman:
Towards #810 

This is the first of three PRs which update the CNV-related steps of the `cell-type-wilms-tumor-06` module. This PR both updates the `05_copykat.R` script and `05_copykat_exploration.Rmd` notebook, and more generally updates code that runs the CNV methods:
 
 - I have replaced `scripts/explore-cnv-methods.R` with a shell script version `scripts/explore-cnv-methods.sh` which performs the same steps. The `00_run_workflow.sh` script now also uses this shell script. 
	- There is one main different in this new shell script - it takes an argument `THREADS` which can be passed into the `copyKAT` script.
- I updated the `05_copykat.R` script as follows:
	-  I updated the usage comments to capture all arguments 
	- I made sure we actually _use the threshold_ for identifying normal cells to use in the reference via the new argument `score_threshold`. 
	- I simplified file naming and saving code so there is no further need for `setwd()`. When I originally reviewed this code, I had requested to save images as PNGs in case we wanted them under version control, but since copyKAT isn't getting used in the end, I went ahead and simplified code by keeping these as JPEGs. This means I was also able to remove `jpeg` from `renv` (this won't prevent the current Docker image from running code though, since a dependency was not _added_).
	- Speaking of `renv`, the `renv/activate.R` file had gotten out of sync with `renv.lock` - they were using different `renv` versions. They now are both at `renv==1.0.7`.
- I updated the `05_copykat_exploration.Rmd`  notebook as follows:
	- I updated copyKAT image file paths with `jpeg` extension, since the script no longer converts them to png
	- Otherwise, the file was styled (again, turn off whitespace!)
- I updated `scripts/README.md` to reflect changes, and remove some text which was now outdated after the updated label transfer.


jaclyn-taroni (approved):
👍🏻

PR #873, opened 2024-11-11
Wilms-06 update CNV methods: infercnv (2/3)

sjspielman:
Towards #810 

This is the secodn of three PRs which update the CNV-related steps of the `cell-type-wilms-tumor-06` module. This PR both updates the `06_copykat.R` script and `06_infercnv_exploration.Rmd` notebook.
-  I updated the `06_infercnv.R` script as follows:
	- I made sure we actually _use the threshold_ for identifying normal cells to use in the reference via the new argument `score_threshold`
	- I added a flag which can be used in CI testing. When run in CI, this script will attempt to use a reference set of normal cells, but we can't expect that the testing data actually has normal cells. So, turning on this flag will ensure no reference is used when the test data is used.
- I fixed a small bug in `06_infercnv_exploration.Rmd` 

----

The third PR will be filed once #872 is merged, since it will update code in the `explore-cnv-methods.sh` script which that PR established to use the new testing flag in `06_infercnv.R` added here.

In addition, the third PR _may_ need to somehow update the rendered `html` notebooks for the exploratory copykat and infercnv code which are stored, for the 5 explored samples in `notebook/{sample_id}/`. I cannot fully render these notebooks (though I confirmed they run with test data!) due to compute limitations, so I see these as the best two ways to handle them:
- We can actually potentially remove these notebooks from the repo and send them to `results` instead, since they are not part of the actual annotation pipeline. In this case, I would take those steps in the third PR.
- We can probably ask Maud to regenerate these notebooks once we merge all this code into `main`. In this case, I would file an issue for followup.


analyses/cell-type-wilms-tumor-06/scripts/06_infercnv.R
```diff
@@ -138,8 +167,6 @@ counts <- GetAssayData(object = srat, assay = "RNA", layer = "counts")
 annot_df <- data.frame(condition = as.character(srat$fetal_kidney_predicted.compartment))
 rownames(annot_df) <- colnames(counts)
 
-
-# We only run the CNV HMM prediction model if the reference is "both"
```
sjspielman, line 142: Note that this comment was removed because it is not accurate.

jaclyn-taroni (approved):
👍🏻

PR #874, opened 2024-11-12
hello-clusters notebook: Perform and evaluate clustering

### LINKED ISSUE #796: Add notebook to demonstrate clustering with rOpenScPCA

sjspielman:
Closes #796 

This PR adds the first notebook to the `hello-clusters` module. A first round of high-level review might be good to start with for comments on organization (including within the notebook, and the notebook's location itself), content, and scope. Or, go for a fuller review if you think it's reasonable enough already!
Here is the rendered notebook to help with review: 
[01_perform-evaluate-clustering.nb.html.zip](https://github.com/user-attachments/files/17721023/01_perform-evaluate-clustering.nb.html.zip)

In addition to the notebook, I updated the module README and activated the module workflow for testing this notebook.


analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +58,378 @@ data_dir <- file.path(repository_base, "data", "current")
...
+# Print resulting table
+cluster_results
+```
+
+
+
+### Clustering on a Seurat object
+
+When performing clustering on a Seurat object, `rOpenScPCA::calculate_clusters` assumes that the principal components are stored in a reduced dimension slot named `"pca"`.
+If your Seurat has a different name for the principal components, you can provide the name with the argument `pc_name`.
```
jashapiro, line 161:
```suggestion
If your Seurat object has a different name for the principal components, you can provide the name with the argument `pc_name`.
```

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +58,378 @@ data_dir <- file.path(repository_base, "data", "current")
...
+
+
+## Evaluate clustering results
+
+This section demonstrates how to use several functions for evaluating cluster quality and reliability. 
+All examples will use the SCE object for demonstration, but a Seurat object can be provided instead.
+
+All functions presented in this section take the following required arguments:
+
+* An SCE or Seurat object that contains principal components
```
jashapiro, line 186: Can't we also just pass in the PCA matrix?

jashapiro (commented):
High level comments:

I think the overall content is fine here, but I found the integration of SCE and Seurat somewhat confusing/distracting. I think I would arrange it to do all of the content with the PCA matrix directly (assuming that I am correct about the organization), and then have separate sections about the considerations for input and output with SCE or Seurat objects. 

My other main thought is that the statistics here are all kind of hard to interpret on their own. I would probably couch this notebook as a demonstration of the evaluation functions rather than an actual evaluation. The real evaluation would come with some comparisons among different clustering parameters, which I would expect in a later notebook. 

I also want to quibble with your repeated strong recommendation of setting the seed for every function with a random component. Since R uses a global RNG, this can be a bit of a dangerous practice. It is often better to set the seed once in a notebook, rather than continuously resetting it. In this case it may not matter, but if there is any looping (for example if bootstrapping and calculating calculating statistics on each round) you can end up causing trouble.

jashapiro:
> then have separate sections about the considerations for input and output with SCE or Seurat objects.

A secondary thought on this component is that I think we probably want to cover how to use existing cluster assignments (particularly for Seurat) for the silhouette width and purity. It seems likely that people will use default Seurat functions to calculate clusters and then may want to look at those statistics for the Seurat-calculated clusters. Similarly, people may want to look at the default clusters that our SCE objects include.

sjspielman:
> A secondary thought on this component is that I think we probably want to cover how to use existing cluster assignments (particularly for Seurat) for the silhouette width and purity.

Great call, incoming.

> My other main thought is that the statistics here are all kind of hard to interpret on their own. I would probably couch this notebook as a demonstration of the evaluation functions rather than an actual evaluation. 

I agree they are not really the most informative without a full evaluation/comparison. Would you suggest removing plots altogether here then and just focusing on the function usage?

sjspielman:
This is now ready for another look! Changes broadly include:
- Code now uses a pca matrix throughout, except for the new section towards the end that shows how to use an object
- A section for evaluating existing cluster results from Seurat or the ScPCA ones as examples
- The Seurat object used throughout the examples is now generated via a Seurat pipeline from the raw counts, which is more realistic for how contributors would be using a Seurat object (based on our experience so far). I figure in the future, we can replace the conversion code here with a function we add to rOpenScPCA for doing the conversion.
- I pitched evaluation more as "calculating QC metrics" rather than evaluating per se

Here is the current version of the rendered notebook:
[01_perform-evaluate-clustering.nb.html.zip](https://github.com/user-attachments/files/17780583/01_perform-evaluate-clustering.nb.html.zip)

sjspielman:
Small question here: I'm on the fence for keeping `params$seed` vs hardcoding a seed in there, which would be "visually more appealing" in the html. Do you have a thought?

jashapiro:
> Here is the current version of the rendered notebook:
> [01_perform-evaluate-clustering.nb.html.zip](https://github.com/user-attachments/files/17780583/01_perform-evaluate-clustering.nb.html.zip)

This version doesn't have any results/plots in it, and the version in the repo is out of date. Can you update the rendered version in the repo?

jashapiro:
> Small question here: I'm on the fence for keeping `params$seed` vs hardcoding a seed in there, which would be "visually more appealing" in the html. Do you have a thought?

Hardcoding the seed here seems fine.

sjspielman:
> This version doesn't have any results/plots in it, and the version in the repo is out of date. Can you update the rendered version in the repo?

Boo, sorry, I'll regenerate.
That said, I did remove the plots which is tentatively what I took your review to mean (see https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/874#issuecomment-2477078917). But, can easily restore!

sjspielman:
Better! [01_perform-evaluate-clustering.nb.html.zip](https://github.com/user-attachments/files/17780736/01_perform-evaluate-clustering.nb.html.zip)

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+)
+
+stability_results_leiden
+```
+
+
+
+##  Calculate QC metrics on existing clusters
+
+Even if you perform clustering with other tools, you can still use the `rOpenScPCA` functions to evaluate your clusters.
```
jashapiro, line 297:
```suggestion
If you already have clustering results calculated with other tools, you can still use the `rOpenScPCA` functions to evaluate your clusters.
```

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+
+In this section, we'll present examples of how you can calculate the silhouette width, neighborhood purity, and cluster stability from existing cluster assignments.
+
+### Evaluating Seurat clusters
+
+If you are analyzing your data with a Seurat pipeline that includes calculating clusters, you can use `rOpenScPCA` to evaluate them.
+We previously created a Seurat object called `seurat_obj` with clusters:
```
jashapiro, line 304: Create the object here, I think.

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+
+```{r seurat purity}
+seurat_purity_df <- calculate_purity(
+  seurat_pca_matrix,
+  seurat_cluster_df
+)
+seurat_purity_df
+```
+
+We do not recommend using `rOpenScPCA::calculate_stability()` on Seurat clusters due to differences in the underlying clustering approach between Seurat and the `bluster` package which `rOpenScPCA` uses.
```
jashapiro, line 347: 👍🏼 I was all set from the intro to this part to say you needed a note about this.

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+
+### Evaluating ScPCA clusters
+
+ScPCA cell metadata already contains a column called `cluster` with results from an automated clustering. 
+These clusters were calculated using `bluster`, the same tool that `rOpenScPCA` uses, with the following parameters:
+
+* The Louvain algorithm
+* Jaccard weighting
+* 20 nearest neighbors
```
jashapiro, line 356: I would actually show that we have these parameters stored in the object, and use those to retrieve the values:

```
metadata(sce)$cluster_algorithm
metadata(sce)$cluster_weighting
metadata(sce)$cluster_nn
```

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+
+In this example, we'll show how to use the cluster evaluation functions on these clusters.
+
+To begin, we'll prepare a data frame with two columns: `cell_id` containing cell barcodes, and `cluster` containing the cluster identities.
+
+```{r prepare scpca data frame}
+scpca_cluster_df <- data.frame(
+  cell_id = colnames(sce),
+  cluster = colData(sce)$cluster
```
jashapiro, line 365: SCE objects are smart; no need for `colData` here!
```suggestion
  cluster = sce$cluster
```

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+  resolution = 0.5,
+  nn = 15
+)
+
+stability_results_leiden
+```
+
+
+
+##  Calculate QC metrics on existing clusters
```
jashapiro, line 295: I would move this to after the object section and demonstrate these using the objects themselves.

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+reducedDimNames(sce) <- c("PCA_matrix", "UMAP")
+reducedDimNames(sce)
+
+# Calculate clusters from an SCE object using default parameters
+cluster_results_df <- calculate_clusters(
+  sce,
+  pc_name = "PCA_matrix"
+)
+cluster_results_df
+```
```
jashapiro, line 438: This feels like a detail we don't need to demonstrate. I would remove this section, or at least the code for it.

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+These automatic clusters were not evaluated, and their parameters were not optimized for any given library.
+
+To avoid ambiguity between the existing and new clustering results, you may want to rename incoming columns.
+
+We'll therefore begin by adding the prefix `ropenscpca_` to all column names in the `cluster_results_df` data frame before adding them to a given object.
+
+```{r rename columns}
+# First, rename columns in `cluster_results_df` to avoid ambiguity
+cluster_results_renamed_df <- cluster_results_df |>
+  # add the prefix openscpca_ to all columns
```
jashapiro, line 458:
```suggestion
  # add the prefix "ropenscpca_" to all columns
```

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+  dplyr::rename_with(~ paste0("ropenscpca_", .x))
```
jashapiro, line 459: Modern R
```suggestion
  dplyr::rename_with(\(x) paste0("ropenscpca_", x))
```

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+colData(sce)
+```
+
+### Saving results to a Seurat object 
+
+Follow these steps save the clustering results to a Seurat object's cell metadata:
+
+```{r add to seurat}
+# Add the cluster result data frame to the Seurat object metadata
+seurat_obj <- AddMetaData(seurat_obj, cluster_results_renamed_df)
```
jashapiro, line 497: Does Seurat assume that the data frame is in the same order? If so, we might want to caution about that. 

In the interests of simplicity, you might also want to show making that assumption above, in which case you can first check that it is true with `all(colnames(sce) == cluster_results_df$cell_id)` or something like that, then just add the clusers more simply with `scr$ropenscpca_clusters <- cluster_results_df$cluster`. I don't think we really need to show adding all of the cluster parameters?

jashapiro (commented):
This looks pretty good. 

I had a few relatively small comments, with the most recurrent one (I stopped after a couple) about printing out large tables of results, which I think we should probably avoid. I wasn't actually saying not to include plots at all; I do think they are useful for showing the range of each statistic. I would just keep the plotting code as simple as possible, which probably means not trying to include median lines, etc. 

I also suggest moving all the `Seurat` content together, rather than building the object then abandoning it for a while. I'd also show the "using previous" results in that context; in a section where you are already working with Seurat or SCE objects.

Finally, I think we can simplify some of the end where you are adding results to an object to just show adding a single column; the renaming a table and joining seems like it is straying from the main goal of the notebook.

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+We'll also establish a corresponding processed Seurat object from its raw counts that we'll use for some examples.
+
+```{r read data}
+# Read the SCE file
+sce <- readRDS(input_sce_file)
+```
+
+```{r create process seurat, message = FALSE}
+# Convert to a Seurat object
+seurat_obj <- CreateSeuratObject(counts = counts(sce), assay = "RNA")
```
sjspielman, line 86: This needs to be updated with `rOpenScPCA::sce_to_seurat()` once https://github.com/AlexsLemonade/rOpenScPCA/pull/15 goes in.

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +62,443 @@ data_dir <- file.path(repository_base, "data", "current")
...
+  SCTransform() |>
+  RunPCA() |>
+  FindNeighbors() |>
+  FindClusters()
 ```
 
-## Analysis content
 
-Organize the remainder of your content into sections and subsections as appropriate for your analysis.
+We'll also extract its PCA matrix, which we'll use to demonstrate how to calculate and evaluate clusters.
```
sjspielman, line 96:
```suggestion
For the initial cluster calculations and evaluations, we will use the PCA matrix extracted from the SCE object.
It's also possible to use an SCE object or a Seurat object directly, which we will demonstrate later.
```

sjspielman:
This has now be revived and refreshed based on reviews and ready for another look! The code has generally be simplified and rearranged according to reviews, and I restored some plots (with a green fill that I accept might not survive).
[01_perform-evaluate-clustering.nb.html.zip](https://github.com/user-attachments/files/18152717/01_perform-evaluate-clustering.nb.html.zip)

Note that I do want to eventually link to the forthcoming Seurat module https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/945 in the part where I actually use `sce_to_seurat()`, but that will probably be a separate PR due to relative timing?

jashapiro (approved):
I think this looks good to go in. I will insist on changing colors only to make sure you are using different colors for silhouette, purity, and stability. I don't care what they are, as long as they are distinct.

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +59,434 @@ data_dir <- file.path(repository_base, "data", "current")
...
+head(purity_results)
+```
+
+
+We can visualize these results by plotting purity clusters as violin plots, for example:
+
+```{r violin purity}
+ggplot(purity_results) +
+  aes(x = cluster, y = purity) +
+  geom_violin(fill = "darkolivegreen3") +
```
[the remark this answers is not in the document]
sjspielman: > I do care that when you are plotting different metrics you should use different colors.

Now _that_ I can do 🎨

analyses/hello-clusters/01_perform-evaluate-clustering.Rmd
```diff
@@ -49,32 +59,434 @@ data_dir <- file.path(repository_base, "data", "current")
...
+Generally speaking, we only recommend evaluating clusters with `rOpenScPCA::calculate_stability()` if you know the original parameters used.
+
+```{r scpca stability}
+scpca_stability_df <- rOpenScPCA::calculate_stability(
+  sce,
+  scpca_cluster_df,
+  # provide ScPCA clustering parameters
+  algorithm = "louvain",
+  weighting = "jaccard",
+  nn = 20
```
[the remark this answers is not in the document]
sjspielman: I feel fairly sure the answer here isn't "let's add another dependency!", but I did think this was useful to know about https://cran.r-project.org/web/packages/strex/vignettes/argument-matching.html

PR #876, opened 2024-11-12
Wilms-06 update CNV methods (3/3), and run module in CI

sjspielman:
This PR follows #872 and #873 to update code for CNV methods and makes the following changes (_edited with additional changes found during debugging_:

- Adds the a `TESTING` variable to `scripts/explore-cnv-methods.sh`
  - Both that script and `00_run_workflow.sh` now create a `$test_string` variable to use when calling `06_infercnv.R` 
- The workflow script now has fully uncommented code to now run fully through CI!
  - Accordingly, I removed the `00_run_workflow.R` script, so now all we have is `00_run_workflow.sh`.
- `.github/workflows/run_cell-type-wilms-tumor-06.yml` now also downloads OpenScPCA metadata, which does not get downloaded when the `--samples` flag is used
- I added a `testing` param to the `04` and `07` notebooks which process all samples. We expect when running in CI, not input files for all sample IDs will not be present since only a subset of samples are processed. The new `testing` param allows files to be missing _only_ if we're testing.
  - In the `04` notebook, I also modified a `dplyr::select` statement to accommodate the fact that not all kidney compartments may be present in test data label transfer results. 
  - I regenerated these notebooks with the updated code 
- I turned off eval when testing for a couple more chunks in the `03` clustering notebook, which I caught when running all test data through the exploratory steps  

This now runs all the way through with test data locally, including with exploratory steps turned on, which are off in CI. There are some stochastic copykat errors, but that's off in CI anyways so we should be fine there. 

We do not expect this to pass CI until the sample metadata is fixed, but I will wait for then to remove draft status and request review.


sjspielman:
After some debugging, this runs fully with test data locally (including with exploratory steps turned on, which are off in CI). But, we do not expect this to pass CI until the sample metadata is fixed, but I will wait for then to remove draft status and request review.

Additional changes include:

- `.github/workflows/run_cell-type-wilms-tumor-06.yml` now also downloads OpenScPCA metadata, which does not get downloaded when the `--samples` flag is used
- I added a `testing` param to the `04` and `07` notebooks which process all samples. We expect when running in CI, not input files for all sample IDs will not be present since only a subset of samples are processed. The new `testing` param allows files to be missing _only_ if we're testing.
  - I regenerated these notebooks with the updated code 
- In the `04` notebook, I modified a `dplyr::select` statement to accommodate the fact that not all kidney compartments may be present in test data label transfer results.
- I turned off eval when testing for a couple more chunks in the `03` clustering notebook, which I caught when running all test data through the exploratory steps

sjspielman:
🎉  ready!

jaclyn-taroni:
Should we wait until this _will_ pass CI?

sjspielman:
> Edit: I'll review, but we shouldn't merge without CI.

:100:

Either way, there will still be one more PR to send this into `main`. I'm just realizing we're actually missing a `push` trigger in the workflow file to run in CI with _all_ test data. I'm going to add those lines in once this passes CI, so that once we merge it will get tested with all data before we file into `main`.

analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
```diff
@@ -39,6 +39,13 @@ data_dir="../../data/current"
 notebook_template_dir="notebook_template"
 notebook_output_dir="notebook"
 
+# Define test data string to use with 06_infercnv.R
+if [[ $TESTING -eq 1 ]]; then
```
jaclyn-taroni, line 43:
```suggestion
if [[ $IS_CI -eq 1 ]]; then
```

jaclyn-taroni: But maybe you just want to do the following at line 28 to be consistent with the other variables and shell script:

```sh
TESTING=${TESTING:-0}
```

Don't forget to change lines 136 and 174 (at least), too, if you go this way

jaclyn-taroni (commented):
One minor but important change to the shell script is required

analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
sjspielman: Ended up going with  `TESTING=${TESTING:-0}` and updating other lines accordingly.

analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
jaclyn-taroni, line 1: I no longer get the error I was seeing locally when running following the instructions in the module README. I also checked for other instances of `IS_CI` and didn't find them.

jaclyn-taroni (approved):
👍🏻

PR #904, opened 2024-11-19
Update creating-vcs.md to no longer reference selecting an Availability Zone

jaclyn-taroni:
Related to: https://github.com/AlexsLemonade/OpenScPCA-infra/pull/53

Once that's deployed, you will no longer need to select an Availability Zone when creating a Lightsail for Research instance via Service Catalog.

It probably also means ~~some of the screenshots are outdated~~ one screenshot is outdated, but I think it would need to make a material difference to re-do ~~them~~ it, and I'm not sure it meets that threshold.


sjspielman (approved):
👍 
I think if the question comes up we can update the docs, but otherwise they're probably fine.

PR #907, opened 2024-11-21
Shrink and fix simulated SCEs

jashapiro:
Related to https://github.com/AlexsLemonade/scpca-nf/issues/803, I wanted to start by trimming down the simulated data, as these results are particularly bad for the simulated data. 

I also noticed that I was trimming the cellassign_predictions table incorrectly, as that is a tibble without row names... so now I am correctly filtering based on the barcode column.

### DIFF
--- a/analyses/simulate-sce/scripts/simulate-sce.R
+++ b/analyses/simulate-sce/scripts/simulate-sce.R
@@ -134,11 +134,17 @@ simulate_sce <- function(sce, ncells, replacement_metadata, processed) {
   # reduce the cell type data matrices, if present
   if (!is.null(metadata(sce_sim)$singler_results)) {
     sim_cells <- cell_subset[cell_subset %in% rownames(metadata(sce_sim)$singler_results)]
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
LGTM

I had mixed feelings about doing this because I do think the list of genes can be useful in a handful of cases, but I do agree that saving on the overall size of the object outweighs the need to have them.

jashapiro:
> LGTM
> 
> I had mixed feelings about doing this because I do think the list of genes can be useful in a handful of cases, but I do agree that saving on the overall size of the object outweighs the need to have them. Edit: doing it in scpca-nf, but I definitely agree that having it for the simulated data makes sense.

I was kind of mixed myself, but it can be recreated if needed. In the full workflow it is a pretty small % of the data, but I also don't really see it being used directly. I feel like if people want that level of detail they are probably going to be exploring elsewhere (but we can discuss more in https://github.com/AlexsLemonade/scpca-nf/issues/803)

Issue #896, opened 2024-11-19
Update report to compare clustering results to cell types and marker gene sets in Ewing module


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#292 

### Describe the goals of the changes to the analysis module.

Following the changes in #895, we will need to fully update the second report for the clustering workflow. This report will be generated after looking at the initial clustering evaluation in `01-clustering-metrics.Rmd` and aim to look at the overlap of clusters with assigned cell types and the expression of marker gene sets across clusters. 

### What will your pull request contain?

Updates to the `02-clustering-celltypes.Rmd` report added in #895. Specifically this report should: 

- Take as input a table with clustering results for a given library and a desired set of clustering parameters to use. 
- Compare the cell types assigned with SingleR to clusters
- Plot expression of key marker gene sets across clusters using marker genes in [`references/visser-all-marker-genes.tsv`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-ewings/references/visser-all-marker-genes.tsv)

### Will you require additional software beyond what is already in the analysis module?

No 

### Will you require different computational resources beyond what the analysis module already uses?

No

allyhawkins:
After looking at some of the reports generated in #915, it's pretty easy to distinguish differences between the clustering algorithms and the resolution. I feel confident we could pick "good" parameters for clustering for each sample at least when it comes to the algorithm (mostly this is leiden with modularity) and resolution (mostly this is 0.5 or 1). From there we would need to pick a good value of k to use for assigning clusters which is less obvious. There are many instances where the metrics are similar for a range of k values. 

This is where I think the notebook described here will be helpful. We should have it take in the clustering TSV file with results for clustering across all parameters and then params that specify which algorithm and resolution to use. Then the actual notebook will look at how the cell types compare to the cluster assignments and validate expected marker gene expression in clusters (e.g., clusters with high tumor marker gene expression should be composed of tumor cells). Much of this is already present in [`02-clustering-celltypes.Rmd`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-ewings/template_notebooks/clustering-workflow/02-clustering-celltypes.Rmd) since we had it in the original clustering report but it will just need to be updated to use the TSV file and specific parameters as input. 

[...]

PR #908, opened 2024-11-22
Add workflow for evaluating clustering to Ewing's module

### LINKED ISSUE #686: Workflow for clustering cells in SCPCP000015

#292 

Once we have created a template notebook to calculate clusters with different parameters (#685), we will want to create a small workflow to run that notebook on all samples in the project. 

The PR will contain a bash script that will render the template notebook for clustering for all samples in the project. The output will be the rendered report and TSV file containing cluster assignments. 

We should also add this to the GHA for this module. 

No 

No 

_No response_

### LINKED ISSUE #897: Script for calculating clusters across parameters in Ewing module

#292 

In #895 we are performing clustering using a variety of clustering parameters. We should move the code that calculates those clusters and saves them to a table to a script that can be run outside of the report. Then the report should take the clustering results as input to generate the plots. 

A new script that can be used to look at clusters across a set of parameters. One thought is that maybe this script should live in `hello-clusters` rather than in the Ewing module. 

No

No 

_No response_

allyhawkins:

Closes #897 
Closes #686 

This PR takes the notebook that was modified in #895 and moves it into a workflow that can be run to perform and evaluate clustering across all samples in the Ewing's project (SCPCP000015). The workflow includes two steps: 

1. Run clustering across a set of parameters 
2. Render a report summarizing the clustering metrics across all parameters tested, generating one report per library

- I first moved the actual clustering out of the notebook and into its own script. Now the script takes in different parameter values to test and an SCE object and outputs a TSV file with the cluster results from all parameters tested. In the script I put in arguments for each louvain, leiden-cpm, and leiden-modularity to specify running each of them, but I could also be convinced to remove those arguments and just run those by default? 
- The notebook `01-clustering-metrics.Rmd` now reads in this TSV file and creates the plots looking at individual metrics. 
- I added a `evaluate-clusters.sh` workflow that has been added to CI and I tested locally with the test data. This workflow runs the clustering script and renders the clustering report for all samples in the project. 
- I updated the necessary README files including the main README to describe what is currently in the workflow. 

Yes. Next I am going to work on the second notebook in the workflow (see #896). This notebook will take the clustering results TSV and a set of parameters to look at and compare clusters to cell types and marker gene set scores. I'm still thinking about exactly how this notebook will look and where it will fit into the workflow, but the idea is to be able to look at the output from the metrics report rendered and choose a set of parameters to narrow in on and look at those clusters in a biological context. 

Should be able to run this locally without any issues. 

Just a note that the actual workflow itself has the exact same logic as the `aucell-singler-annotation.sh` and a lot of the same code. The biggest code changes that will need to be reviewed is the clustering script since that is brand new. 


analyses/cell-type-ewings/README.md
```diff
@@ -118,6 +118,69 @@ The `gene-set-scores.tsv` file contains the scores (mean and sum) for all genes
...
+The `evaluate-clusters.sh` workflow can be used to perform clustering of all Ewing sarcoma samples from SCPCP000015.
+The workflow can be run using the following command:
+
+```sh
+./evaluate-clusters.sh
+```
+
+### Input files 
+
+The `evaluate-clusters.sh` workflow requires the processed `SingleCellExperiment` objects (`_processed.rds`) from SCPCP0000015.
```
sjspielman, line 145:
```suggestion
The `evaluate-clusters.sh` workflow requires the processed `SingleCellExperiment` objects (`_processed.rds`) from SCPCP000015.
```

analyses/cell-type-ewings/README.md
```diff
@@ -118,6 +118,69 @@ The `gene-set-scores.tsv` file contains the scores (mean and sum) for all genes
 The `aucell-singler-annotation.sh` uses 4 CPUs and can be run locally on a laptop or on a virtual computer on Lightsail for Research.
 Note that `SingleR` does take up to 1 hour to run with 4 CPUs for some samples in this project.
 
+## Clustering workflow 
+
+The clustering workflow (`evaluate-clusters.sh`) can be used to perform and evaluate clusters across a range of parameters. 
+For all processed `SingleCellExperiment` objects that are part of `SCPCP000015`, clustering is performed using the following options: 
+
+- Algorithms: Louvain, Leiden with CPM, and Leiden with modularity
```
sjspielman, line 126: Might want to note also that it's always jaccard

analyses/cell-type-ewings/evaluate-clusters.sh
```diff
@@ -0,0 +1,80 @@
...
+# perform clustering on each library 
+for sample_id in $sample_ids; do
+
+    # make sure sample results directory exists
+    sample_results_dir="${results_dir}/${sample_id}"
+    mkdir -p $sample_results_dir
+
+    for sce_file in $data_dir/$sample_id/*_processed.rds; do
+
+    # define library ID
```
sjspielman, line 53: spacing is usually wrong for suggestions though, so maybe fix on your own
```suggestion
        # define library ID
```

analyses/cell-type-ewings/scripts/clustering-workflow/01-clustering.R
```diff
@@ -0,0 +1,166 @@
...
+  as.numeric()
+
+# Clustering -------------------------------------------------------------------
+
+# make a list of clustering options
+cluster_opt_list <- list()
+if(opt$louvain){
+  cluster_opt_list$louvain <- list(
+    algorithm = "louvain", 
+    # objective_function = "CPM", # we don't want this to be null
```
sjspielman, line 124: I am guessing this should have been uncommented?
```suggestion
    objective_function = "CPM", # we don't want this to be null 
```

analyses/cell-type-ewings/scripts/clustering-workflow/01-clustering.R
```diff
@@ -0,0 +1,166 @@
...
+
+# create output directory if it doesn't exist
+output_dir <- dirname(opt$output_file)
+fs::dir_create(output_dir)
+
+
+# read in input sce files
+sce <- readr::read_rds(opt$sce_file)
+
+# get nn and resolution as numbers
```
sjspielman, line 104: A little helper function couldn't hurt, but also I am partial to little helper functions and it's not totally necessary here. Up to you!

Either way, protip: `str_split_1` is a very helpful variant which returns a vector, not a list!

```suggestion
# get nn and resolution as numbers
split_numbers <- function(param) {
  param |>
    string::str_split_1(param, ",") |>
    as.numeric()
 }
```

analyses/cell-type-ewings/scripts/clustering-workflow/01-clustering.R
```diff
@@ -0,0 +1,166 @@
...
+    type = "character",
+    help = "Comma separated list with a range of values to use for resolution with the Leiden clustering algorithm and modularity objective function.
+      Only used with the `--leiden_mod` flag."
+  ),
+  make_option(
+    opt_str = c("--cpm_res_range"),
+    default = "0.001,0.005,0.01",
+    type = "character",
+    help = "Range of values to use for resolution with the Leiden clustering algorithm and CPM objective function.
+      Only used with the `--louvain` flag."
```
sjspielman, line 66:
```suggestion
      Only used with the `--louvain_cpm` flag."
```

sjspielman (commented):
I did a quick-end-of-friday review, but I plan to do a more careful look next Monday! I think it's probably all there now as is, but I don't 100% trust my brain at the moment to be sure of that 🥴 

> In the script I put in arguments for each louvain, leiden-cpm, and leiden-modularity to specify running each of them, but I could also be convinced to remove those arguments and just run those by default?

I definitely understand why you did this, but I think since this script is very specific to this module and this circumstance, it would be fine to rip them out. But it doesn't hurt to keep them, I think. I did think a little bit about whether there might be a different way to specify parameters and algorithms, and I couldn't come up with a better approach given how all the different parameters interact, and again since this is quite specific to your module I think it's fine.

allyhawkins:
> > In the script I put in arguments for each louvain, leiden-cpm, and leiden-modularity to specify running each of them, but I could also be convinced to remove those arguments and just run those by default?

> I definitely understand why you did this, but I think since this script is very specific to this module and this circumstance, it would be fine to rip them out. But it doesn't hurt to keep them, I think. I did think a little bit about whether there might be a different way to specify parameters and algorithms, and I couldn't come up with a better approach given how all the different parameters interact, and again since this is quite specific to your module I think it's fine.

I too was struggling with how best to specify the parameters! I decided to keep in the algorithm flags for now and figured it would be easier to remove in the future if we need to, but also will allow for flexibility if we ever want to use this script to just run one algorithm. 

I made the minor changes you found and then updated to use the helper function for reading in the list of parameters. This should be ready for another look!

sjspielman (approved):
LGTM! 

I made one suggestion to fix a bug I probably introduced with my earlier suggestion, and I have one other comment that I can't make directly here - Looking at the resulting plots in an HTML, I realize we probably want some explicit breaks in the x-axis across nn, so there's a tick/value for each parameter value. Right now ggplot2 is choosing breaks for you since it's treating nn as numeric. So, we should either add some breaks for the actual nn value, or tell R to treat it like a factor instead of numeric. But I don't think I need to see again!

Issue #887, opened 2024-11-15
Assign cell ontology identifiers to cell types in CellAssign reference


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#853 

### Describe the goals of the changes to the analysis module.

Before we can identify any consensus cell types between SingleR and CellAssign, we need to ensure both methods have an associated cell ontology id that can be used. We have that for SingleR, but we don't have that for CellAssign. The first step in this module will be updating our reference from PanglaoDB to include a column with cell ontology ID. Then we can incorporate these labels into the SCE objects. 

### What will your pull request contain?

I plan on creating a table that matches cell types found in the [PanglaoDB reference](https://github.com/AlexsLemonade/scpca-nf/blob/main/references/PanglaoDB_markers_2020-03-27.tsv) to [cell ontology IDs](https://www.ebi.ac.uk/ols4/ontologies/cl). 

I think the best way to do this would be to go organ by organ so that all cell types in a single organ are being labeled in a single PR. Hopefully this will ease the review burden. 

### Will you require additional software beyond what is already in the analysis module?

No software yet since we are just assigning labels. 

### If known, when do you expect to file the pull request?

This will be a series of PRs that assigns cell type labels one organ at a time (or group of related organs depending on the number of cell types in that organ).

PR #909, opened 2024-11-22
Initial assignment of cell ontology IDs to panglao cell types

allyhawkins:

Related to #887 

Here I am starting the process of assigning cell ontology IDs to the cell types present in the Panglao reference we use when running `CellAssign` as part of `scpca-nf`. This initial PR does some preliminary R setup and adds a script to assign ontology IDs to cell types that have exact matches. 

- I created an R project for this module so that file is being added here. This also includes additions to the lock file. 
- I wrote a script that takes as input the [reference file from PanglaoDB](https://github.com/AlexsLemonade/scpca-nf/blob/main/references/PanglaoDB_markers_2020-03-27.tsv) and returns a TSV file with the ontology ID, human readable value and original value in the Panglao ref file for all cell types. 
- Before doing any matching I did do some minor modifying of the original cell type names. I made everything lower case since that's how the names appear in CL (with the exception of B and T cells which have a capital B and T) and I also made everything singular. All the cell types either ended with "cells" (endothelial cells vs. endothelial cell) or ended in s (neurons vs. neuron). I looked at all the cell types to make sure that none of the ones listed actually end in "s" as part of the cell type name. 
- The output of this script is a table that has ontology IDs for any matches and NA for any of the ones that don't match. I am thinking we can then fill out this table and replace any of the NAs with the ID we feel is most appropriate. That part needs to be done manually so the output of this script is going to get modified outside of running this script. To account for us making changes to the table manually, I added a check for if the file already exists. If that's the case then any cell types that are in the existing file are removed prior to joining with the CL labels. Then they are added back to the table. This means we can manually fill in an ontology ID and then still re-run this script and not lose that information. I don't know how often we would actually re-run this script but it seemed easy to implement vs. accidentally over writing something in the future. 

I did want to point out that currently the actual reference file from Panglao is not in the repo because it's ~ 1000 KB and we have a 200 KB limit on TSV files. How do we want to proceed here? We probably don't really need it and I could just make a list of the cell types and save as a text file to read in and point to where the file lives in [`scpca-nf`](https://github.com/AlexsLemonade/scpca-nf/blob/main/references/PanglaoDB_markers_2020-03-27.tsv) or we could just make an exception for this file? 

Yes. After assigning labels this way there are 92 cell types that will need to be manually assigned. I'm thinking I'll break this up into ~ 4-5 PRs and do 20-25 cell types at a time. 

Any packages needed to run this script have been recorded in the lock file. 

#### Are there particular areas you'd like reviewers to have a close look at?

I first want to get some feedback on the overall approach here and make sure we are okay with the decisions I made in the script. Are there things you would change about the overall setup here? Once we are on the same page regarding building this new file with the ontology IDs and the use of the script I'll add to the README and document this process. We can do that in this PR or a new PR. 

Any thoughts on how to handle storing the large Panglao file? 

### DIFF
--- a/analyses/cell-type-consensus/scripts/prepare-cell-type-ontologies.R
+++ b/analyses/cell-type-consensus/scripts/prepare-cell-type-ontologies.R
@@ -1,0 +1,85 @@
+#!/usr/bin/env Rscript
+
+# This script is used to match cell types present in PanglaoDB to cell ontology IDs
+# The output is a TSV file that contains `ontology_id`, `human_readable_value`, and `panglao_cell_type`
+
+# If the output file already exists any cell types with assigned ontology IDs will not be modified 
+# Only cell types without ontology IDs will be used when matching to avoid any NA replacements 
+
+# File paths -------------------------------------------------------------------
+module_base <- rprojroot::find_root(rprojroot::is_renv_project)
+
+# read in original ref file 
+ref_file <- file.path(module_base, "references", "PanglaoDB_markers_2020-03-27.tsv")
+ref_df <- readr::read_tsv(ref_file)
+
+# cell ontology ref file 
+ontology_output_file <- file.path(module_base, "references", "panglao-cell-type-ontologies.tsv")
+
+# Prep ontology terms ----------------------------------------------------------
+
+# get uberon ontology terms and ids
+ol <- rols::Ontologies()
+cell_ontology <- ol[["cl"]]
+terms <- rols::Terms(cell_ontology)
+labels <- rols::termLabel(terms) 
+
+# data frame of id and human readable value 
+label_df <- data.frame(
+  ontology_id = names(labels),
+  human_readable_value = labels
+)
+
+# Get existing ontology assignments --------------------------------------------
+
+# check if ontology ID assignments already exist 
+if(file.exists(ontology_output_file)){
+  
+  # remove any cell types without an assigned ontology
+  existing_ontology_df <- readr::read_tsv(ontology_output_file) |> 
+    dplyr::filter(!is.na(ontology_id))
+  
+  # get a list of cell types with existing ontology IDs 
+  existing_ontology_cell_types <- existing_ontology_df$panglao_cell_type
+  
+} else {
+  # if no output file just make sure these variables exist 
+  existing_ontology_df <- NULL
+  existing_ontology_cell_types <- NULL
+}
+
+# Match ontology terms ---------------------------------------------------------
+
+# make human readable values for all cell types present in Panglao reference file 
+cell_type_df <- ref_df |>
+  dplyr::select(panglao_cell_type = `cell type`) |>
+  # remove any NAs
+  tidyr::drop_na(panglao_cell_type) |>
+  # add a column for joining
+  # generally cell type terms are in lower case and singular
+  dplyr::mutate(human_readable_value = tolower(panglao_cell_type) |> 
+                  # make everything singular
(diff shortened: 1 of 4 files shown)


analyses/cell-type-consensus/scripts/prepare-cell-type-ontologies.R
```diff
@@ -0,0 +1,85 @@
...
+# The output is a TSV file that contains `ontology_id`, `human_readable_value`, and `panglao_cell_type`
+
+# If the output file already exists any cell types with assigned ontology IDs will not be modified 
+# Only cell types without ontology IDs will be used when matching to avoid any NA replacements 
+
+# File paths -------------------------------------------------------------------
+module_base <- rprojroot::find_root(rprojroot::is_renv_project)
+
+# read in original ref file 
+ref_file <- file.path(module_base, "references", "PanglaoDB_markers_2020-03-27.tsv")
```
jaclyn-taroni, line 13: A few thoughts:

- Do we want to include a data download script that grabs this from `scpca-nf` (provided I interpreted your comment I include below correctly) in this module?
- Should we explicitly ignore this file in this module?

I think that's how I'd address this concern:

> I did want to point out that currently the actual reference file from Panglao is not in the repo because it's ~ 1000 KB and we have a 200 KB limit on TSV files. How do we want to proceed here? We probably don't really need it and I could just make a list of the cell types and save as a text file to read in and point to where the file lives in [`scpca-nf`](https://github.com/AlexsLemonade/scpca-nf/blob/main/references/PanglaoDB_markers_2020-03-27.tsv) or we could just make an exception for this file?

jaclyn-taroni (commented):
I agree with your approach. I am not approving yet because we should address what to do about the Panglao DB file that's too big before merging. I added my thoughts in an inline comment.

allyhawkins:
> A few thoughts:
> 
> * Do we want to include a data download script that grabs this from `scpca-nf` (provided I interpreted your comment I include below correctly) in this module?
> * Should we explicitly ignore this file in this module?

I liked the idea of adding a script so I did that and then also included the file in the gitignore for this module. I also added a readme for the scripts and references folders. That should at least get us started on having documentation but I imagine I will expand/ update those readmes as we continue on this process. @jaclyn-taroni this should be ready for another look.

analyses/cell-type-consensus/scripts/00-download-panglao-ref.sh
```diff
@@ -0,0 +1,19 @@
...
+
+set -euo pipefail
+
+# navigate to where script lives
+cd $(dirname "$0")
+scripts_dir=$(pwd)
+
+# define path to ref file and url
+ref_file="${scripts_dir}/../references/PanglaoDB_markers_2020-03-27.tsv"
+ref_url="https://raw.githubusercontent.com/AlexsLemonade/scpca-nf/refs/heads/main/references/PanglaoDB_markers_2020-03-27.tsv"
```
jaclyn-taroni, line 14: I suppose `main` is fine here (instead of a permalink) since the file name captures version information, and we'd probably want to use whatever the current version of this file is anyway.

jaclyn-taroni (approved):
LGTM

analyses/cell-type-consensus/scripts/00-download-panglao-ref.sh
allyhawkins: > and we'd probably want to use whatever the current version of this file is anyway.

This exact thought was my reasoning for using `main` here.

Issue #906, opened 2024-11-20
Regenerate notebooks for cell-type-wilms-tumor-06


sjspielman:
This issue is consolidating #875 as well as new todo items for `cell-type-wilms-tumor-06`. Currently, notebooks in this module are out-of-date.

### Issue #875 

> This issue tracks updating the `copyKAT` and `inferCNV` exploratory notebooks in the `cell-type-wilms-tumor-06` module, once code being developed towards #810 has been completed and merged into `main`.
> 
> See here for some additional context: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/873#issue-2649770493
> 
> >  I cannot fully render these notebooks (though I confirmed they run with test data!) due to compute limitations, so I see these as the best two ways to handle them:
> >
> >  - We can actually potentially remove these notebooks from the repo and send them to results instead, since they are not part of the actual annotation pipeline. In this case, I would take those steps in the third PR.
> > - We can probably ask Maud to regenerate these notebooks once we merge all this code into main. In this case, I would file an issue for followup.

### New items

Additional notebooks are/will be out of date due to these currently-in-progress changes:
- Spelling updates following #905. This will affect `03_clustering_exploration.Rmd`
- Changes related to #902. This will affect `00b_characterize_fetal_kidney_reference_Stewart.Rmd`, `02a_label-transfer_fetal_full_reference_Cao.Rmd`, and `02b_label-transfer_fetal_kidney_reference_Stewart.Rmd`
## Agenda

We'll want to re-run the module in full to regenerate notebooks, including all exploratory steps. It makes the most sense to wait until the next data release (currently planned for late November: https://github.com/AlexsLemonade/OpenScPCA-admin/issues/286) is out. 

That said, these diffs make reviewing hard, because GitHub UI does not even want to tell how which/how many files changed. When committing these, we'll therefore either have to do it in chunks with several PRs, or the reviewer can use GitKraken to review instead. In my experience, GitKraken will give a full diff view that GitHub will not render.

allyhawkins:
> We can actually potentially remove these notebooks from the repo and send them to results instead, since they are not part of the actual annotation pipeline. In this case, I would take those steps in the third PR.

[...]

PR #912, opened 2024-11-27
Regenerate wilms tumor notebooks: Part 1

sjspielman:
Towards #906 

_This PR may not work out due to diffs, let's see! There are not many rendered notebook changes here, but notebooks have been moved from `notebook_template` --> `notebooks` (this move will make things smoother in the future), and GitHub UI (not unreasonably) doesn't love that. If this PR can still go through, please use GitKraken to look at the diff._

This is the first of 2 PRs (but might be more if GitHub UI gets mad) to update notebooks in the `cell-type-wilms-tumor-06`. They have been updated after running the module workflow through with the 2024-11-25 data release, which did necessitate a couple small code changes.
This PR focuses on code changes and some reorganization, and the next PR will be the updated notebooks themselves, again provided we don't anger GitHub UI too much.

Changes include:

- Various README updates
- To reduce notebook confusion, we now have these two notebook directories:
  - `cnv-exploratory-notebooks` now holds the exploratory CNV notebooks (05 and 06) and has a README explaining they are not necessarily up-to-date.
  - Given how notebooks ended up being arranged/named for this module in the end, it makes more sense to collapse `notebook` and `notebook_template` into a single `notebooks` directory. All non-CNV notebooks therefore live here. This will make life easier for this module moving forward, and the original contributor does plan to keep working! _**This is all the notebook changes in this PR!!!**_ All diffs are just"renaming".
- I updated the workflow script:
  - One script (exploratory) was missing from being run in `00_run_workflow.sh` so I added it. It was also missing (boo) a Seurat library load, which I added and the script was also styled.
  - I decided to also add a variable `RUN_CNV_EXPLORATORY` to `00_run_workflow.sh` so we can more easily control whether the 05 and 06 exploratory cnv notebooks get rendered which we generally don't want/need to do. 
- In the 03 clustering notebook, I had to update some code where Seurat didn't run anymore on the new data release when there were fewer than 500 cells, which is now the case for at least one sample.


sjspielman:
No, this is a mess. Even though it will improve the module's organization to change where the notebooks live, it can't be reviewed. Will reopen without this change.

PR #919, opened 2024-11-27
Make sure cluster assignments are factors 

allyhawkins:
I was running the clustering workflow on the Ewing samples and noticed that the plot colors were on a continuous scale so the clusters were hard to distinguish on the UMAP. This is because we are now reading in a data frame with cluster assignments and R thinks it should be a double. This just converts them into factors after reading in the clustering results.


sjspielman (approved):
👍

PR #925, opened 2024-12-02
Specify number of threads used when calculating cluster stability 

allyhawkins:
This is just a small fix to specify the number of threads to use when getting cluster stability in the clustering workflow for Ewing samples. It was taking a really long time to run this, so this should hopefully help with that.


sjspielman (approved):
🚀

PR #934, opened 2024-12-05
Compare clusters to cell types and marker gene set expression

### LINKED ISSUE #896: Update report to compare clustering results to cell types and marker gene sets in Ewing module

allyhawkins:

Closes #896 

Here I am updating the second report for the clustering workflow. Originally this was all part of one report, but I pulled out the plots in this report as part of #895. The purpose of this report is to be able to help pick the best clustering parameters by looking at cell types found in each cluster and expression of cell type specific marker genes in each cluster. 

After looking at the metric reports, I saw that Leiden-modularity and Louvain were pretty similar. Leiden-CPM for the most part was pretty bad. Because of this, I'm choosing to only generate these reports using Lieden-modularity. 

- The report now has been updated to take in a single clustering algorithm, objective function and resolution. Then I have the same plots that look at cell type assignments vs. cluster assignments for all values of nn and marker gene set expression for each cluster across all values of nn. To me, the differences between nn values were very subtle when looking at metrics, so this report should help us narrow in on a value of nn to use for that sample. 
- I also added some UMAPs to the report showing the cluster assignments and the cell type assignments. I think its helpful to have the UMAP shown again here for easy reference. 
- I updated the `evaluate-clusters.sh` workflow to render this report. In doing that I also added an argument that allows you to skip running the clustering and metrics. That part is really time consuming (took me ~24 hours to run all the samples locally), so I wanted to be sure we could skip it if we ever change rendering the report which is downstream. 
 - When looking at the resolution, there were clear differences between .5, 1, and 1.5, with .5 generally having the fewest clusters and 1.5 having the most. Using 1.5 showed clusters that had lower purity and stability across the board, so I don't think we want to use that resolution. Depending on the sample .5 and 1 were the "best". However, looking at two resolutions and all nn made the report really busy, so I updated the workflow to generate a single report for .5 and another report for 1.

Yes, but probably not before the end of 2024. 

`s3://researcher-211125375652-us-east-2/cell-type-ewings/results/clustering`

New reports have been added looking at cell type vs. clusters and marker gene expression across clusters for each sample. 

Generally these samples are pretty uniform and most are tumor cells. Using a resolution of both 1 or .5 with either k=20 or k=25 gives clusters that have clear separation of marker gene expression (e.g., clusters with tumor cells express tumor cell markers and do not express immune cell markers). I think we could realistically use any of those parameter options for all samples. 

However, part of the reason we are doing this is to define tumor cell states eventually. So we expect multiple clusters of tumor cells in anticipation that these clusters have cells from different tumor cell states. But we don't really know how heterogeneous these samples are. Using a resolution of .5 produces clusters that have higher purity and stability for the most part and a fewer number of clusters. My guess is that a lot of these samples are homogenous. I think before we dive into defining tumor cell states we may want to look at these reports in depth. Another alternative will be to subcluster only tumor cells (which for most samples is most cells). But that will all be answered at a later time and is outside of the scope of this PR. 

Attached is an example of one of the reports for easy review: 

[SCPCL000822_0.5_cluster-celltype-report.html.zip](https://github.com/user-attachments/files/18028852/SCPCL000822_0.5_cluster-celltype-report.html.zip)

I ran this locally using `skip_metrics=1`

### Author checklists

_Check all those that apply._

### DIFF
--- a/analyses/cell-type-ewings/template_notebooks/clustering-workflow/02-clustering-celltypes.Rmd
+++ b/analyses/cell-type-ewings/template_notebooks/clustering-workflow/02-clustering-celltypes.Rmd
@@ -24,7 +24,7 @@ We expect that "good" clusters will be compromised of cells belonging to the sam
 Clustering was calculated using the following parameters: 
 
 - Algorithm: `r params$algorithm` 
-- Objective function: `r params$objective_function`
+- Objective function (if Leiden): `r params$objective_function`
 - Resolution: `r params$resolution`
 
 Results will be shown for all values of nearest neighbors used to perform clustering found in the cluster results file. 
@@ -92,9 +92,9 @@ clusters_df <- readr::read_tsv(params$cluster_results_file) |>
                 resolution == params$resolution) |> 
   # modify the nn column to be a character for easy plot labels
   dplyr::mutate(cluster = as.factor(cluster),
-                nn_char = glue::glue("nn-{nn}") |> 
-                  # make sure 5 comes first 
-                  forcats::fct_relevel("nn-5", after = 0)) |> 
+                # make sure 5 comes first 
+                nn_char = forcats::fct_relevel(glue::glue("nn-{nn}"), "nn-5", after = 0)
+         ) |>
   # rename barcode column for joining
   dplyr::rename(barcodes = "cell_id")
 ```
--- a/analyses/cell-type-ewings/template_notebooks/clustering-workflow/02-clustering-celltypes.Rmd
+++ b/analyses/cell-type-ewings/template_notebooks/clustering-workflow/02-clustering-celltypes.Rmd
@@ -11,14 +11,23 @@ params:
   library_id: "SCPCL000822"
   sce_file: "../../../../data/current/SCPCP000015/SCPCS000490/SCPCL000822_processed.rds"
   singler_results_file: "../../results/aucell_singler_annotation/SCPCS000490/SCPCL000822_singler-classifications.tsv"
-  cluster_results_file: "../../results/cluster_exploration/SCPCS000490/SCPCL000822_clusters.tsv"
+  cluster_results_file: "../../results/clustering/SCPCS000490/SCPCL000822_cluster-results.tsv"
   marker_genes_file: "../../references/visser-all-marker-genes.tsv"
+  algorithm: "leiden"
+  objective_function: "modularity"
+  resolution: 1
 ---
 
-TODO: Update this notebook to take as input a clustering algorithm, nn (range of nn?), resolution, and optional objective function to display in SingleR and marker gene plots
+This notebook will compare cluster assignments to cell type assignments and look at expression of tumor and normal marker genes. 
+We expect that "good" clusters will be compromised of cells belonging to the same cell type and express marker genes for a single cell type (e.g., only tumor or only immune markers). 
 
-- The clusters are then compared to the results from running `SingleR` in the `aucell-singler-annotation.sh` workflow.
-- We then look at marker gene expression across the clusters and assign each cluster to a cell type. 
+Clustering was calculated using the following parameters: 
+
+- Algorithm: `r params$algorithm` 
+- Objective function: `r params$objective_function`
+- Resolution: `r params$resolution`
+
+Results will be shown for all values of nearest neighbors used to perform clustering found in the cluster results file. 
 
 ## Setup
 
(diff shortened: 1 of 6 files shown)


analyses/cell-type-ewings/template_notebooks/clustering-workflow/02-clustering-celltypes.Rmd
```diff
@@ -104,20 +148,28 @@ cluster_classification_df <- all_cluster_results |>
...
+```{r, fig.height=15, fig.width=7}
 # get heatmap for showing cluster assignments vs. singler assignments
 cluster_celltype_heatmap(cluster_classification_df)
 ```
 
 
-## Marker gene expression {.tabset} 
+## Marker gene expression
 
 Finally, we will look at the expression of marker genes across each cluster for each value of `k`.
```
sjspielman, line 174: Might be better to refer to this as `nn` instead of `k`, since that's how you've labeled the plots

sjspielman (commented):
I left a couple small comments, this mostly looks good!

There is one thing I was thinking which may be beyond the scope of this PR, but I think it's small + helpful enough to be worth doing - Now that I have finished the training clustering exercise and along the explored metrics across _lots_ of ScPCA dataset clusterings, I have a much greater appreciation for the limitations of the clustering metrics and how important knowing the number of clusters is to contextualize them. But, we never actually report the precise number of clusters in the report with the metrics. It would be good to add this in somewhere in that report, even just printing.I still think your conclusions here are reasonable to use leiden with modularity, so I wouldn't see any need to regenerate the reports (certainly don't recalculate metrics if it takes a day!). In addition, we've seen examples of other contributors using your module (yay!!) to get analysis code and strategies, so it would be good to have this information in your clustering report should it inspire more folks in the future 😃 
(also, the report apparently needs a new line before the bullet list of metrics in the intro)

allyhawkins:
> But, we never actually report the precise number of clusters in the report with the metrics. It would be good to add this in somewhere in that report, even just printing.

This makes sense to me so I just added a small table that gets print out at the top of the report, nothing fancy but I think it does the job! 

I also turned off the clustering in the heatmap and addressed the other small comments you had. 
Here's an updated report: 
[SCPCL000822_0.5_cluster-celltype-report.html.zip](https://github.com/user-attachments/files/18030133/SCPCL000822_0.5_cluster-celltype-report.html.zip)

I updated the files on S3 also so this should be ready for another look.

sjspielman:
This report looks good to me at this point 👍 

But...

> This makes sense to me so I just added a small table that gets print out at the top of the report, nothing fancy but I think it does the job!

This table is exactly what I had in mind!! ...Unfortunately (sorry if I wasn't totally clear here!), I had it in mind for the _other_ (first) clustering notebook which actually shows the metrics, since these counts are helpful for contextualizing those metrics _in order to_ choose the 0.5&1 leiden with modularity in the first place. But again, I don't necessarily think you need to regenerate those notebooks, since these are almost certainly still good choices. Having these numbers would help drive the point home that CPM isn't great, since a lot of the CPM metrics (...when they can be calculated) look good on their surface but really just reflect "there are two clusters". Actually now that I write this, adding a statement to that notebook like "missing distributions in plots can happen when there is only 1 cluster" would probably be helpful too. 

 Is this a pain to do? Let me know!

allyhawkins:
> Is this a pain to do? Let me know!

It's not a pain to do but it would take some time since I would need to add the table and the comment in those notebooks and then regenerate them and upload to S3. I'm going to leave the table that I added here because I still think the number of clusters is helpful when looking across different values for nn. But I'll file an issue to address your comment about updating the first notebook and do that in a separate PR if that's okay?

sjspielman:
> I'm going to leave the table that I added here because I still think the number of clusters is helpful when looking across different values for nn. But I'll file an issue to address your comment about updating the first notebook and do that in a separate PR if that's okay?

All good!!!

sjspielman (approved):
🐈

PR #938, opened 2024-12-12
Add table with number of clusters to Ewing clustering report

### LINKED ISSUE #935: Update clustering metrics report for Ewing module to include number of clusters
This table is exactly what I had in mind!! ...Unfortunately (sorry if I wasn't totally clear here!), I had it in mind for the _other_ (first) clustering notebook which actually shows the metrics, since these counts are helpful for contextualizing those metrics _in order to_ choose the 0.5&1 leiden with modularity in the first place. But again, I don't necessarily think you need to regenerate those notebooks, since these are almost certainly still good choices. Having these numbers would help drive the point home that CPM isn't great, since a lot of the CPM metrics (...when they can be calculated) look good on their surface but really just reflect "there are two clusters". Actually now that I write this, adding a statement to that notebook like "missing distributions in plots can happen when there is only 1 cluster" would probably be helpful too. 

 Is this a pain to do? Let me know!

_Originally posted by @sjspielman in https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/934#issuecomment-2523414673_

We should update the first report `01-clustering-metrics.Rmd` in the Ewing module to include a table with the number of clusters for each set of parameters used.

allyhawkins:

Closes #935 

Here I'm adding a table that summarizes the number of clusters for each set of parameters used for clustering in `01-clustering-metrics.Rmd`. 

Before reporting any stats there should now be a print out of the number of clusters for all parameters used. I also added a sentence stating that if there was only one cluster there would be no distributions shown. 
The tables aren't super fancy or anything but I think they do the job and are readable. 

Yes 

`s3://researcher-211125375652-us-east-2/cell-type-ewings/results/clustering` 

Rendered reports for each sample. Note that I will re-run the workflow and update the bucket once these changes are approved. 

No changes to the results with this change other than the addition of the table. The conclusions are the same. 

I rendered the example report locally, but running the full workflow takes ~ 24 hours on a laptop. 

Here's an example of a rendered report: 
[01-clustering-metrics.html.zip](https://github.com/user-attachments/files/18117464/01-clustering-metrics.html.zip)

### Author checklists

_Check all those that apply._


allyhawkins:
The updated clustering reports have been added to S3. This already passed CI today and then new changes were added to `main` unrelated to this module so I'm cancelling the latest run of the workflow.

Issue #945, opened 2024-12-16
Create module for Seurat conversions


jashapiro:
### Please link to the GitHub Discussion for this proposed analysis.

N/A

### Describe the goals of this analysis module.

As many users may want to use Seurat objects for their analysis, it would be convenient to have a central source of converted objects. This module will serve both as a source for those converted objects and as example code for use of the `rOpenScPCA::sce_to_seurat()` function.

### What software will you require?

The primary code will be in R with the `rOpenScPCA` and `Seurat` packages.

### What will your first pull request contain?

I will break the PR into the standard initialization PR followed by a code PR with the conversion scripts.

### What computational resources will you require?

All analysis should be possible on a laptop with minimal resources (only requiring memory to hold one SCE object)

### If known, when do you expect to file the first pull request?

By 12/17

jashapiro:
Do I name this module `hello-seurat` or something a bit more descriptive like `seurat-conversion`?

sjspielman:
> Do I name this module hello-seurat or something a bit more descriptive like seurat-conversion?

- I like `hello-` as an indicator for example/demonstration modules in general, but it does feel a bit odd for this one, even though I enjoy it. I don't think you can go wrong with either, since both are very clickable for different reasons!
- _But_, to the extent we want to implement other Seurat-related functionality, `hello-seurat` does allow for increased flexibility to present items other than conversion. I can't immediately think of what this might be, though.

PR #947, opened 2024-12-16
Add example table to analysis/README

sjspielman:
Inspired by the brief conversation in #945 about naming, I took a stab at expanding `analysis/README.md` with a table indicating which modules are provided as examples of some flavor, rather than bona fide scientific analyses.

Let me know what you think!


sjspielman:
Might need to rethink this, since the `seurat-conversion` module is actually doing conversions, not necessarily demonstrating usage, which indeed we may not need to demonstrate now that we're planning on providing objects(?).

jashapiro:
> Might need to rethink this, since the `seurat-conversion` module is actually doing conversions, not necessarily demonstrating usage, which indeed we may not need to demonstrate now that we're planning on providing objects(?).

Yeah, I don't have plans to create notebooks for Seurat conversion. It might be nice to have some demonstrations if people want to work with some of the options, but I actually think this may be better provided in the docs.

sjspielman:
I think this README is still helpful, so I'll just rip out that line in the table.

analyses/README.md
```diff
@@ -2,3 +2,13 @@ This directory contains the [OpenScPCA analysis modules](https://openscpca.readt
...
+
+## Example modules
+
+Several modules are provided as examples for how to use certain frameworks in OpenScPCA:
+
+| Module name | Purpose
+|-------------|---------
+| [`hello-R`](hello-R/) | Demonstrates an example structure of an R-based module
+| [`hello-python`](hello-python/) | Demonstrates an example structure of a Python-based module
+| [`hello-cluster`](hello-cluster/) | Provides notebooks demonstrating the use of clustering and cluster evaluations functions in the [`rOpenScPCA` package](https://github.com/AlexsLemonade/rOpenScPCA/)
```
jashapiro, line 14:
```suggestion
| [`hello-clusters`](hello-clusters/) | Provides notebooks demonstrating the use of clustering and cluster evaluations functions in the [`rOpenScPCA` package](https://github.com/AlexsLemonade/rOpenScPCA/)
```

jashapiro (approved):
LGTM, with typo fix. 

But now I'm a bit worried the links might not work as they are. I don't think a full url is worth doing, but I guess we can test after merge.

PR #950, opened 2024-12-17
Update dockerfile in consensus module to use renv

allyhawkins:
Here I'm just filling out the Dockerfile for `cell-type-consensus` to create the image for this module. I also cherry picked the commits from #936 to run the GHA to build this image and all the relevant changes to the lock file. 

I think we want to do this first and then I can adjust the workflow in #936 to use this image rather than trying to set up `renv` in CI. I believe this should help solve our [igraph problems](https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/936#discussion_r1889018879). 

I tested building this locally and didn't have any issues.

### DIFF
--- a/analyses/cell-type-consensus/Dockerfile
+++ b/analyses/cell-type-consensus/Dockerfile
@@ -1,10 +1,40 @@
-# A template docker file for creating a new analysis
-FROM ubuntu:22.04
+# Dockerfile for cell-type-consensus module
+FROM bioconductor/r-ver:3.20
 
 # Labels following the Open Containers Initiative (OCI) recommendations
 # For more information, see https://specs.opencontainers.org/image-spec/annotations/?v=v1.0.1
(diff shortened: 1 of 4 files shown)


analyses/cell-type-consensus/Dockerfile
```diff
@@ -1,10 +1,40 @@
-# A template docker file for creating a new analysis
-FROM ubuntu:22.04
+# Dockerfile for cell-type-consensus module
+FROM bioconductor/r-ver:3.20
```
sjspielman, line 2: noting we probably want 3.19 which is our preferred OpenScPCA-wide version.

allyhawkins: Based on this [comment](https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/936#discussion_r1889030481) we are okay with 3.20 here.

jaclyn-taroni (approved):
👍🏻

PR #955, opened 2024-12-19
Use docker for seurat-conversion tests

### LINKED ISSUE #945: Create module for Seurat conversions

jashapiro:
closes #945

This PR converts the actions for the seurat-conversion module to use the Docker image for the module. 

I also added the module to the various run-all workflows.

### DIFF
--- a/.github/workflows/docker_all-modules.yml
+++ b/.github/workflows/docker_all-modules.yml
@@ -43,6 +43,7 @@ jobs:
           - cell-type-nonETP-ALL-03
           - cell-type-ETP-ALL-03
           - cell-type-consensus
+          - seurat-conversion
     uses: ./.github/workflows/build-push-docker-module.yml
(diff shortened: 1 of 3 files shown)


sjspielman (approved):
👍

PR #961, opened 2024-12-20
Only skip push step in build action

jashapiro:
When performing a test run of https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/12434358502, I found that all the jobs were being skipped, which kind of defeats the testing!

So here I am updating the action to only skip the `push` step of the action.


jaclyn-taroni (approved):
👍🏻

PR #966, opened 2024-12-20
try to catch copykat error

### LINKED ISSUE #965: Analysis module failing in CI

jashapiro:
I couldn't resist trying to fix #965 with a quick error catch. Let's see if this works.


jashapiro:
Well it seemed to work, and the downstream steps already had `tryCatch()` statements to account for copykat failure, so maybe this is worth including.

jaclyn-taroni (approved):
👍🏻

Issue #972, opened 2025-01-03
Bug report: Zellkonverter/Basilisk initialization in Docker images doesn't work with rstudio users


jashapiro:
### Describe the bug

This is a followup issue to the problem noted in https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/968

In Docker images that use the `zellkonverter` package, we include a couple steps in the Dockerfile to pre-install the conda environment:

```
# Complete installation of zellkonverter conda env
ENV BASILISK_EXTERNAL_DIR /usr/local/renv/basilisk
RUN Rscript -e "proc <- basilisk::basiliskStart(env = zellkonverter::zellkonverterAnnDataEnv(), testload = 'anndata'); \
  basilisk::basiliskStop(proc); \
  basilisk.utils::cleanConda()"
```
e.g. https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/82a2f659771960b4aee53a41165b1169c19f73df/analyses/cell-type-wilms-tumor-06/Dockerfile#L33C1-L37C32

This works well for the `root` user, which is used when running scripts within the Docker environment, as is done in Nextflow workflows. However, when using the docker image interactively, having the `BASILISK_EXTERNAL_DIR` environment variable 
set causes us to run into a permissions issue with the lockfiles that are used by basilisk/conda.

### Steps to reproduce

1. Launch a rstudio-based Docker image that has a zellkonverter setup, such as `public.ecr.aws/openscpca/cell-type-wilms-tumor-06`
2. Login to RStudio with the `rstudio` user
3. Run `proc <- basilisk::basiliskStart(env = zellkonverter::zellkonverterAnnDataEnv(), testload = 'anndata')` in the R console

This will result in an error. 

### Consequences of the bug

The code above, and generally using zellkonverter as the `rstudio` user, results in the following error. 

```
Error in lock(plock, exclusive = FALSE) : 
  Cannot open lock file: Permission denied
```

This is at least in part because of the permissions settings of files in `/usr/local/renv/basilisk`, which are owned and initially only readable by the `root` user.

### Proposed solution

I initially tried changing the permissions of the `/usr/local/renv/basilisk` directory to make them read/writable by all users, but I am afraid that might not be sufficient, or we may need to adjust other permissions in that directory to make the permissions "sticky" so that all files in those subdirectories are accessible. We might also want to move the directory to `/usr/share` to reflect that usage.

[...]

PR #974, opened 2025-01-04
Install zellkonverter env as rstudio user

jashapiro:
This PR is a partial fix to #972

I updated the Dockerfile for the `cell-type-wilms-tumor-06` module to install zellkonverter packages as the `rstudio` user. This allows both the `rstudio` user and `root` to have access to the installed files, and prevents the lockfile permissions issue that we were seeing. 

When running R from the terminal, this seems to solve the zellkonverter problem, and everything works as expected.

However, when running `zellkonverter::readH5AD()` (or a `basilisk::basiliskStart()`) in the RStudio console, I now get the following error: 

```
Error in py_module_import(module, convert = convert) : 
  ImportError: /usr/lib/x86_64-linux-gnu/libssl.so.3: version `OPENSSL_3.2.0' not found (required by /usr/share/basilisk/1.16.0/zellkonverter/1.14.1/zellkonverterAnnDataEnv-0.10.6/lib/python3.12/site-packages/h5py/../../.././libcurl.so.4) 
```

This can be avoided by running `Sys.unsetenv("BASILISK_EXTERNAL_DIR")` and reinstalling the `zellkonverter` environment, so I suspect that the RStudio startup is somehow setting something differently from R in the terminal, but I can't yet figure out what that might be.

### DIFF
--- a/analyses/cell-type-wilms-tumor-06/Dockerfile
+++ b/analyses/cell-type-wilms-tumor-06/Dockerfile
@@ -30,8 +30,13 @@ RUN Rscript -e 'renv::restore()' && \
   rm -rf /tmp/downloaded_packages && \
   rm -rf /tmp/Rtmp*
 
-# Complete installation of zellkonverter conda env
-ENV BASILISK_EXTERNAL_DIR /usr/local/renv/basilisk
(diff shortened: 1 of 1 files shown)


jashapiro:
Based on https://github.com/AlexsLemonade/scpcaTools/pull/306, it appears that the proper solution to this is to set `BASILISK_USE_SYSTEM_DIR=1` before installation. I will update this to do that!

PR #977, opened 2025-01-08
Workflow to assign consensus cell types to ScPCA samples

### LINKED ISSUE #969: Create a table with consensus cell types across all ScPCA samples

#853 

Once we create a reference table with the consensus cell types we will want to assign these labels to all ScPCA samples. We want to create a table with all cells and all labels that we can then use to create plots that summarize these assignments. 

This PR will contain a small workflow that will have two steps: 

In step 1, we will have a script that runs on every SCE object in ScPCA. This script will read in the SCE object, grab the cell type assignments from the coldata and export that data as a TSV file. 

In step 2, all TSV files will be read in and combined into a single table. Then we will assign consensus cell types to all cells using the reference generated in #951. The final output will be a table with each cell, the sample/library/project id, SingleR and CellAssign annotations, and the consensus annotation. 

Alternatively, we could do the consensus assignment in Step 1, but I don't think it really matters that much here. We will also want to add in the CellAssign ontology ID when adding the consensus label. 

No 

No 

After we create the reference in #951.

allyhawkins:

Closes #969 

Here I'm adding a workflow to assign the consensus cell type labels for all cells in ScPCA. This produces a single TSV file that contains all cells in all samples from ScPCA (except cell line samples) and all cell type annotations. Included in the table are the ids (project, sample, library), the original annotations from `scpca-nf`, the ontology id and name for the panglao term, and new columns for the consensus labels. Any combination that is present in [`consensus-cell-type-reference.tsv`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv) is assigned and then all other combinations get assigned to "Unknown" as the `consensus_annotation` and `NA` for the `consensus_ontology`. 

I wrote two short scripts to do this: 

- The first script reads in a single SCE object, grabs the cell type annotations from the `colData`, and saves it to a single TSV file. 
- The second script reads in a directory with all the individual TSV files from step 1 and combines those into a single data frame. This script also takes in the consensus reference file and adds in the consensus labels for any combinations in the reference. All other are "Unknown"

These two steps are combined into a single workflow that can be run on all samples in ScPCA. 

A few things to note: 

- When making the final TSV file, I am also incorporating the `blueprint_annotation_fine` column, since that's the name that's directly linked to the ontology term in `SingleR`. 
- I also added in the columns with the assigned ontology and name for the original Panglao terms. 
- Since any samples with cell lines won't have columns for cell type annotations, I skip those. Are there any other scenarios I should account for? 
- I have not yet run this on the entire ScPCA dataset, but I tested it with a small set of samples. I also added the workflow to GHA here so it should run on all of the test data. Once the code is reviewed I will initiate the run to actually generate the TSV file and save it to S3. 
- Along those same lines, there are some small `renv` changes that are needed and since the GHA for this module runs with the Docker image, I'll file a separate PR to update that first. Until that gets in, I expect CI to fail. 
- I also updated the README to describe how to run the workflow. 

Yes! Next up we will actually look at the results. 

This will get saved to `s3://researcher-211125375652-us-east-2/cell-type-consensus` once I generate the file for all samples. 

A TSV file with all cells in ScPCA and all cell type annotations (SingleR, CellAssign, and consensus). 

That is coming up next! 

I was able to run this locally on a subset of samples. 

### DIFF
--- a/analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R
+++ b/analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R
@@ -8,7 +8,7 @@
 # consensus_annotation: human readable name associated with the consensus label 
 # consensus_ontology: CL ontology term for the consensus cell type 
 
-project_root <- rprojroot::find_root(rprojroot::is_git_root)
+project_root <- rprojroot::find_root(rprojroot::has_dir(".github"))
 
 library(optparse)
 
--- a/analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R
+++ b/analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R
@@ -34,7 +34,8 @@ option_list <- list(
   make_option(
     opt_str = c("--output_file"),
     type = "character",
-    help = "Path to file where combined TSV file will be saved, must end in `.tsv`"
+    help = "Path to file where combined TSV file will be saved. 
+      File name must end in either `.tsv` or `.tsv.gz` to save a compressed TSV file"
   )
 )
 
@@ -46,7 +47,8 @@ opt <- parse_args(OptionParser(option_list = option_list))
 # make sure reference files exist 
 stopifnot(
   "panglao reference file does not exist" = file.exists(opt$panglao_ref_file),
-  "cell type consensus reference file does not exist" = file.exists(opt$consensus_ref_file)
+  "cell type consensus reference file does not exist" = file.exists(opt$consensus_ref_file),
+  "output file must end in `.tsv` or `.tsv.gz`" = stringr::str_detect(opt$output_file, ".tsv|.tsv.gz")
 )
 
 # read in ref files 
--- a/analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R
+++ b/analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R
@@ -1,0 +1,106 @@
+#!/usr/bin/env Rscript
+
+# This script is used to combine all TSV files containing cell types into a single TSV file 
+# The output TSV file will include the following added columns: 
+# panglao_ontology: CL term assigned to panglao term
+# panglao_annotation: human readable value associated with the CL term for panglao term 
+# blueprint_annotation_fine: Fine-grained annotation from blueprint associated with singler_celltype_ontology 
+# consensus_annotation: human readable name associated with the consensus label 
+# consensus_ontology: CL ontology term for the consensus cell type 
+
+project_root <- here::here()
+
+library(optparse)
+
+option_list <- list(
+  make_option(
+    opt_str = c("--celltype_tsv_dir"),
+    type = "character",
+    help = "Path to directory containing TSV files with cell type annotations from single samples.
+      All TSV files in this directory will be combined into a single file."
+  ),
+  make_option(
+    opt_str = c("--panglao_ref_file"),
+    default = file.path(project_root, "references", "panglao-cell-type-ontologies.tsv"),
+    type = "character", 
+    help = "Path to file with panglao assignments and associated cell ontology ids"
+  ),
+  make_option(
+    opt_str = c("--consensus_ref_file"),
+    default = file.path(project_root, "references", "consensus-cell-type-reference.tsv"),
+    type = "character",
+    help = "Path to file containing the reference for assigning consensus cell type labels"
+  ), 
+  make_option(
+    opt_str = c("--output_file"),
+    type = "character",
+    help = "Path to file where combined TSV file will be saved, must end in `.tsv`"
+  )
+)
+
+# Parse options
+opt <- parse_args(OptionParser(option_list = option_list))
+
+# Prep ref files ---------------------------------------------------------------
+
+# make sure reference files exist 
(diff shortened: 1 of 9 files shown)


analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R
```diff
@@ -0,0 +1,106 @@
...
+  )
+
+# grab singler ref from celldex
+blueprint_ref <- celldex::BlueprintEncodeData()
+
+# get ontologies and human readable name into data frame for blueprint
+# in scpca-nf we don't include the fine label so this lets us add it in 
+blueprint_df <- data.frame(
+  blueprint_ontology = blueprint_ref$label.ont,
+  blueprint_annotation_fine = blueprint_ref$label.fine
```
jashapiro, line 78: Are there any cases where the ontology maps to more than one fine label? I feel like this is the case, which would result in duplicate rows when joining.

analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R
```diff
@@ -0,0 +1,106 @@
...
+    opt_str = c("--consensus_ref_file"),
+    default = file.path(project_root, "references", "consensus-cell-type-reference.tsv"),
+    type = "character",
+    help = "Path to file containing the reference for assigning consensus cell type labels"
+  ), 
+  make_option(
+    opt_str = c("--output_file"),
+    type = "character",
+    help = "Path to file where combined TSV file will be saved, must end in `.tsv`"
+  )
```
jashapiro, line 38: I assume the output file here is large: do we want to allow compressed file output? You don't have code to enforce the `.tsv` ending, and `write_tsv` supports this just by specifying the compression in file name (`.tsv.gz`, etc.) so just updating the help text would be sufficient here.

jashapiro (commented):
This looks good overall, with a few caveats:

- The first is the use of `here`, which could cause trouble in the future (if you made an R project for this module, which I actually recommend doing)
- The other concern I have is about the addition of `label.fine`; I seem to recall that there were cases where there was not a 1:1 mapping from ontology to the fine label, which may result in duplicate rows (by cell id) in the output table. 
- A few other minor comments.

analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R
allyhawkins: No. The ontology and fine label are 1:1. I checked this by making sure the length of the unique values in this data frame are equivalent and equal to the number of rows in the dataframe. Also from the [celldex vignette](https://bioconductor.org/packages/release/data/experiment/vignettes/celldex/inst/doc/userguide.html): 

> Typically, each reference provides three levels of cell type annotation in its column metadata:

> label.main, broad annotation that defines the major cell types. This has few unique levels that allows for fast annotation but at low resolution.
> label.fine, fine-grained annotation that defines subtypes or states. This has more unique levels that results in slower annotation but at much higher resolution.
> label.ont, fine-grained annotation mapped to the standard vocabulary in the [Cell Ontology](https://www.ebi.ac.uk/ols/ontologies/cl). This enables synchronization of labels across references as well as dynamic adjustment of the resolution.

This tells me that the `label.fine` are the names associated with the ontology IDs in the cell type ontology. I checked a few of these manually and they match up.

allyhawkins: > So I guess this is fine, but I guess I'm not sure what the value of it is? If we are just getting an alternative to the ontology label with a different naming convention, do we need it at all?

Perhaps the better approach here is to directly map the ontology ids to the name associated with them ourselves and remove the `blueprint_annotation_fine` column all together. I think I'll go ahead and do this but in the script that creates the consensus reference to begin with and not here. I'll file a new issue and new PR to address this before we actually run this on all ScPCA samples.

analyses/cell-type-consensus/scripts/04-combine-celltype-tables.R
```diff
@@ -0,0 +1,106 @@
...
+
+# This script is used to combine all TSV files containing cell types into a single TSV file 
+# The output TSV file will include the following added columns: 
+# panglao_ontology: CL term assigned to panglao term
+# panglao_annotation: human readable value associated with the CL term for panglao term 
+# blueprint_annotation_fine: Fine-grained annotation from blueprint associated with singler_celltype_ontology 
+# consensus_annotation: human readable name associated with the consensus label 
+# consensus_ontology: CL ontology term for the consensus cell type 
+
+project_root <- here::here()
```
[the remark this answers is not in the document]
allyhawkins: Just noting that this fails in CI with the following error: 

```
Error: No root directory found in /__w/OpenScPCA-analysis/OpenScPCA-analysis/analyses/cell-type-consensus or its parent directories. Root criterion: one of
- contains a directory ".git"
- contains a file ".git" with contents matching "^gitdir: "
Execution halted
```
see https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/12694163418/job/35383339441 

I'm looking into it, but posting it here in case you have any obvious solutions.

PR #978, opened 2025-01-08
Add SingleCellExperiment to renv for consensus cell typing

allyhawkins:
This PR makes the relevant `renv` updates to be able to run the scripts added in #977, specifically adding in `SingleCellExperimnt` and `optparse`.


jashapiro (approved):
LGTM

PR #971, opened 2025-01-03
Add marker gene lists for tumor cell states 

allyhawkins:
#### Please link to the GitHub issue that this pull request addresses.

Closes #939 

#### What is the goal of this pull request?

Here I am adding a marker gene list that can be used to define different tumor cell states that we expect to find in the Ewing sarcoma samples. Right now the goal is to label cell states that have been shown to be consistently present across Ewing sarcoma studies and are not looking to find novel cell states. Generally it's been shown that EWS cells lie on a continuum of EWS-FLI1 expression and publications have categorized cells into EWS-FLI1 high and EWS-FLI1 low cells. There's also a population of cells that has been labeled as "proliferative" across multiple publications (although there is some old argument that EWS-FLI1 high cells are also proliferative). Because of this, I made a table that has a column for the name for the cell state (either proliferative, EWS-low, or EWS-high) and then each row is an individual marker gene. I also included any custom gene lists that we may consider using to identify these cell states that were published by others. 

#### Briefly describe the general approach you took to achieve this goal.

There is no clear consensus of exactly which set of genes describes EWS-FLI1 high or low cells, but there are some key marker genes that were identified and validated in some of the publications. These are the ones that I chose to include in the main marker gene table, `tumor-cell-state-markers.tsv`. 

- Targets for EWS-FLI1 high and low cells from [Aynaud et al](https://doi.org/10.1016/j.celrep.2020.01.049) came from this sentence: 
> We also checked the single-cell expression dynamics of eight genes known to be directly modulated by EWSR1-FLI1 (upregulated genes: PRKCB, LIPI, CCND1, and NR0B1; downregulated genes: IGFBP3, IL8, LOX, and VIM) ([Figures 1](https://www.cell.com/cell-reports/fulltext/S2211-1247(20)30074-7?_returnURL=https%3A%2F%2Flinkinghub.elsevier.com%2Fretrieve%2Fpii%2FS2211124720300747%3Fshowall%3Dtrue#fig1)C and [S1](https://www.cell.com/cell-reports/fulltext/S2211-1247(20)30074-7?_returnURL=https%3A%2F%2Flinkinghub.elsevier.com%2Fretrieve%2Fpii%2FS2211124720300747%3Fshowall%3Dtrue#mmc1)).

[...]

Issue #668, opened 2024-07-25
Future idea: Use SingleR with SCPCL000822 and SCPCL000824 as references to annotate remaining Ewing samples


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#292 

### Describe the goals of the changes to the analysis module.

I noticed that when using all tumor cells as a reference and celldex references with `SingleR` we get a lot of cells that are tumor cells that have lower expression of marker genes than I would naively expect. I think it's likely that some cells will be more similar to cells from other Ewing samples than to normal cell types in the `celldex` reference, resulting in more cells being called tumor than actually should be. 

One thought I had is to create a reference using the two samples that have a good mix of tumor/normal cells in them (SCPCL000822 and SCPCL000824) and use those as a reference with celldex to annotate the rest of the samples in the project. 

### What will your pull request contain?

This will probably be at least 2 PRs that will do the following: 

- Generate a reference using the annotations from 822 and 824. 
- Create a workflow that uses that reference and the `04-run-singler.R` script to run `SingleR` for all other samples in the project. This workflow could probably also use the same template report `02-singler-results.Rmd` to evaluate the results. 

### Will you require additional software beyond what is already in the analysis module?

No

### Will you require different computational resources beyond what the analysis module already uses?

No 

### If known, when do you expect to file the pull request?

We should do this once we have confidently annotated normal cells in 822 and 824. I also don't know that we need to do this, but I think it could be helpful in refining the annotations that we currently have so I'm filing this as an idea for future analysis.

PR #981, opened 2025-01-10
Explore expression of marker genes for tumor cell states in 2 samples

### LINKED ISSUE #940: Notebook to explore identifying tumor cell states in Ewing samples using marker gene lists

#696 

After we have compiled a list of marker genes for the expected tumor cell states (see #939), we should explore actually using those marker genes to label tumor cell states in a handful of samples in SCPCP000015. 

This will be an exploratory notebook that looks at labeling tumor cell states in 2 Ewing samples. We should probably use SCPCS000490 and then one other sample that is more homogeneous. Generally, we want to label each cluster that contains tumor cells with a specific tumor cell state (e.g., EWS-FLI1 high/middle/low, MSC-like, proliferative, etc.). 

Some ideas of things we might want to use that are not fully baked: 

- AUCell with the marker gene lists created in #939 
- look at the expression of marker gene sets across each cluster
- Marker genes for each of the clusters and look at overlap with marker gene lists for each cell state 
- Enrichment of specific gene signatures that might indicate a certain cell state, e.g., MSC signature or proliferative gene signature. 

I don't think so

No 

January 2025

allyhawkins:

Closes #940 

Here I'm adding a notebook that looks at expression of marker genes for tumor cell states in 2 different Ewing libraries, SCPCL000822 and SCPCL000824. Generally I am using this notebook to answer the following questions: 

- Do we see expression of the genes in the tumor cell state marker gene list in our samples? 
- Do we see variation in expression of these genes across tumor cells? 
- Using these marker gene lists are there any clear cell states that we can define? 

The first thing I did was pick out the tumor cells. To do this I'm looking at the output from `aucell-singler-annotation.sh` and `evaluate-clusters.sh`. The `aucell-singler-annotation.sh` workflow identifies tumor cells using both `SingleR` and `AUCell` and the clustering workflow looks at different cluster assignments and gene expression for tumor marker genes across the clusters. I use the output of both to pick out clusters of cells that are tumor cells. 

After getting the cell type assignments, I looked at the expression of the marker genes in the `tumor-cell-state-markers.tsv` file and the genes in the custom gene signatures we added as part of #971. To do this I calculated the sum of all marker genes in the gene set and used that for plotting throughout the notebook. I noticed that for both samples there was a cluster of cells that had high expression of EWS-FLI1 low targets, so I pulled those out as tumor cells too. 

Once I had the tumor cells, I re-did clustering on just those cells. Then I looked at expression of the gene sets in each of those clusters to see if I could assign clusters to different cell states (mainly EWS-FLI1 low/high) based on expression of that gene set. I was able to easily pick out low vs. high cells, but there looks like there might be some variation and potentially a group of "EWS-FLI1 middle" cells that express EWS-FLI1 high marker genes but to a lesser degree. 

A few key observations that I made along the way: 

- There isn't much expression of the proliferative gene set, so I'm not sure that's really going to be useful. 
- There is really clear separation between low and high, but then there is some in-between expression that's hard to call. Expression of the EWS-FLI1 high targets has been shown to lie on a continuum so this is what I would expect. If we want to come up with more classifications we would have to think about how to bin the cells into different EWS-FLI1 levels and do things that are outside the scope of this notebook. 
- I did think about using something like AUCell, but you really need some clear separation of gene expression and we don't have a lot of that outside of EWS high/low, which may be fine 🤷‍♀️ . I also thought this was getting a little long, so AUCell should be a separate notebook if we do want to use it. 
- I think the two custom gene sets that we identified from Aynaud and Wrenn are the most helpful. We see more variation in expression in those than the gene list we came up with, so I'm inclined to use those moving forward. 

Yes

I think the first thing we want to do is finalize cell type annotations using the SingleR/ clustering results and then classify cells as EWS-FLI1 high/low. I think we can add in the "middle" group later. 

Here's the rendered copy of the notebook for review: 
[06-tumor-cell-state-assignment.html.zip](https://github.com/user-attachments/files/18380834/06-tumor-cell-state-assignment.html.zip)

As part of the notebook I also saved the annotations to a TSV file. 

- There are clear groups of EWS-FLI1 high and low cells
- It's hard to identify any proliferative cells since expression of those genes is pretty non-existent
- There may be a group of cells that lie in the middle. They don't have expression of the low marker genes and they don't have quite as high expression of the high marker genes. I think if we want to classify these moving forward we should make some rules/ be more stringent about how we do it. 

### Author checklists

_Check all those that apply._


analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+cluster_files <- file.path(cluster_results_dir, cluster_file_names) |>
+  purrr::set_names(library_ids)
+```
+
+Below we set up the data for plotting throughout this notebook. 
+
+1. The output from `aucell-singler-annotation.sh` is read in and annotations are lumped so that we can easily plot the top cell types. 
+2. Clustering results from `evaluate-clusters.sh` is read in and subset to the desired parameters we would like to use for assigning clusters (leiden, modularity, resolution 0.5, nearest neighbors 20). 
+3. The marker genes for tumor cell states and custom gene signatures are used to calculate the sum of all marker genes in each gene set for each library. 
+4. UMAP embeddings for each library is pulled out into a data frame and combined with `SingleR` annotations, cluster assignments, and marker gene expression for all marker gene sets.
```
sjspielman, line 98:
```suggestion
4. UMAP embeddings for each library are pulled out into a data frame and combined with `SingleR` annotations, cluster assignments, and marker gene expression for all marker gene sets.
```

analyses/cell-type-ewings/exploratory_analysis/README.md
```diff
@@ -20,6 +20,9 @@ This includes comparing annotations to those obtained from marker gene annotatio
 5. `05-cluster-exploration.Rmd`: This notebook looks at clustering across different parameters to choose the most optimal clusters for `SCPCL000822` and `SCPCL000824`. 
 Additionally, we look at expression of marker genes from `references/visser-all-marker-genes.tsv` across all clusters and use that to refine tumor cell annotations obtained from running `aucell-singler-annotation.sh`. 
 
+6. `06-tumor-cell-state-assignments.Rmd`: This notebook looks at assigning tumor cell states in `SCPCL000822` and `SCPCL000824`. 
+Tumor cells are gruoped into `EWS-FLI` high, `EWS-FLI` low, and `EWS-FLI` middle based on expression of marker genes.
```
sjspielman, line 24:
```suggestion
Tumor cells are grouped into `EWS-FLI` high, `EWS-FLI` low, and `EWS-FLI` middle based on expression of marker genes. 
```

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+      library_id == "SCPCL000822" & cluster == 5 ~ "macrophage",
+      # cluster 8 has high MSC marker gene expression and a mix of MSC like cell types
+      library_id == "SCPCL000822" & cluster == 7 ~ "mesenchymal-like cell",
+      library_id == "SCPCL000824" & cluster %in% c(2, 3, 4, 5, 6, 7) ~ "tumor", 
+      library_id == "SCPCL000824" & cluster == 8 ~ "endothelial cell",
+      # cluster 9 has high immune markers and is mostly macrohpages
+      library_id == "SCPCL000824" & cluster == 9 ~ "macrophage",
+      # cluster 1 has high MSC marker gene expression and a mix of MSC like cell types
+      library_id == "SCPCL000824" & cluster == 1 ~ "mesenchymal-like cell",
+      # if they didn't have high expression of any markers, we label them as unknown
```
sjspielman, line 255: Suggesting some comments to make this a smidge more readable at a glance. do you think this helps?

```suggestion
      ############## Library SCPCL000822 ###############
      library_id == "SCPCL000822" & cluster %in% c(1, 3, 4) ~ "tumor",
      library_id == "SCPCL000822" & cluster == 6 ~ "endothelial cell",
      # cluster 2 has predominantly chondrocytes and has MSC marker gene expression
      library_id == "SCPCL000822" & cluster == 2 ~ "chondrocyte",
      # clusters with high "immune" marker gene expression are macrophage
      library_id == "SCPCL000822" & cluster == 5 ~ "macrophage",
      # cluster 8 has high MSC marker gene expression and a mix of MSC like cell types
      library_id == "SCPCL000822" & cluster == 7 ~ "mesenchymal-like cell",
      ############## Library SCPCL000824 ###############
      library_id == "SCPCL000824" & cluster %in% c(2, 3, 4, 5, 6, 7) ~ "tumor", 
      library_id == "SCPCL000824" & cluster == 8 ~ "endothelial cell",
      # cluster 9 has high immune markers and is mostly macrohpages
      library_id == "SCPCL000824" & cluster == 9 ~ "macrophage",
      # cluster 1 has high MSC marker gene expression and a mix of MSC like cell types
      library_id == "SCPCL000824" & cluster == 1 ~ "mesenchymal-like cell",
      # if they didn't have high expression of any markers, we label them as unknown
```

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+    cluster_classification_df |>
+      ggplot(aes(x = UMAP1, y = UMAP2, color = !!sym(geneset))) +
+      geom_point(alpha = 0.5, size = 0.1) +
+      facet_wrap(vars(library_id)) +
+      scale_color_viridis_c() +
+      theme(
+        aspect.ratio = 1
+      )
+  }) |> 
+  patchwork::wrap_plots(ncol = 2)
```
sjspielman, line 313: I'm not sure I'd wrap these vs just print them one on each line, since they are a bit small in the HTML compared to other plots, and these are important ones to be able to see

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+```
+
+It looks like there's a cluster of cells in both libraries that have expression of the `EWS-low` and `wrenn-ews-low` signatures. 
+Those cells correspond to the cells that we are labeling as "chondrocyte" in `SCPCL000822` and "mesenchymal-like" in `SCPCL000824`. 
+The chondrocytes also express high markers of mesenchymal-like genes and the literature talks about tumor cells with low `EWS-FLI1` expression being "MSC-like". 
+I think there's a possibility that these are tumor cells, so let's relabel them.
```
sjspielman, line 319: This seems like a pretty decent result here!

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+- `proliferative` : Marker genes from `tumor-cell-state-markers.tsv` that correspond to the `proliferative` cell state. 
+Note that a lot of the Ewing sarcoma literature argues that `EWS-FLI1` high cells are proliferative. 
+- `aynaud-ews-high`: Marker genes from `gene_signatures/aynaud-ews-targets.tsv` that correspond to genes activated by `EWS-FLI1`. 
+- `wrenn-ews-low`: Marker genes from `gene_signatures/wrenn-nt5e-genes.tsv` that correspond to genes repressed by `EWS-FLI1`. 
+
+```{r, fig.height=7, fig.width=10}
+# pull out columns that have sum of marker gene expression 
+marker_gene_cols <- colnames(cluster_classification_df)[stringr::str_detect(colnames(cluster_classification_df), "_sum$")]
+
+# look at the expression of all gene sets across both libraries
```
sjspielman, line 301: Looking at these plots, I'm curious if the cells with the high proliferative levels are actually a subset of the EWS-FL1 high cells or whether they are distinct cells within the same cluster. I think there are more high expression proliferative cells in the data since the plot suggests at first glance and lighter points are less frequent so they are hiding behind the darker points. If these are indeed mostly a subset, this could actually be a "sub-state" worth classifying. Might be worth a quick check, maybe a scatterplot or so?

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+- There are some very obvious `EWS-FLI1` low groups of cells. 
+- The `EWS-high` and `aynaud-EWS-high` show clusters that have high expression of `EWS-FLI1` targets, but the `aynaud-EWS-high` appears to have some more separation across clusters. 
+That tells me that gene signature might be more helpful in classifying cells across the `EWS-FLI1` continuum. 
+
+### Sub cluster tumor cells 
+
+We would expect that tumor cells that are in similar cell states would cluster together. 
+So here I am going to take just the tumor cells and re-assign clusters. 
+I'm using the same parameters we use for the whole sample, leiden, modularity, resolution of 0.5, and nearest neighbors of 20. 
+I think if we really wanted to be stringent we would evaluate clusters here and find the best parameters to use, but I think starting with these is good.
```
sjspielman, line 382: I don't think there are "perfect" answers here since clustering is its own special can of worms. But, I do want to say that I wonder what the difference would be between using the same parameters on subsetted data, vs using the full dataset with a higher resolution. I agree this is a decent enough place to start though, since it's anyone's guess how to subset clusters anyways!

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+    # pull out the barcodes that are from tumor cells
+    tumor_cells <- tumor_cells_df |> 
+      dplyr::filter(library_id == id) |> 
+      dplyr::pull(barcodes)
+    
+    # filter the sce to only those barcodes
+    filtered_sce <- sce[ , tumor_cells]
+    
+    # grab pcs for clustering
+    pcs <- reducedDim(filtered_sce)
```
sjspielman, line 399: My poor brain when seeing the phrase `filtered_sce` 😂 
I'm suggesting this in case "future someone" looks at this code and does a similar double take. It's of course fine as is, but I can't help myself!

```suggestion
    # filter the sce to only those barcodes
    tumor_sce <- sce[ , tumor_cells]
    
    # grab pcs for clustering
    pcs <- reducedDim(tumor_sce)
```

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+        aspect.ratio = 1
+      ) +
+      labs(title = id) +
+      scale_color_brewer(palette = "Dark2") +
+      guides(color = guide_legend(override.aes = list(alpha = 1, size = 1.5)))
+  })
+```
+
+Now we will look at the marker gene expression across all clusters for each sample. 
+First using a density plot and then a heatmap.
```
sjspielman, line 446:
```suggestion
Now we will look at the marker gene expression across all clusters for each sample, first using a density plot and then a heatmap. 
```

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+    filtered_df <- tumor_cells_df |> 
+      dplyr::filter(library_id == id)
+    
+    full_celltype_heatmap(filtered_df, marker_gene_cols, "sub_cluster", plot_name = id)
+    
+  })
+
+```
+
+It looks like for both samples we have some clear division between EWS high and EWS low cells.
```
sjspielman, line 483: Hmm.. what do we make of the fact that it barely looks like `annotation` (which is the sub clusters - can you modify the heatmap code to get that labeled as "sub_cluster" and not as "annotation"?) is clustered on the 2nd heatmap? Of course, the hierachical clustering algorithm that the heatmap is using is decently different from the one used to cluster in the first place, but the total lack of relationship I find odd. I'm curious, what does the heatmap look like if we first arrange the data by subcluster, and then turn off x-axis clustering so that the subclusters indeed are grouped together?

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+```{r}
+# categorize tumor cells based on EWS expression 
+tumor_cells_df <- tumor_cells_df |> 
+  dplyr::mutate(
+    cell_state = dplyr::case_when(
+      library_id == "SCPCL000822" & sub_cluster == 2 ~ "EWS-low",
+      library_id == "SCPCL000822" & sub_cluster %in% c(1, 4) ~ "EWS-mid",
+      library_id == "SCPCL000822" & sub_cluster == 3 ~ "EWS-high",
+      library_id == "SCPCL000824" & sub_cluster == 1 ~ "EWS-low",
+      library_id == "SCPCL000824" & sub_cluster %in% c(6, 8) ~ "EWS-mid",
```
sjspielman, line 511: aren't there are only 7 subclusters here? we should check this code

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
sjspielman: Ok, now I'm looking at the next version of this heatmap after you group into putative cell states, and I feel less concerned! The rainbow look, shall we say, was probably just brought on by some _very similar_ sub-clusters.

sjspielman (commented):
Overall I think this is a really reasonable exploration of results! I very much agree that if you want to consider AUCell, that's a separate notebook; this is long enough. 

One additional item I wanted to point out beyond the comments I left was that, I'm not sure we can rule out proliferative being useful based on 2 libraries alone, although I do agree based on the density plots that there seems to be a lot less going in within that gene set here. I did leave a comment about the relationship between proliferative and high, which might also help to get a sense of how much we want to use proliferative going forward if at all!

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,606 @@
...
+  library(ggplot2)
+})
+
+# Set default ggplot theme
+theme_set(
+  theme_classic()
+)
+
+# set seed
+set.seed(2024)
```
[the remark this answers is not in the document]
allyhawkins: Maybe by February, I'll remember it's a new year 😂

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
allyhawkins: I agree! And I also was revisiting [this issue](https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/668) that I filed about using these samples to come up with a base reference for SingleR. One of the reasons I was worried about using just SingleR the way we set it up was because a lot of cells were getting called tumor cells that had low expression of the EWS markers. But if they have low expression of the EWS targets and high expression of the EWS low signature then I think those are actually tumor cells! 

One idea I have as follow up here is to label these two samples as I'm trying to do here and revisit using them with SingleR. Probably need a third sample but that might be a good route.

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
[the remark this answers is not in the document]
sjspielman: I am very much still at the <types 2024...deletes the 4 and types 5> stage! 🥴

allyhawkins:
> One additional item I wanted to point out beyond the comments I left was that, I'm not sure we can rule out proliferative being useful based on 2 libraries alone, although I do agree based on the density plots that there seems to be a lot less going in within that gene set here. I did leave a comment about the relationship between proliferative and high, which might also help to get a sense of how much we want to use proliferative going forward if at all!

This comment made me take another look at this and I do see some expression of the proliferative gene set, but it looks so much lower because the values are lower. This is probably a side effect of having marker gene sets with varying numbers of genes in them. Because of that, I actually think we want to use the mean here. It doesn't change the overall conclusions and the plots mostly look the same, but it does keep things on a similar scale. It also makes it easier to see all the gene sets in the heatmaps. 

In revisiting this, I think there is one group of cells that are "proliferative" in SCPCL000824 that also line up with EWS-high. I feel comfortable labeling these as such. The one thing that does change when using mean is that the differences between EWS-mid and EWS-high are not as striking to see in the heatmap, but I see it in the UMAP. Maybe that's okay and we could also just label cells as low, high, or proliferative? 

Just an FYI that I wrote a new function to calculate means rather than modifying the existing function to calculate sums because we use that function in lots of places and I was trying not to break any previous things that use it. 

I also updated the heatmap to not cluster the columns, but I wasn't able to change the name of the `annotation` bar. That also gets used by a lot of other notebooks and would require moving the definition of the ComplexHeatmap annotation out of the function. If we really wanted to do it then I would need to move the code into this notebook and there's already so much in this notebook. I did add a note about what the annotation label means and can add it if you feel like it's really necessary. 

New rendered notebook:
[06-tumor-cell-state-assignment.html.zip](https://github.com/user-attachments/files/18382971/06-tumor-cell-state-assignment.html.zip)

sjspielman (commented):
The changes you've made here seem good to me (and totally fine about the annotation label, it's good enough!)!

> The one thing that does change when using mean is that the differences between EWS-mid and EWS-high are not as striking to see in the heatmap, but I see it in the UMAP. Maybe that's okay and we could also just label cells as low, high, or proliferative?

I like the idea of labeling as high-proliferative as you have done (for that one cluster at least), but I think we should bear in mind the caveat that this does mean other high cells aren't also proliferative. That said, I was intrigued that a cluster came out for proliferative here at all, and in reminding myself of the marker genes we're using, I wondered if we just don't have enough proliferative markers for sufficient power. So I did another quick search and uncovered* the contender `VRK1` from https://doi.org/10.1016/j.ccell.2014.10.004. Locally, I added this to the marker gene TSV and re-rendered the notebook. I think this may be a contender to include and achieve a bit more signal, what do you think?
[06-tumor-cell-state-assignment-with-VRK1.html.zip](https://github.com/user-attachments/files/18397426/06-tumor-cell-state-assignment-with-VRK1.html.zip)

analyses/cell-type-ewings/exploratory_analysis/06-tumor-cell-state-assignment.Rmd
```diff
@@ -0,0 +1,629 @@
...
+`SCPCL000822`: 
+`EWS-FLI1` low  - 2
+`EWS-FLI1` middle - 1 & 4
+`EWS-FLI1` high - 3 
+
+`SCPCL000824` : 
+`EWS-FLI1` low - 1
+`EWS-FLI1` middle - 6 
+`EWS-FLI1` high - 3, 4, 5, 7
+`EWS-FLI1` high proliferative - 2
```
sjspielman, line 520: Forgot to add this comment before - this is each appearing on one line so let's actually make it bullets

```suggestion
Based on this I would make the following assignments: 

`SCPCL000822`: 

* `EWS-FLI1` low  - 2
* `EWS-FLI1` middle - 1 & 4
* `EWS-FLI1` high - 3 

`SCPCL000824` : 

* `EWS-FLI1` low - 1
* `EWS-FLI1` middle - 6 
* `EWS-FLI1` high - 3, 4, 5, 7
* `EWS-FLI1` high proliferative - 2
```

allyhawkins:
> I like the idea of labeling as high-proliferative as you have done (for that one cluster at least), but I think we should bear in mind the caveat that this does mean other high cells aren't also proliferative. That said, I was intrigued that a cluster came out for proliferative here at all, and in reminding myself of the marker genes we're using, I wondered if we just don't have enough proliferative markers for sufficient power. So I did another quick search and uncovered* the contender `VRK1` from https://doi.org/10.1016/j.ccell.2014.10.004.

I think we should figure out what exactly we mean as "proliferative". It's been shown that EWS-FLI1 high cells have a more proliferative phenotype and that EWS-FLI1 low cells have a more metastatic phenotype (see https://www.nature.com/articles/onc2016498). So if we use that logic, then all EWS-FLI1 high cells should be proliferative. And then if you add in VRK1, an EWS-FLI1 target which has been shown to contribute to tumor cell growth and proliferation in the paper you posted, then you see more of the EWS-FLI1 high cells overlapping with proliferative cells. I think if we want to use that logic, then we don't need to have the "proliferative" category at all and just categorize by EWS-FLI1 levels. 

However, those studies were all done in bulk and all cell lines. The "proliferative" class that I was trying to characterize is based on the "Ewing sarcoma proliferating cells" labeled in https://www.biorxiv.org/content/10.1101/2024.01.18.576251v2.full. These were specifically identified by looking at the two markers that we have on the list and are a subset of Ewing tumor cells in patient tumor cells using single-cell. That doesn't mean it should be the gold standard, but I think if we want to isolate a specific "proliferating" population, then we can't just use EWS-FLI1 targets, since those should be high regardless. 

I did revisit [Figure 2](https://www.biorxiv.org/content/biorxiv/early/2024/06/20/2024.01.18.576251/F2.large.jpg) from Goodspeed et al,. and I wonder if we want to add `TOP2A` to the list? 
I think if we want to add VRK1 it should be to the `EWS-FLI` high list. 

Also, I think we should add ICAM1 based on Franzetti et al. I completely forgot about that paper until now! 

Here's an example of the report with `ICAM1`, `TOP2A`, and `VRK1`. 
[06-tumor-cell-state-assignment.html.zip](https://github.com/user-attachments/files/18399738/06-tumor-cell-state-assignment.html.zip)

sjspielman (approved):
> That doesn't mean it should be the gold standard, but I think if we want to isolate a specific "proliferating" population, then we can't just use EWS-FLI1 targets, since those should be high regardless.

Got it, thanks for the biology! Totally makes sense.

> I did revisit [Figure 2](https://www.biorxiv.org/content/biorxiv/early/2024/06/20/2024.01.18.576251/F2.large.jpg) from Goodspeed et al,. and I wonder if we want to add TOP2A to the list? I think if we want to add VRK1 it should be to the EWS-FLI high list.

If we're including the others from Goodspeed, then I agree TOP2A should be in there per that linked heatmap, and I agree with adding VRK1 to the high list.

Looking at the updated report, it all looks good to me at this point actually and I think we have a pretty solid gene set here to move forward with. The only changes to make are to add the additional DOI we have to the `references/README.md` where the marker gene file is described, and we do want to get bullets in that one spot I commented on since otherwise it's all on the same line in the report - 
Screenshot 2025-01-13 at 1 00 06 PM


Otherwise, good to go! 🐈

allyhawkins:
> Looking at the updated report, it all looks good to me at this point actually and I think we have a pretty solid gene set here to move forward with. The only changes to make are to add the additional DOI we have to the references/README.md where the marker gene file is described, and we do want to get bullets in that one spot I commented on since otherwise it's all on the same line in the report -

Sorry about that, I thought I had added the bullets but they should be there now! And then I updated the README to include the new references we're using. 

Since I made some small changes to functions that are used in other places used by some of the scripts run in CI, I'm going to make sure checks pass before merging this.

PR #983, opened 2025-01-13
Map blueprint ontology to CL names

### LINKED ISSUE #979: Use names from cell type ontology (CL) instead of blueprint names in consensus reference
> So I guess this is fine, but I guess I'm not sure what the value of it is? If we are just getting an alternative to the ontology label with a different naming convention, do we need it at all?

Perhaps the better approach here is to directly map the ontology ids to the name associated with them ourselves and remove the `blueprint_annotation_fine` column all together. I think I'll go ahead and do this but in the script that creates the consensus reference to begin with and not here. I'll file a new issue and new PR to address this before we actually run this on all ScPCA samples.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/977#discussion_r1909264179_

allyhawkins:

Closes #979 

Here I am adding a column that maps the ontology ID in `BlueprintEncodeData` to the CL name found in the cell type ontology. 

- I updated the script that creates the consensus reference table to include a column for the mapped name `blueprint_annotation_cl`.  This also means a huge diff for this file since it includes a new column. 
- Since the reference that we have doesn't have every combination of assignments (due to size constraints), I also map the IDs in the script to combine all tables and add consensus labels. Ultimately, I chose to remove the `blueprint_annotation_fine` column altogether and just use the `blueprint_annotation_cl` column. 
- I then updated the readme to reflect this change. 

### Author checklists

_Check all those that apply._


jashapiro (approved):
LGTM

allyhawkins:
I'm going to cancel the workflow run for this module since the changes I just had to merge in from main aren't relevant.

Issue #856, opened 2024-11-05
Improve Wilms Tumor Dataset Annotation (SCPCP000006) - explore `predicted.score` and `has_cnv.score` thresholds


maud-p:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

This issue follows the PR https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/844 and the 2 comments:

- https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/844#discussion_r1824688578
- https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/844#discussion_r1824673660

### Describe the goals of the changes to the analysis module.

I would like to explore difefrent thresholds for filtering and annotating based on the `predicted.score` and `cnv.score`.
I would like to:

- [ ] improve the umpa reduction visualization with a 2-colors plot showing only one annotation and the rest in grey. 

- [ ] look at the distribution of `predicted.score` for each of the `predicted.compartment` and `predicted.cell_type`. So far, we only used the `predicted.score` to select _normal_ cells (i.e. endothelial and immune cells), but don't use it to filter out cells with very low confident annotation (label as _unknown_). 
- [ ] render few notebook with a cnv_threshold of 0, 1 or 2 and evaluate the identification of normal cells. I'd like to check the distribution of the `predicted.score` of endothelial, immune, _normal_ kidney and _normal_ stroma cells using each of the threshold. It can be that, due to false positive cnv, normal cells showed some infered cnv. If this is the case, we should expect to recover more normal cells with high `predicted.score` using higher `cnv_threshold`.

### What will your pull request contain?

Few changes in the `07` notebook

### If known, when do you expect to file the pull request?

~ November

sjspielman:
Hi @maud-p, glad to see you back here in issues! I wanted to give you a heads up about continuing this module - I am still working behind the scenes on your module to get it all running in CI. I have updated the label transfer code. It's not yet merged into main (but will be within the next 2 weeks I think 🤞), since I am still working in a separate branch to fix some bugs we are now able to find with all code running in CI. 

(FYI - one silly (!!!) bug I found is that somehow we never actually _applied_ the score threshold in inferCNV, so I am making sure that happens too!).

[...]

PR #994, opened 2025-01-17
improve_07_annotation

maud-p:

This PR is linked to the [issue 856](https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/856).

I aim here to improve the workflow of the annotations performed in `07_combined_annotation_across_samples_exploration.Rmd`.

The suggested changes aim to overcome some misclassification that have been spotted in the first version of the annotation while working on the integration of the data. 

1. In the first version of the annotation, I realized that a lot of immune cells were filtered out because of the `cnv-threshold`. See the expression of PTPRC = CD45 below: 
image

We requested immune and endothelial cells to have no infered cnv to be labeled as normal. I think that the few CNV that were infered for immune cells are cell type specific expression of some genes that colocalized on one chromosome. Thus I decided to allow a bit of flexibility in regards to CNV for immune and endothelial cells and only based the annotation of immune and endothelial cells on the `predicted.cell-type` and associated `prediction.score`.

2. Similar to the immune cells, there is a high probability that we also infer false positive CNV in normal epithelial/stroma cells. I am not sure that it will be possible to differentiate between normal cells with false positive CNV from cancer cells. For that reason, I decided to label as `normal` cells with a `cnv-score <= cnv-threshold` and as `cancer` cells with a `cnv-score > cnv-threshold + 2`.
The choice of `+2` is a bit arbitrary, but based on the approximate number of CNV that can be detected in `endothelial` and `immune` cells. 

3. There is the possibility of cancer cell without any CNV. It would be extremly difficult in that case to differentiate them from normal cells. The strategy I decided to go with to reduce the risk of misclassification is to use the prediction score of label transfer. Indeed, we expect normal cells to resemble the cells from the fetal kidney reference and thus have a high `prediction.score`.

4. While working on the integration (see umap reduction here colored by `cnv_score`), I realized that samples -177, -180, -181, -190, -197 have strikingly very low `cnv_score`. The `cnv_score` is simply the number of chromosome presenting a CNV for each patient. Of note, these 5 samples are the ones with no immune and/or endothelial cells for which we couldn’t run `infercnv` properly. We ran `infercnv` without any reference. In that case, the mean over all cellular expression profiles is taken as a reference. If the sample is mostly composed of cancer cell, the cancer-associated CNV are taken as the normal reference...

image

I am afraid not to have a solution for this, but I am quite sure that the cells for these 5 patients are mis-classified. This will affect the downstream differential expression analysis cancer versus normal and the potential finding of Wilms tumor histology specific markers. This is why I would strongly recommand forcing the annotations for these patients to `unknown`.
image

6. I slightly changed the summarized CNV heatmap with `do_Feature_mean`. Initially, the `feature` ploted with `do_Feature_mean` was `prop_cnv_chr[i]`, the proportion of the chromosome i affected by a CNV. The function `do_Feature_mean` took then the mean over all cells in a group. This can be misleading are it is then impossible to interpret the mean value. Example with a mean of 0.5, we could say:
- half of the cells in the group gained/lossed the entire chromosome, half are unaffected
- all cells gained/lossed half of the chromosome
- ... all other combinations possible!

For that reason, I changed the `feature` to `has_cnv_chr[i]`, which is a binary information of the presence/absence of CNV in the chromosome i for the given cell.  While taking the mean over all cells in a group, we then have the information of the percentage of cells within the group having any kind of CNV in this chromosome. 

-[ ] one for the integration of the samples
-[ ] one for differential expression analysis, looking for Wilms tumor, histology specific markers

### Results

The annotation file is generated automatically while running the script and will be saved in the `results` folder.
The notebook is saved in the `notebook` folder.

### DIFF
--- a/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
+++ b/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
@@ -147,7 +147,7 @@ for sample_dir in ${data_dir}/${project_id}/SCPCS*; do
     fi
 
     # Run inferCNV
-    Rscript scripts/06_infercnv.R --sample_id $sample_id --reference $reference --HMM i3 ${test_string}
+    Rscript scripts/06_infercnv.R --sample_id $sample_id --reference $reference --HMM i3 ${test_string} --score_threshold ${predicted_celltype_threshold}
 done
 
 
--- a/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
+++ b/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
@@ -31,7 +31,7 @@ TESTING=${TESTING:-0}
 RUN_EXPLORATORY=${RUN_EXPLORATORY:-0}
 THREADS=${THREADS:-32}
 project_id="SCPCP000006"
-
+predicted_celltype_threshold=0.85
 # Ensure script is being run from its directory
 module_dir=$(dirname "${BASH_SOURCE[0]}")
 cd ${module_dir}
@@ -150,9 +150,10 @@ for sample_dir in ${data_dir}/${project_id}/SCPCS*; do
     Rscript scripts/06_infercnv.R --sample_id $sample_id --reference $reference --HMM i3 ${test_string}
 done
 
+
 # Render notebook to make draft annotations
 Rscript -e "rmarkdown::render('${notebook_dir}/07_combined_annotation_across_samples_exploration.Rmd',
-                        params = list(predicted.celltype.threshold = 0.85, cnv_threshold = 0, testing = ${TESTING}),
+                        params = list(predicted.celltype.threshold = ${predicted_celltype_threshold},  testing = ${TESTING}),
                         output_format = 'html_document',
                         output_file = '07_combined_annotation_across_samples_exploration.html',
                         output_dir = '${notebook_dir}')"
--- a/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
+++ b/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
@@ -161,7 +161,7 @@ done
 
 # Render notebook to make draft annotations
 Rscript -e "rmarkdown::render('${notebook_dir}/07_combined_annotation_across_samples_exploration.Rmd',
-                        params = list(predicted.celltype.threshold = ${predicted_celltype_threshold}, cnv_threshold_low = ${cnv_threhold_low}, cnv_threshold_high = ${cnv_threhold_high},  testing = ${TESTING}),
+                        params = list(predicted.celltype.threshold = ${predicted_celltype_threshold}, cnv_threshold_low = ${cnv_threshold_low}, cnv_threshold_high = ${cnv_threshold_high},  testing = ${TESTING}),
                         output_format = 'html_document',
                         output_file = '07_combined_annotation_across_samples_exploration.html',
                         output_dir = '${notebook_dir}')"
--- a/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
+++ b/analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
@@ -31,7 +31,15 @@ TESTING=${TESTING:-0}
 RUN_EXPLORATORY=${RUN_EXPLORATORY:-0}
 THREADS=${THREADS:-32}
 project_id="SCPCP000006"
-predicted_celltype_threshold=0.85
+
+# Define the `predicted_celltype_threshold` which is used to identify cells with a high confident label transfer from the fetal kidney reference.
+# This score is used in the script `06_infercnv.R` and in the notebook `07_combined_annotation_across_samples.Rmd`
+predicted_celltype_threshold=0.75
+
+# Define the `cnv_threshold_low/high` which are used to identify cells with no CNV (`cnv_score` <= `cnv_threshold_low`) or with CNV (`cnv_score` > `cnv_threshold_high`)
+cnv_threshold_low=0
+cnv_threshold_high=2
+
 # Ensure script is being run from its directory
 module_dir=$(dirname "${BASH_SOURCE[0]}")
 cd ${module_dir}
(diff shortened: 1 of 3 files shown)


sjspielman:
Hi @maud-p, just wanted to give you a heads-up for time frame! I anticipate getting to review this next Tuesday 1/21, since we are out of the office Monday 1/20 for a holiday and I don't think I'll have a chance to get to this today. Have a good weekend in the meantime! :)

maud-p:
Hi @sjspielman ,
Thank you for letting me know. Please no rush for it, as I wanted to send the PR before leaving for vacations ;)
I'll be back Monday 26/01.
I also wish you a great weekend!

sjspielman (commented):
Generally speaking, these annotation changes you are proposing seem reasonable to me. I have a few initial thoughts about some of the way this is implemented. Once you make these changes, I will do a second round of more thorough code review. Let me know if I can clarify anything here! 

- Most importantly, the score threshold (0.85 being changed to 0.75) was not only used in this notebook. It was also used in several exploratory steps (we don't need to worry about those at this point), but more importantly it was used for inferCNV inference. We would like to keep this parameter consistent whenever it is used to create the annotations, which is during inferCNV inference and in this notebook. Therefore, to change the threshold, you'll want to take a step back here to apply the changes more broadly and make use of the arguments and parameters we have already established. I recommend that you:
	- Define a variable at the top of this script `00_run_workflow.sh` with the chosen threshold you'd like to use, 0.75
	- Pass this variable into both the inferCNV step on line 150, and this notebook on the following line
	- Then, re-run these steps to regenerate both the inferCNV results and this notebook; you'll want to make sure those new inferCNV results are also updated in your researcher results S3 bucket for me to see them as part of review.
	- I recommend leaving all defaults at 0.85 for now.
- The choice of `+2` indeed seems sort of arbitrary, which isn't necessarily a problem. But, since it is not a rigorously tested value, I recommend making it a notebook parameter in case this changes in the future after additional investigation.

sjspielman:
> I followed your advice to keep the predicted_celltype_threshold to 0.85 for now, however, re-running the script I realized we are missing quite a lot of immune and endothelium cells like this.

To clarify, what I meant by this is to keep the defaults in the R code as 0.85 (e.g., the inferCNV script argument and the notebook parameter), but you can assign 0.75 in the shell script `00_run_workflow.sh` and override those defaults. So, in the line where you created this variable in the shell script, you can make it equal to 0.75 so that value will  be used instead of 0.85.

analyses/cell-type-wilms-tumor-06/00_run_workflow.sh
```diff
@@ -147,12 +155,13 @@ for sample_dir in ${data_dir}/${project_id}/SCPCS*; do
...
     # Run inferCNV
-    Rscript scripts/06_infercnv.R --sample_id $sample_id --reference $reference --HMM i3 ${test_string}
+    Rscript scripts/06_infercnv.R --sample_id $sample_id --reference $reference --HMM i3 ${test_string} --score_threshold ${predicted_celltype_threshold}
 done
 
+
 # Render notebook to make draft annotations
 Rscript -e "rmarkdown::render('${notebook_dir}/07_combined_annotation_across_samples_exploration.Rmd',
-                        params = list(predicted.celltype.threshold = 0.85, cnv_threshold = 0, testing = ${TESTING}),
+                        params = list(predicted.celltype.threshold = ${predicted_celltype_threshold}, cnv_threshold_low = ${cnv_threhold_low}, cnv_threshold_high = ${cnv_threhold_high},  testing = ${TESTING}),
```
sjspielman, line 164: I think typo is causing the CI failure, based on the most recent run:

```suggestion
                        params = list(predicted.celltype.threshold = ${predicted_celltype_threshold}, cnv_threshold_low = ${cnv_threshold_low}, cnv_threshold_high = ${cnv_threshold_high},  testing = ${TESTING}),
```

maud-p: Thank you!

maud-p:
Dear @sjspielman ,

I just wanted to let you know that there will be 2 ~major changes to come tha will impact this PR:

- in `07_combined_annotation_across_samples_exploration.Rmd` in the second level annotation of normal cells. I understood why we were loosing immune and endothelial cells 🥳 in `06_infercnv.R` we have been indeed renaming the `immune` and `endothelium` that passed the `score_threshold` as `normal`. Thus, the selection of `immune` and `endothelium` cells lines 461-499 should be changed as the following
 ```{r 
cell_type_df <- cell_type_df |>
  mutate(second.level_annotation = case_when(
     ell_type_df$compartment %in% c("normal") &
      cell_type_df$cell_type %in% c("endothelial cell") ~ "endothelium",
    cell_type_df$compartment %in% c("normal") &
      cell_type_df$cell_type %in% c("B cell",
                                    "CD4-positive, alpha-beta T cell",
                                    "conventional dendritic cell",
                                    "lymphocyte",
                                    "macrophage",
                                    "mast cell",
                                    "monocyte",
                                    "natural killer cell",
                                    "neutrophil",
                                    "plasmacytoid dendritic cell") ~ "immune",

    .default = "unknown"
  ))
```

- another major change concern the `gene_order_file` defined in `06a_build-geneposition.R` and used for `infercnv`. I changed it to have a segmentation for each chromosome arms (p and q) per chromosome. I would indeed use this information to validate the infercv step. 

I made all these changes locally and re-ran the workflow. I am now exploring the results to make sure it is worth going into it.

I would then suggest to postpone this PR and open another one to first change the `06a_build-geneposition.R` and re-run the `06_infercnv.R` with the new `gene_order` and `predicted_celltype_threshold = 0.75`.

Then maybe once this is fixed we can re-open the PR#994 and finalize the remaining changes?

Thank you in advance!

sjspielman:
>  I understood why we were loosing immune and endothelial cells 🥳 in 06_infercnv.R we have been indeed renaming the immune and endothelium that passed the score_threshold as normal.

🥳🥳🥳 great catch!!

> I would then suggest to postpone this PR and open another one to first change the 06a_build-geneposition.R and re-run the 06_infercnv.R with the new gene_order and predicted_celltype_threshold = 0.75.
>
> Then maybe once this is fixed we can re-open the PR #994 and finalize the remaining changes?

That's fine with me! I was thinking that it might be fine to keep this PR open and come back to it later after merging in those other changes you describe, but the results might be different enough that a new PR is easier for you. So in any case - do what is easier for you :)

sjspielman:
As discussed in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1026, @maud-p will be opening a new PR to replace this one, so we can close this one now.

PR #1000, opened 2025-01-23
Remove outdated CNV annotation from CI in Ewing module

### LINKED ISSUE #992: Remove outdated workflows from CI for Ewing module
In the GHA that runs the workflow for the Ewing module, we are running a workflow that we don't use anymore, `cnv-annotation.sh`. This workflow runs `CellAssign`, `inferCNV`, and `copyKAT` so we probably don't need to waste resources on it if we aren't actually using any of those things for annotation. We should remove running this from CI.

allyhawkins:
Closes #992 

We are no longer using the `cnv-annotation.sh` workflow or any of the outputs for assigning cell types to the Ewing samples.  Since we don't need this workflow to work anymore, we don't need to run it in CI every time we make updates to that code. This should reduce our computational cost of running CI for this module by a lot. 

I thought about moving it from the root directory of this module and reorganizing things, but that seemed like overkill. Instead, I added a bunch of warnings both in the workflow script and in the docs stating that the workflow was not part of CI and was no longer maintained. For the README.md in the scripts directory I ended up moving the description of the scripts in this workflow to the end rather than being at the beginning. Let me know if you think I missed any areas where we should address this in the docs.

### DIFF
--- a/.github/workflows/run_cell-type-ewings.yml
+++ b/.github/workflows/run_cell-type-ewings.yml
@@ -47,7 +47,6 @@ jobs:
         run: |
           conda activate openscpca-cell-type-ewings
           cd $MODULE_PATH
-          bash cnv-annotation.sh
           bash aucell-singler-annotation.sh
           bash evaluate-clusters.sh
           bash run-aucell-ews-signatures.sh
--- a/analyses/cell-type-ewings/README.md
+++ b/analyses/cell-type-ewings/README.md
@@ -265,6 +265,8 @@ The `auc-ews-gene-signatures.tsv` file contains the following columns:
 
 ## CNV annotation workflow
 
+**NOTE:** This workflow is has been removed from CI and is no longer maintained! 
+
 The CNV annotation workflow (`cnv-annotation.sh`) can be used to identify potential tumor cells in a given sample.
 Annotations are obtained by running the following methods within the workflow:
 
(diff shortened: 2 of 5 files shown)


sjspielman (approved):
I can't find any other spots besides the ones you got to update, so this looks good! I'm happy to approve now, but I do think it would be good to also be explicit in the sentences/comments you've added that this workflow is not used for annotation at all, not only that it's not run in CI. 

What I mean is something simple like changing from this...
> THIS WORKFLOW HAS BEEN REMOVED FROM CI AND IS NO LONGER BEING MAINTAINED

to something like this....
> THIS WORKFLOW IS NO LONGER USED IN THE CELL TYPE ANNOTATION ANALYSIS. IT HAS BEEN REMOVED FROM CI AND IS NO LONGER BEING MAINTAINED

...in the 3 spots you added the caveat.

allyhawkins:
I condensed the statement you made in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1000/commits/587678e72549cfd1d66e4a686ce25fbd4adb7058 to be: 

`This workflow is no longer used in the cell type annotation analysis, has been removed from CI, and is no longer maintained! `

allyhawkins:
Canceling the ewing workflow since changes merged in from main are for a different module and this already passed checks.

Issue #993, opened 2025-01-16
Write a template notebook to use for "finalizing" cell type annotations for Ewing samples


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#696 

### Describe the goals of the changes to the analysis module.

After we use `AUCell` with our custom marker gene sets on all samples, we plan to go through the results from `AUCell`, the clustering workflow, and the `SingleR` workflow to come up with some sort of "final" annotations that include tumor cell states and normal cells. The plan is to do this for each library so that there is one notebook summarizing the results and annotations for each library in the Ewing project. Before we do that, we want to come up with a template of sorts that can be used to guide the notebooks for each library. 

Because there is some manual involvement in assigning cell types, we won't have a traditional template that gets rendered across each sample, but instead I'm thinking we want to come up with an example notebook and a set of functions that will help streamline a lot of the repetitive code that gets used. This would include turning a lot of the set up and plots into functions that we can use in each notebook. 

### What will your pull request contain?

I think this will be a series of PRs that would add in an example notebook and associated functions. Here's an outline of what I think would be present in the notebook: 

- Set up including reading in the SCE files and associated results from `AUCell`, `SingleR` and clustering. This would include prepping the data for plotting and combining all results into a single data frame. 
- Calculation of gene set scores for our small lists of marker genes (see `visser-all-marker-genes.tsv` and `tumor-cell-state-markers.tsv`). 
- Summary plots of results looking at all cells: 
  - UMAPs showing `SingleR` assignments, cluster assignments, AUC scores, and gene set scores 
  - Density plots showing AUC scores and gene set scores by cluster
  - Heatmap showing AUC scores and gene set scores with cluster annotation 
 - Using the above plots we would pull out the cells that are tumor and re-cluster and create updated plots using the new sub-clusters of tumor cells 
 - Assign tumor cell states and create combined plots with tumor cell states and normal cell annotations 
 - Validate these with the same UMAP, density plot, and heatmap combination 
 - Export the final.final.but.not.really annotations 

We would need functions for setup, calculation of gene set scores, and creating the plots. We have some of this already, but we should be sure to make them fit the needs of this notebook or write new ones. I think for each section of the notebook we should have a PR that adds those functions and an example version of that section to the example notebook. 

### Will you require additional software beyond what is already in the analysis module?

No, we should have everything we need already. 

### Will you require different computational resources beyond what the analysis module already uses?

[...]

PR #1008, opened 2025-01-28
Add summary plots to guide notebook for Ewing annotation 

allyhawkins:

Continuing on #993 

Here I'm filling in the section of the notebook that we will use to summarize the outputs from all the workflows, SingleR, clustering, AUCell and expression of custom marker genes. The goal is to be able to use these plots to help validate normal and tumor cell assignments. Then the next section will focus on adjusting any assignments based on these results and pulling out tumor cells for sub clustering and assignment of tumor cell states. 

- The first part looks at the SingleR and clustering results. Here I just show UMAPs of the results. Because we won't have a set number of clusters for each sample, it's hard to define a color palette. Because of that I show the faceted UMAPs with red as the cell type or cluster in that panel rather than different colors. Personally I find the faceted UMAP easier to see which cells belong to which cluster or which cell type, but let me know if you feel differently and want to see a single compiled UMAP. 
- The next section focuses on the output from AUCell using the gene sets from MsigDB. I included a faceted UMAP where cells are colored by the AUC, which I don't love, but maybe it will be helpful for some samples. I show the density plots of the AUC values, colored by whether or not AUCell classified that gene signature as being present or not. And then the last two plots look at AUC values across clusters and cell types in a density plot and heatmap. I know this is a lot of plots, but I find both the density and the heatmap helpful here. 
- The final section repeats the same plots used for AUCell but with the mean gene expression of our custom marker gene sets rather than the AUC values from MSigDB gene sets. Again, the UMAP isn't great, but I don't think it hurts to keep in. And then I included a density plot of each marker gene set colored by whether or not the cells are considered tumor or not tumor. 
- I added a few functions to a new script, `plotting-functions.R` that includes functions for creating plots that are repeated (density, UMAPs, heatmaps). I envision being able to use these functions in later sections of the notebook. 
- When possible, I used functions that we already use in other notebooks rather than write entirely new functions. 
- I made some small adjustments to the setup function based on issues I ran into while plotting. 

Yes, next up will be the section that looks at sub clustering of tumor cells. 

Here's a rendered copy of the notebook for review: 

[celltype-exploration.html.zip](https://github.com/user-attachments/files/18577085/celltype-exploration.html.zip)


analyses/cell-type-ewings/template_notebooks/celltype-exploration.Rmd
```diff
@@ -44,6 +44,10 @@ theme_set(
 
 # set seed
 set.seed(2024)
+
+# quiet messages
+options(readr.show_col_types = FALSE)
```
sjspielman, line 49: ugh it is so tempting to put this in my `~/.Rprofile` 😂

sjspielman (commented):
This seems fine to me overall, and I don't have many comments! Just some small items...

- I don't love `auc_assignment` as a label for the density plot where it appears since I don't think it's very informative. Maybe we don't even need to include AUCell in the label and just go with something like `in_geneset`?
- The density plot in the section `Mean expression of custom gene sets` needs to be bigger to see all the strip labels
- Some UMAPs have axis labels/ticks, and some don't. I'd remove these all around 
- I might again turn off the row clustering in the heatmaps - in other words, don't have the heatmap cluster the clusters. But if this view is more informative for you, or if you don't think it matters, then I wouldn't bother

allyhawkins:
@sjspielman I made all the changes you suggested in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1008#pullrequestreview-2582126964 in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1008/commits/f7829aaca8bef2e81d32a1e5d2e68d8f171cb766. This should be ready for another look! 

New copy of the report: 
[celltype-exploration.html.zip](https://github.com/user-attachments/files/18594036/celltype-exploration.html.zip)

sjspielman (approved):
LGTM!

PR #1010, opened 2025-01-28
Update consensus cell type assignment to output results for each library

### LINKED ISSUE #1006: Update consensus assignment to output results by sample

#853 

As mentioned in https://github.com/AlexsLemonade/OpenScPCA-nf/issues/114, we would like to modify the consensus cell type workflow to output a single TSV for each sample. Currently we output one single TSV file with all cells in ScPCA, so we should modify the existing scripts to output one TSV for each sample. 

This PR should modify the existing [`assign-consensus-celltypes.sh`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/assign-consensus-celltypes.sh) to output one TSV for each sample in ScPCA. To do this we should remove the script that combines data frames and then move the part of that script that adds consensus labels to the `03-save-coldata.R` script. Ultimately there will just be one script that reads in the SCE, grabs the colData, and assigns the consensus label. 

No 

No 

_No response_

allyhawkins:

Closes #1006 

Here I'm updating the existing workflow for assigning consensus cell types to create a single TSV file for each library rather than output one TSV file with all cells across all ScPCA samples. Just a note that the original issue says to output by sample, but I think in fact we do want to output by library and organize files so that they are in project/sample folders. 

The main change I made here is combining the contents of what was previously `03-save-coldata.R` and `04-combine-celltype-tables.R` into one single script, now named `04-assign-consensus-celltypes.R`. 

- I also incorporated some changes we had made when porting this to [`OpenScPCA-nf` to deal with cell lines as a sample type](https://github.com/AlexsLemonade/OpenScPCA-nf/blob/main/modules/cell-type-consensus/resources/usr/bin/save-coldata.R). 
- This single script now reads in the SCE file, grabs the annotation columns from the `colData`, and then assigns consensus cell types before saving the TSV file. 
- The `assign-consensus-celltypes.sh` workflow has also been updated to run that single script on every library in a single project. I think having the ability to run this on a single project will be helpful so we can easily use it when doing cell typing for an individual project. 
- To account for that change, I updated the GHA to run the workflow on all projects.  
 
Previously we were grabbing the blueprint reference directly from `celldex` and then mapping the ontology IDs to the human readable name in CL within the `04-combine-celltype-tables.R` script. It seemed like overkill to have to do this process for every single library now that everything is happening on a library level. 

- To avoid that, I wrote a new script that can be used to save the table with the Blueprint ontologies and cell type names. 
- I then saved that table to the repo, similar to how we save the panglao ontology reference file and consensus cell type file. 
- Since we are now saving that file, I updated the assigning script and script for creating the consensus ref to now take that file as input. 
- Because of these changes, I did change some of the numbers in the script file names to reflect the new order in which things should be run (prepare ontologies, prepare blueprint, create consensus, assign consensus). 

I also updated the readmes to reflect these changes. 

I also noticed the lock file was out of date and missing `patchwork` which is used in one of the exploratory notebooks. I added that here. 

Yes, but not workflow changes. Next up will be back to analysis. 

This produces a single TSV file for each library in a ScPCA project. Note that I did not actually run this on all ScPCA samples, since we will do that with OpenScPCA-nf. But I did test this on SCPCP000015 locally and things looked as expected. CI will test running this on test data for all projects. 

### Author checklists

_Check all those that apply._


analyses/cell-type-consensus/scripts/04-assign-consensus-celltypes.R
```diff
@@ -0,0 +1,171 @@
...
+
+# Set up -----------------------------------------------------------------------
+
+# make sure input files exist
+stopifnot(
+  "sce file does not exist" = file.exists(opt$sce_file),
+  "blueprint reference file does not exist" = file.exists(opt$blueprint_ref_file),
+  "panglao reference file does not exist" = file.exists(opt$panglao_ref_file),
+  "cell type consensus reference file does not exist" = file.exists(opt$consensus_ref_file),
+  "output file must end in `.tsv` or `.tsv.gz`" = stringr::str_detect(opt$output_file, ".tsv|.tsv.gz")
```
jashapiro, line 58: Making the test more specific.
```suggestion
  "output file must end in `.tsv` or `.tsv.gz`" = stringr::str_ends(opt$output_file, "\\.tsv|\\.tsv\\.gz")
```

analyses/cell-type-consensus/scripts/04-assign-consensus-celltypes.R
```diff
@@ -0,0 +1,171 @@
...
+# for now just combine sample ids into a single string and don't worry about demultiplexing 
+sample_id <- metadata(sce)$sample_id |> 
+  paste0(collapse = ";")
+project_id <- metadata(sce)$project_id
+
+# check if cell line since cell lines don't have any cell type assignments
+# account for having more than one sample and a list of sample types
+# all sample types should be the same theoretically
+sample_type <- unique(metadata(sce)$sample_type)
+is_cell_line <- sample_type == "cell line"
```
jashapiro, line 83: This can now be cell line and pdx, right? Do pdx cell lines have assignments?

analyses/cell-type-consensus/scripts/04-assign-consensus-celltypes.R
```diff
@@ -0,0 +1,171 @@
...
+  # now add in all the blueprint columns
+  dplyr::left_join(blueprint_df, by = c("singler_celltype_ontology" = "blueprint_ontology")) |>
+  # then add consensus labels
+  dplyr::left_join(consensus_ref_df,
+                   by = c(
+                     "singler_celltype_ontology" = "blueprint_ontology",
+                     "cellassign_celltype_annotation" = "original_panglao_name",
+                     "panglao_ontology"
+                   )
+  ) |>
```
jashapiro, line 165: Just formatting for consistency
```suggestion
  dplyr::left_join(
    consensus_ref_df,
    by = c(
      "singler_celltype_ontology" = "blueprint_ontology",
      "cellassign_celltype_annotation" = "original_panglao_name",
      "panglao_ontology"
    )
  ) |>
```

jashapiro (approved):
This looks good. My only real question was about the interaction with the recent changes for pdx samples. That won't affect this directly until we create a new data release, but it seems worth getting ahead of if we need to. 

I also had a comment about maybe making the test run a bit faster by not downloading all the test files, and a little suggestion about the regex you used.

All of this seems pretty minor, so I don't think I need to see it again, unless you have questions.

analyses/cell-type-consensus/scripts/04-assign-consensus-celltypes.R
allyhawkins: Ohh good catch. And no they don't. I'll update this to check for either just `cell line` or the combo.

PR #1013, opened 2025-01-30
06 infercnv update

maud-p:

This PR is following the [comment](https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/994#issuecomment-2621462463) from the PR#994 

-[ ] Major change: I modify the `infercnv` input `gene_order` to split the chromosoms into p and q arms. 
To do so, I updated the script `06a_build-geneposition.R`

The reason of this major change is that Wilms tumor have specific loss/gain of chromosome arms. I would like to see if having the arm information, I would be able to reproduce similar results as presented in [Cresswell et al. Fig 1A](https://www.biorxiv.org/content/10.1101/2024.09.04.610994v1). This would allow us to gain confidence in the `infercnv` step, which is a crucial step for the annotation.

-[ ] Minor change: I added a parameters at the top of `00_workflow.sh` to set the `predicted_celltype_threshold`.

`06_infercnv.R` has been re-ran with the new `gene_order` and `predicted_celltype_threshold`. Results will be updated to the s3 bucket.

Yes, finishing PR#994
and another related PR investigating more into details the `infercnv` results. 

researcher-008971640512-us-east-2

-[ ] in `results/reference`, the gene and arms position files, in a `.txt` format
-[ ] in `results/SCPCS{sample_id}`, the seurat object `06_infercnv_HMM-i3_{sample_id}_reference-both.rds` containing the output of `06_infercnv.R` and a `06_infercnv` subfolder with some outputs of `06_infercnv.R` that we decided to save. 

### Author checklists

_Check all those that apply._


sjspielman:
@maud-p I started reviewing this code, but some of the spacing in the script makes this difficult for me to read and run clearly. I was commenting in spots where I think this can be fixed to make the code easier for me to follow, but I decided it will actually be faster if I just update the spacing myself and file a PR to your branch. Then you can approve it and merge into this branch, and we'll continue with review here. Working on it right now!

sjspielman:
While working on this, I realized there was a way to dramatically simplify the code as well. But, while working on it, I noticed there are some discrepancies and I want to make sure the references you are using are from compatible genome annotations.

For example, consider this gene `ENSG00000171163`. 

In the `gene_order`, I see these coordinates:
```
   ensembl_id      chrom gene_start  gene_end
   <chr>           <chr>      <dbl>     <dbl>
 1 ENSG00000171163 chr1   249144205 249153343
```

But, in `chromosome_arms`, I see these coordinates for `chr1` overall:

```
   chrom                arm       start       end
   <chr>                <chr>     <int>     <int>
 1 chr1                 "p"           0 123400000
 2 chr1                 "q"   123400000 248956422
```

Therefore, it looks like the coordinates for this gene are higher than the end position for the q-arm. We used the Ensembl 104 annotation in the `scpca-nf` pipeline, so you'll want to find annotation files that are compatible with that reference. I see now that your files are at different versions - `v19` and `v29` - but the version of gencode we need is [`v38` which corresponds to Ensembl 104](https://genome.ucsc.edu/cgi-bin/hgTables?db=hg38&hgta_group=genes&hgta_track=wgEncodeGencodeV38&hgta_table=wgEncodeGencodeCompV38&hgta_doSchema=describe+table+schema). Therefore, we need to back up here and get the right files - I imagine this might also effect the cell type annotation results too!

What I suggest we do now is: You can please find and update the URLs in this script to point to a download for annotation files at the correct version. Alternatively, we can use the same GTF file that was used in the `scpca-nf` pipeline to find coordinates. This is publicly available from S3 - let me know if you want to go that route and I can help you with the code to download and/or parse it. 

Once we get the correct annotation files here, I can return to updating the script spacing to file a PR to you.

maud-p:
@sjspielman the workflow keeps failing as the changes made for `06_infercnv.R` impact the following notebook `07_combined_annotation_across_samples_exploration.Rmd`. Do we want to have the entire workflow running before merging this PR or could we still split the changes into 2 PR?

Most important, the script `06_infercnv.R` is now running with the new `gene position`. 🥳  The output on the first 3 samples do not show huge differencies, which is quite reinsuring, but still some differences. Will be running over night/weekend!

analyses/cell-type-wilms-tumor-06/scripts/06a_build-geneposition.R
```diff
@@ -1,25 +1,87 @@
 # This script downloads and adapt the genome position file for use with infercnv
 
+# load library
+library(tidyverse)
+library(stringr)
```
sjspielman, line 5: Not needed since included in tidyverse
```suggestion
```

analyses/cell-type-wilms-tumor-06/scripts/06a_build-geneposition.R
```diff
@@ -1,25 +1,87 @@
...
+if (!file.exists(gene_arm_order_file)) {
+  
+  # Download the updated gene order file for GENCODE v38 (Ensembl 104)
+  if (!file.exists(gene_order_file)) {
+    download.file(url = "https://ftp.ebi.ac.uk/pub/databases/gencode/Gencode_human/release_38/gencode.v38.annotation.gtf.gz",
+                  destfile = gene_order_file)
+  }
+  
+  # Read the GTF file and extract gene positions
+  gtf_data <- read.table(gzfile(gene_order_file), header = FALSE, sep = "\t", comment.char = "#", stringsAsFactors = FALSE)
```
sjspielman, line 27: This will be faster, but importantly, it uses `X` instead of `V` for column names. I do think you should use `read_tsv` as I suggest, but rather than changing everything from `V` -> `X`, instead rename the columns immediately to something informative, and then start working with the data. This will make code much easier to work with and maintain.

```suggestion
  gtf_data <- read_tsv(gene_order_file, col_names = FALSE, comment = "#")
```

analyses/cell-type-wilms-tumor-06/scripts/06a_build-geneposition.R
```diff
@@ -1,25 +1,87 @@
...
+  
+  # Filter for gene entries only
+  gene_order <- gtf_data %>%
```
sjspielman, line 30: Before entering this code, rename the data frame and then use those names directly - this is what you do for the cytoband file, and it would help here too.

analyses/cell-type-wilms-tumor-06/scripts/06a_build-geneposition.R
```diff
@@ -1,25 +1,87 @@
...
+  
+  
+  # Download chromosome arm information (cytoBand data)
+  if (!file.exists(arm_order_file)) {
+    download.file(url = "http://hgdownload.soe.ucsc.edu/goldenPath/hg38/database/cytoBand.txt.gz",
+                  destfile = arm_order_file)
+  }
+  
+  # Load cytoBand file
+  cytoBand <- read.table(gzfile(arm_order_file), header = FALSE, sep = "\t", stringsAsFactors = FALSE)
```
sjspielman, line 46:
```suggestion
  cytoBand <- read_tsv(arm_order_file, col_names = FALSE)
```

analyses/cell-type-wilms-tumor-06/scripts/06a_build-geneposition.R
```diff
@@ -1,25 +1,87 @@
...
+    summarize(
+      Start = min(start),
+      End = max(end),
+      .groups = "drop"
+    )
+  
+  # Add chromosome arm information
+  gene_order <- gene_order %>%
+    # Join chromosome arm information with gene_order based on Chromosome
+    left_join(chromosome_arms, by = c("Chromosome" = "chrom")) %>%
```
sjspielman, line 65: Something is wrong here with the joining because you are ending up with duplicate columns. This can be addressed, but it's not the only issue; there are no `p` arms after the join.
Also in the end here, it seems you never actually export the arm, but isn't that we we need for the eventual downstream?

sjspielman (commented):
This is definitely heading in the right direction and code is easier to read, so thank you! That said, there are still some issues with the parsing as I note in my comments. I'd like to try again to help refactor this to make sure we're exporting the correct information, so what I'd like to confirm with you here is: Which columns are needed for inferCNV and _how does this differ_ from what you'll need for the 07 notebook? Another way to ask, if we provide a column with `arm` to inferCNV, will it still work, or does that information need to be tracked separately?

As I noted in one of my comments, currently it doesn't seem that `arm` is ever exported, but I'm sure you'll want it since that's a reason we're doing this, right?

In terms of the CI, we can temporarily comment out that notebook code from running in `00_run_workflow.sh`, and revive it in the next PR. So, we should aim to get this to pass with the understanding that we'll have to update in the next PR.

analyses/cell-type-wilms-tumor-06/scripts/06a_build-geneposition.R
```diff
@@ -1,25 +1,87 @@
...
+arm_order_file <- file.path(module_base, "results", "references", "hg38_cytoBand.txt")
+gene_arm_order_file <- file.path(module_base, "results", "references", "gencode_v38_gene_pos_arm.txt")
+
+if (!file.exists(gene_arm_order_file)) {
+  
+  # Download the updated gene order file for GENCODE v38 (Ensembl 104)
+  if (!file.exists(gene_order_file)) {
+    download.file(url = "https://ftp.ebi.ac.uk/pub/databases/gencode/Gencode_human/release_38/gencode.v38.annotation.gtf.gz",
+                  destfile = gene_order_file)
+  }
```
sjspielman, line 24: If we're going to use a GTF file directly here, we should go ahead and use the exact one used with `scpca-nf` that prepared the data you're working with. That is publicly available from S3 and can be obtained with this command:
```
aws s3 cp s3://scpca-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.gtf.gz {LOCAL TARGET DESTINATION} --no-sign-request
```

We should go ahead and replace the `download.file` with a system call to sync the file. I can potentially also take care of this in my refactor.

analyses/cell-type-wilms-tumor-06/scripts/06a_build-geneposition.R
maud-p: True, in the next version I'll simplify this. The only information that I want to position a gene on the p or q arm is the centromere position = `Start` of the q arm = `End` of the p arm. If I only keep the p arm from the `chromosome_arms` before joining with `gene_order`, the centromere information will be in the `End` column and I avoid duplicating columns. Will be updated soon-ish 😄 

I export the arm information when I do `mutate(Chromosome = paste0(Chromosome, arm))`, because for `infercnv`, the `gene_order` input is a table with 4 columns: the gene name/ID, the chromosome info to segment the heatmaps (in our case chromosome+arm) and the start/end position of the gene.

analyses/cell-type-wilms-tumor-06/scripts/06a_build-geneposition.R
maud-p: I now added the downloading command in the `00_run_worflow.sh`, but not sure if this is the right place or the right way of doing it, please let me know 😄

maud-p:
@sjspielman I just see now that you worked on it yesterday and open a PR. I'll try to include your changes with my changes now! Thank you!

sjspielman:
This PR has been replaced with https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1026, so we can close this now.

PR #1015, opened 2025-01-30
Freeze ubuntu runners at 22.04

### LINKED ISSUE #1014: Change standard GHA runner to `ubuntu-22.04`
> Based on the error and the lass successful run, I'm wiling to bet this is due to GHA changing `ubuntu-latest` from ubuntu 22 to 24. We should probably update our requested runner across the board to ubuntu 22, at least for a while. 

 _Originally posted by @jashapiro in [#1011](https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1011#issuecomment-2624582309)_

I would expect some instability as the change from `ubuntu-22.04` to `ubuntu-24.04` rolls out; it probably makes sense to make these standard and change as needed.

jashapiro:
Closes #1014
(and maybe #1011)

Here I am updating the GHA runners for most of the `run_*` actions to `ubuntu-22.04`. 

I decided that actions that run with docker containters could probably safely be left with `ubuntu-lastest`, and also that the `docker_*` actions were also safe to leave that way. I also left the `hello-python` runner as `latest`, maybe as a kind of bellweather. If that breaks, we have some real trouble that we probably want to know about.

### DIFF
--- a/.github/workflows/run_cell-type-ETP-ALL-03.yml
+++ b/.github/workflows/run_cell-type-ETP-ALL-03.yml
@@ -32,7 +32,7 @@ on:
 jobs:
   run-module:
     if: github.repository_owner == 'AlexsLemonade'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     defaults:
       run:
         shell: bash -el {0}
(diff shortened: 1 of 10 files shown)


jashapiro:
Well, that wasn't it... it seems to be a problem with installing the `parallelDist` package, and that error persists with `ubuntu-22.04`. Not sure where that leaves this PR/issue as a good or bad idea.

jaclyn-taroni (approved):

PR #1016, opened 2025-01-31
Look at consensus cell types in PDX vs. non-PDX samples

### LINKED ISSUE #1003: Look at percentage of cells classified in PDX and non-PDX samples in ScPCA

#853 

We have a hunch that the consensus cell type assignments in PDX are largely "unknown" and cell types are not able to be identified. This is probably because a) we are only using a human reference and any non-tumor cells here will be from mouse and/or b) PDX samples generally are comprised of mostly tumor cells and very few normal cells. We should confirm this suspicion prior to doing any more analysis and remove these samples from downstream analysis if that is the case. 

A notebook that compares the percentage of cell types classified in PDX vs. non-PDX across all samples. We may just be able to add a new section to the existing exploratory notebook of the results since this should not be very involved. 

No 

No 

_No response_

allyhawkins:
Closes #1003 

We want to pull out the PDX samples specifically and look at the consensus cell types that are assigned. We had noticed when looking at the original plots that projects with PDX samples might show differences in consensus cell type distribution based on if the sample came from patient tissue or PDX. In particular, we want to know if we should continue to use the PDX samples as we dig deeper or if we might consider throwing them out. 

This PR does a little bit of exploration to confirm that in some projects we do see very few cell types identified in PDX vs. non-PDX samples, but this is not the case for all of them. 

I updated the existing notebook that we had looking at consensus cell types to help answer this question. Since we are deciding to focus in on a few projects, I figured we could dive deeper into either of those projects as needed outside of this initial exploratory notebook. 

- I added a plot that looks at only projects with PDX tumors and compares the percentage of cells classified as "unknown" between PDX and patient tissue samples. 
- I updated the existing stacked bar charts to facet by sample type. This made it easy to compare the composition of the cell types classified between patient tissue and PDX samples for each project. 

Next, we'll dive into the osteo projects. 

Rendered exploratory notebook, which I'm putting here for review: 
[02-explore-consensus-results.nb.html.zip](https://github.com/user-attachments/files/18624258/02-explore-consensus-results.nb.html.zip)

- For the ALL project and NB project, we do see pretty much no cells classified in the PDX samples compared to the patient tissue samples. 
- For all the sarcomas: RMS, soft tissue sarcoma, Ewing sarcoma, and osteo, we see a handful of PDX samples that do have cells that are identified. However, they are generally identifying muscle cells and fibroblasts with the occasional endothelial cell. 
- Wilms, RB, "other", and rhabdoid don't really have any cells that are classified regardless of sample type. 
- If cells are classified, they are stromal cell types and not immune cells, which makes sense since the PDX are probably from immune-compromised mice. 

I do think one caveat with all of this is that we only use a human reference for quantifying gene expression... I think if there were any non-malignant cells to begin with, we are generally going to have a lower likelihood of being able to capture them unless there is high homology between the transcripts in human and mouse. I know that some percentage of the time we will be able to capture a non-malignant cell because the amount of overlap in the transcripts is not 0, but it's definitely not as accurate as using a combined human/mouse reference. 

I think I would be wary of including the PDX in additional analyses because of this caveat. 


jaclyn-taroni (commented):
The noted limitation is enough to leave PDX samples out of further analyses, in my opinion. I think the analysis included here is reasonable.

jashapiro (approved):
LGTM. I agree with @jaclyn-taroni that it makes sense to not include these in the consensus analysis.

Issue #1017, opened 2025-01-31
Use merged object for Ewing sarcoma samples to assign cell types


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#696 

### Describe the goals of the changes to the analysis module.

Before we spend a lot of time trying to manually validate annotations in each sample (via #1003), I think it might be useful to explore using the merged object to finalize annotations. We have some samples that are heterogeneous and have both tumor cells and a set of normal cells, while most samples are almost exclusively tumor. Working with the homogenous samples is difficult when they all show high expression of tumor markers. But if we were to look at all cells together we would expect to see a more obvious distinction between tumor and normal cell types. In particular I would like to run `AUCell` on the merged object with the gene sets that we have identified from MSigDB and look at that output alongside both the cell types obtained by running `SingleR` with the tumor cell reference and consensus cell types. 

One caveat here is that we don't want to integrate the samples, but just work with the merged and uncorrected data. So it is possible that we might see some technical differences. We should pay close attention to the normal cells in this case since I expect those to be more similar to each other across samples than tumor cells. 

### What will your pull request contain?

This is going to be broken into a few PRs. 

- The first PR will add running the merged object through `AUCell` as an additional step to the script that currently runs `AUCell` on each individual object. In looking at the [`AUCell` vignette](https://www.bioconductor.org/packages/devel/bioc/vignettes/AUCell/inst/doc/AUCell.html) there is a note that rankings could be combined after running independently, but it's best to be sure the same genes are being considered and proceed with caution. Because of this, I think we should just re-run on the merged object rather than use existing results. 
- The second PR will be a script to read in the merged object, all singleR results, all consensus cell types, and all AUC results. The output will be a TSV file with UMAP embeddings, cell type annotations, and AUC scores for each cell across all datasets. This can then be used as input to a notebook rather than having to work with the full merged object. 
- The last PR will be an exploratory notebook that looks at the cell type assignments and AUC values in the merged object. I plan to use the guide notebook that I've been developing as part of #993 to create this notebook. The goal is to adjust any assignments based on findings in this notebook and validate the assignments. 

### Will you require additional software beyond what is already in the analysis module?

No, we should have everything but we will need to use the merged object from `OpenScPCA-nf`. 

### Will you require different computational resources beyond what the analysis module already uses?

No 

### If known, when do you expect to file the pull request?

Hoping to have these items done by middle of next sprint (2/14).

PR #1023, opened 2025-02-03
Add merged object to AUCell workflow

allyhawkins:

Addresses #1017 

I would like to be able to look at cell type assignments on all samples together rather than trying to assign cell types individually. To do that we want to look at the merged object, in particular we want to have AUCell results from the merged object. 

This PR makes the necessary modifications to the workflow and script that runs `AUCell`.

- I added an `is_merged` option to deal with any special circumstances for the merged object. The formatting is slightly different so this affects filtering the SCE object to remove any non-detected genes. 
- I also noticed that the matrix wasn't quite in a sparse form? It's `sparse DelayedMatrix object of type "double"` so is showing the 0s and then causes an error when used as direct input to `AUCell`. I'll file an issue in `scpcaTools` where we make the merged object, because I think we do want to make sure the output there is actually a `dgCMatrix`. 

Yes. I already started looking at the results here to make sure this was a reasonable thing to spend time on. So next will be a PR that looks at the results in an exploratory notebook using the template that we have as a guide to create that notebook. 

`s3://researcher-211125375652-us-east-2/cell-type-ewings/results/aucell-ews-signatures`

There should be TSV files with the results. I re-ran the whole workflow and copied the updated files for individual libraries and the merged library to S3. 

Coming next! 


allyhawkins:
Hmm it looks like GHA doesn't have the credentials to download from the workflow results bucket?
See https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/runs/13121720493/job/36609019909?pr=1023

jaclyn-taroni:
I suspect the more accurate statement is that `download-results.py` relies on listing objects, which is not allowed. cc: @jashapiro

jashapiro:
I think you are missing the `--test-data` flag for the `download-results.py` script, so it is trying to use the true results, which require credentials.

analyses/cell-type-ewings/scripts/aucell-ews-signatures/01-aucell.R
sjspielman, line 1: While I was here I noticed there is probably a bug on line 98 - should be `dirname()` not `dir()`

analyses/cell-type-ewings/scripts/aucell-ews-signatures/01-aucell.R
```diff
@@ -96,8 +102,19 @@ fs::dir_create(output_dir)
 sce <- readr::read_rds(opt$sce_file)
 
 # remove genes that are not detected from SCE object 
-genes_to_remove <- rowData(sce)$detected > 0 
-filtered_sce <- sce[genes_to_remove , ]
+if(!opt$is_merged){
+  genes_to_remove <- rowData(sce)$detected > 0
```
sjspielman, line 106: switching name since logic
```suggestion
  genes_to_keep <- rowData(sce)$detected > 0 
```

analyses/cell-type-ewings/scripts/aucell-ews-signatures/01-aucell.R
```diff
@@ -96,8 +102,19 @@ fs::dir_create(output_dir)
...
 
 # remove genes that are not detected from SCE object 
-genes_to_remove <- rowData(sce)$detected > 0 
-filtered_sce <- sce[genes_to_remove , ]
+if(!opt$is_merged){
+  genes_to_remove <- rowData(sce)$detected > 0 
+} else {
+  
+  # if merged object then need to sum all columns to find genes not present in the object at all 
+  genes_to_remove <- rowData(sce) |>
```
sjspielman, line 110:
```suggestion
  genes_to_keep <- rowData(sce) |> 
```

analyses/cell-type-ewings/scripts/aucell-ews-signatures/01-aucell.R
```diff
@@ -96,8 +102,19 @@ fs::dir_create(output_dir)
...
+  
+  # if merged object then need to sum all columns to find genes not present in the object at all 
+  genes_to_remove <- rowData(sce) |> 
+    as.data.frame() |>
+    dplyr::select(ends_with("detected")) |> 
+    rowSums() > 0
+  
+}
+
+filtered_sce <- sce[genes_to_remove , ]
```
sjspielman, line 117:
```suggestion
filtered_sce <- sce[genes_to_keep , ] 
```

sjspielman (approved):
Looks good! Suggested a few changes, but don't need to see again.

I'd recommend updating the `scripts/README` usage as well here: 
https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/31c786e697efa75b724c0cb3ecde688b056e4d8b/analyses/cell-type-ewings/scripts/README.md?plain=1#L188-L194

Also, while I was finding this line in that README, I came across this line which has the wrong extension - it should be `.R` not  `.Rmd`:
https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/31c786e697efa75b724c0cb3ecde688b056e4d8b/analyses/cell-type-ewings/scripts/README.md?plain=1#L303-L309

Issue #1007, opened 2025-01-28
Create a release: v0.2.1


allyhawkins:
In order to complete https://github.com/AlexsLemonade/OpenScPCA-nf/issues/114, we need a versioned release of the cell-type-consensus Docker image, which means a versioned release of this repository.

**Pre-release checklist:**

- [x] Any mentions of the release tag in the repository have been updated
- [x] All maintenance workflows are passing
  - [x] [Run all analysis modules](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/workflows/run_all-modules.yml) (trigger manually if needed)<br>
  Check that all analysis modules are present in the workflow before running, and file a PR updating the workflow if necessary.
  - [x] [Spellcheck](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/workflows/spellcheck.yml) (trigger manually if needed)
  - [ ] [Code styling](https://github.com/AlexsLemonade/OpenScPCA-analysis/actions/workflows/code-styling.yml) (optional, trigger manually)
- [x] All blocking issues have been resolved
- [x] Write release notes and add them to [`CHANGELOG.md`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/CHANGELOG.md), which should include the following:
  - Which modules have been added, if any?
  - Which modules have been deprecated, if any?
  - What has changed in the repo documentation, if anything?
  - Are there any _major_ shifts in project dependencies?
    For example, is there a package that was used throughout Docker environments that has been replaced?
  - Have there been any changes in repo file organization?
- [x] [Create a release on GitHub](https://github.com/AlexsLemonade/OpenScPCA-analysis/releases/new)
  - populate the contents with the release notes added to the changelog.

allyhawkins:
Release is done, see https://github.com/AlexsLemonade/OpenScPCA-analysis/releases/tag/v0.2.1

PR #1024, opened 2025-02-03
Update changelog to include v0.2.1

allyhawkins:
Related to #1007 

Here I added a changelog entry for v0.2.1. There isn't a lot to report since this only includes some minor updates to the consensus and Ewing module. I mentioned that this version of the consensus module has been ported over to OpenScPCA-nf, is that okay? 

I also mentioned the addition of the Seurat and gene symbol FAQs, since I think those are relevant changes for contributors to know about.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ This workflow is now present in `OpenScPCA-nf`, and results will be made availab
 
 New FAQs covering how to work with ScPCA data as `Seurat` objects and using gene symbols instead of Ensembl IDs have been added to the [OpenScPCA documentation](https://openscpca.readthedocs.io/en/latest/troubleshooting-faq/faq/). 
 
+`Rcpp` packages were updated for compatibility. 
+
 ## v0.2.0
 
 This release adds the first set of community-contributed analyses to the repository.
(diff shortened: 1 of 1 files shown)


jashapiro:
I think I would mention that `Rcpp` packages were updated for compatibility. (with what exactly, we don't know, so we can just leave that there)

allyhawkins:
> I think I would mention that `Rcpp` packages were updated for compatibility. (with what exactly, we don't know, so we can just leave that there)

I added a sentence in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1024/commits/5779b32321f2e39305e4339e5ef47779e8f8754e, should I elaborate more?

CHANGELOG.md
```diff
@@ -9,6 +9,20 @@ Add new release notes in reverse numerical order (newest first) below this comme
...
+
+- `cell-type-ewings`
+- `cell-type-consensus`
+
+In particular, the workflow for assigning consensus cell types has been created in `cell-type-consensus` module. 
+This workflow is now present in `OpenScPCA-nf`, and results will be made available to all contributors. 
+
+New FAQs covering how to work with ScPCA data as `Seurat` objects and using gene symbols instead of Ensembl IDs have been added to the [OpenScPCA documentation](https://openscpca.readthedocs.io/en/latest/troubleshooting-faq/faq/). 
+
+`Rcpp` packages were updated for compatibility.
```
jashapiro, line 24: Maybe just a bit more detail.
```suggestion
`Rcpp` package versions within modules were updated for compatibility. 
```

jashapiro (approved):
LGTM

PR #1028, opened 2025-02-06
Explore consensus cell types in osteosarcoma samples

### LINKED ISSUE #1004: Explore consensus cell types in all osteo projects

#853 

We would like to compare the cell types identified in all osteosarcoma samples/ projects. In particular, we are interested in if we can identify immune "Hot" and "cold" tumors across these samples. 

This PR will add a notebook that looks at all samples from all three osteo projects (17, 18, and 23) in more detail. We should look at the overall distribution of cell types across all these samples and compare the percentage of immune cells vs. other cells. Ultimately, can we identify groups of samples that are immune "Hot" or "cold"? 

No 

No 

_No response_

allyhawkins:

Closes #1004 

The goal is to look at the consensus cell types assigned across all osteosarcoma samples found in three different projects and identify any variation across samples. In particular, we want to look for samples with high and low immune infiltrate (hot and cold tumors). What I'm adding here is pretty basic and just summarizes the consensus cell types and total immune percentage. My understanding is we want to be able to make an observation about samples with different immune populations here, but any biological insight is going to be additional work outside the scope of this PR. 

- First I just looked at the top cell types across all samples in a single plot, similar to the stacked plots we had seen previously. The overall cell types present are pretty similar, but the total percentage of those cell types changes for each sample. 
- I also looked at the top cell types separated by project. One thing to note is that generally project 23 has a lot fewer cells that are classified than both 17 and 18. Also it looks like 18 has more T cells, while 17 has more macrophages. Endothelial cells and smooth muscle cells are pretty consistent across samples if cells are classified.
- I looked at total immune percentage vs. non-immune vs. unknown. Again, there are definitely variations across samples, but I think part of that is some samples having more cells that are classified to begin with. 
- I arbitrarily broke up "hot" and "cold" tumors based on having at least 5% immune cells. This is totally arbitrary, but just to show that we do have different groups of tumors. 
- Lastly I looked at immune composition alongside some of the clinical metadata including primary/mets, disease timing, tissue location and seq unit. There may be some difference in single-cell/nuclei but sample numbers are not comparable. I could definitely do the stats to see if this is the case, but I think if we want to do any real comparisons we might want to look into methods on comparing cell composition in single-cell. 

While I was working on this I also moved the setup function that we used in the first exploratory notebook to it's own R script to source in and added a list of just immune cell types as a reference. 

Yes

Rendered HTML report here: 

[02-explore-consensus-results.nb.html.zip](https://github.com/user-attachments/files/18698953/02-explore-consensus-results.nb.html.zip)

Different samples have different proportions of classified cells and immune cells and this seems to be mostly driven by project. I think the most interesting thing here (besides showing that cell type composition is different in general) is that the cell types obtained is project specific. I'm not sure we want to highlight that... but I think there's definitely some technical variation that can impact which cell types are identified in the first place. 

I think we need to do more rigorous statistical analysis to make any other real conclusions beyond that at this point. 

Are there things missing that you want to see at this stage? 
I generally don't feel great about this, but I'm not sure what else to do before diving down a rabbit hole, so I thought I would stop and get some feedback. 

Note that I did look at just T cell infiltrate since that's been shown to be found in osteo samples before and I didn't see anything worth noting, but I can add that section back in if you would like to see it. 


jaclyn-taroni (commented):
To give some general scientific feedback here:

> I arbitrarily broke up "hot" and "cold" tumors based on having at least 5% immune cells. This is totally arbitrary, but just to show that we do have different groups of tumors.

I would take this out. I think the fact that there is heterogeneity is pretty apparent without arbitrarily picking a threshold.

> Lastly I looked at immune composition alongside some of the clinical metadata including primary/mets, disease timing, tissue location and seq unit. There may be some difference in single-cell/nuclei but sample numbers are not comparable. I could definitely do the stats to see if this is the case, but I think if we want to do any real comparisons we might want to look into methods on comparing cell composition in single-cell.

What I think we should aim for – and we can discuss this in our normal planning process – is using the labels to generate a hypothesis someone could test.

analyses/cell-type-consensus/exploratory-notebooks/03-osteosarcoma-consensus-celltypes.Rmd
```diff
@@ -0,0 +1,424 @@
...
+
+Below I look at the immune cell percentage stratified by each of the mentioned metadata. 
+This is just a preliminary look and if we really want to dig into this further we should consider looking at algorithms for comparing cell type composition in single-cell RNA-seq data, but that is outside the scope of this notebook. 
+
+```{r}
+# do some data wrangling on the metadata columns we want to look at 
+immune_results_df <- immune_results_df |> 
+  dplyr::mutate(
+    primary_or_metastasis = dplyr::if_else(!is.na(primary_or_metastasis), primary_or_metastasis, disease_timing),
+    disease_timing_mod = dplyr::if_else(disease_timing %in% c("Initial diagnosis", "Recurrence"), disease_timing, "other"),
```
jashapiro, line 350: I think we want to add ordering here so the plots are in expected orders. I also am not sure why we seem end up with only "Recurrence" sample in the `primary_or_metastasis`? (Since it is only one, I think it could be dropped, maybe?)

jashapiro (commented):
Overall, I think this is a pretty good exploration, so I don't think I have much in the way of major comments. One note is that the hot vs. cold (generally) is pretty significantly intertwined with project, which makes it very hard to interpret the relative contribution of things like cell vs. nucleus, as those categories don't appear to be evenly distributed between projects. 

I made some coding suggestions to try to remove some of the warnings you are seeing, and to reduce joins, though the latter changes are less important.

allyhawkins:
@jaclyn-taroni and @jashapiro thanks for your feedback. I removed the hot/cold section as requested and then made some minor plot updates and code changes based on reviews. 

> I think we want to add ordering here so the plots are in expected orders. I also am not sure why we seem end up with only "Recurrence" sample in the primary_or_metastasis? (Since it is only one, I think it could be dropped, maybe?)

I'm not totally sure what you mean be expected order, but I assume you meant primary first, initial diagnosis first, etc? So I made that change. As far as the "Recurrence" sample that was showing up it was because for one project the primary/mets information is actually in the `disease_timing` column instead of a separate column. Then for one of those samples, instead of `primary` or `metastasis` it says `Recurrence`. I updated the code to make that sample `NA` and then remove it prior to making the plots. 

Other than that I left everything pretty much the same since we are looking more for a qualitative overview. Here's the updated notebook: 
[03-osteosarcoma-consensus-celltypes.nb.html.zip](https://github.com/user-attachments/files/18742011/03-osteosarcoma-consensus-celltypes.nb.html.zip)

jaclyn-taroni:
I don't think I need to take another look at this, so I'm going to remove my review request.

jashapiro (approved):
This looks good to me, with just a small suggestion about the labels for the plots.

allyhawkins:
Note that I cancelled the workflow to run the module here since these changes are only exploratory notebooks and not run in CI.

Issue #1033, opened 2025-02-10
Check expression of expected marker genes for consensus cell types


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#853 

### Describe the goals of the changes to the analysis module.

The goal is to create a plot (or group of plots) to confirm expected marker genes are expressed in the appropriate cell types as determined by the consensus cell type labels. It would be beneficial to create a plot summarizing the expression of key marker genes across all cells in ScPCA. Alternatively, we could examine the data across different diagnoses and/or projects.

### What will your pull request contain?

- First we will need to look at the cell types that we expect and assign each of those cell types to a broad category (e.g., all T cell subtypes should just be T cell). 
- Then we need to create a list of expected markers for each of those broad categories. I think having 3-5 genes per category would be sufficient.
- Then we can use this information to create a dotplot and/or heatmap showing marker gene expression across cells assigned to each cell type. If we can make one plot that summarizes all cells in ScPCA let's do that without it looking wild, let's do that. 

### Will you require additional software beyond what is already in the analysis module?

No 

### Will you require different computational resources beyond what the analysis module already uses?

No 

### If known, when do you expect to file the pull request?

This sprint

allyhawkins:
Once #1056 goes in, we will want to write a script that renders that notebook across all projects and then save the output HTML files to the repo. Once that's done then this issue should be able to be closed.

PR #1044, opened 2025-02-14
Add a table with possible top marker genes to use for validation of consensus cell types

allyhawkins:

Towards #1033

The goal of this PR is to add a table of the top marker genes we expect to see present in each of the possible consensus cell types. Ultimately, we want to be able to look at expression of these marker genes across cells in a cell type and see that they are expressed higher in that cell type over all other cells.  

The first thing I did was assign some broader cell classes to the consensus cell types. There are 52 possible consensus cell types, many of which are from the same category of cells (e.g., T cell subtypes, different endothelial cells, etc). I condensed this list down to 22 possible cell types based on how similar the cells are to other cells. When doing this I created a new file, `consensus-validation-groups.tsv`, where each row is one of the possible consensus cell types. Then I added a column for `validation_group_ontology` and `validation_group_annotation`. I grouped things based on the granularity that I think makes sense and used the LCA from that group to assign the ontology ID and name. A few notes: 

- We have `lymphocyte of B lineage` and then a bunch of B cells. Technically, `lymphocyte of B lineage` is a parent of B cell so that term would be the LCA term for all the B cells, but I feel like we want to use the B cell term. Also when identifying marker genes using CellMarker (described below), that's the one term that doesn't have a matching cell type in that database. So I chose to group that with the B cell term. 
- I made a myeloid group which has most of the general myeloid and then granulocytes, but I pulled out monocyte and macrophage. I think those are two cell types that are going to be frequently observed and they have distinct markers that we could utilize. 
- I wasn't entirely sure what to do about the neural cell types. I grouped them all together (including neuron, astrocyte, and glial cells) into `neural cell`, but I could be convinced to separate them. I just am the least familiar with those groups so I'm not sure how easy it will be to distinguish those cell types with marker genes. 

The next thing I did was compile the list of markers for each of the groups. At first I tried to do this manually, and then I realized that was crazy. So, I searched for existing marker gene lists and then I remembered that we use one for cell typing! We have a marker gene list from PanglaoDB that we could use, but this list has 100s of marker genes for each cell type, and it might be kind of circular to use the same list of marker genes that we used to run `CellAssign` in the first place. 

This led me to [CellMarker2.0](http://117.50.127.228/CellMarker/index.html) where you can download a list of all genes identified as marker genes for each cell type in all human tissues assayed. The best part is that they have cell ontology IDs associated with their cell types, so I decide to utilize this resource to create a table of marker genes. 

- I wrote a small script to download the file and save it to the repo but ignored it because its pretty big and has thousands of genes in it. It's the [`Human cell markers` from their downloads page](http://117.50.127.228/CellMarker/CellMarker_download.html). This file contains all the tissues assayed and all the cells for each tissue and their marker genes so a lot of cell types are present multiple times and have a slightly different marker gene list depending on the tissue, so we still have to do some work to get the marker genes that could be useful. 
- So, I wrote a script that reads in that marker gene file and the validation groups and creates a table of marker genes we can use to create figures downstream. I do the following in the script: 
  - Read in the CellMarker file and filter to only include marker genes found in the cell types corresponding to our validation group assignments. 
  - Count how many tissues each cell type is observed in 
  - Count how many tissues each gene for each cell type is observed in 
  - Calculate the percentage of tissues that a given marker gene shows up in a given cell type
  - I used this percentage to rank the marker genes for each cell type and grabbed the top 10 genes for each one. In some cases there's a tie, usually when there are very few tissues with that cell type, and in that case all the tied marker genes are also included. 

I picked top 10 to keep the number of genes somewhat manageable, but I think depending on how plots look we could always come back to this and change that number or alter it so that we take all genes that appear in that cell type in more than X percentage of tissues. We also might see that some marker genes just aren't informative. There are some cases with cell types that are only present in a handful of tissues, so I am less confident in the genes listed and we might choose not to use them later. My hope is that with doing it this way, we can start making plots that look at expression of these marker genes and then modify as needed. 

Yes, next I plan on actually plotting these in ScPCA samples. 

- Do we agree on the groups for cell type assignments in `consensus-validation-groups.tsv`? 
- Are we okay with this approach for identifying possible marker genes? 


analyses/cell-type-consensus/references/consensus-validation-groups.tsv
```diff
@@ -0,0 +1,53 @@
...
+CL:0000113	mononuclear phagocyte	CL:0000763	myeloid
+CL:0000764	erythroid lineage cell	CL:0000763	myeloid
+CL:0000556	megakaryocyte	CL:0000763	myeloid
+CL:0000771	eosinophil	CL:0000763	myeloid
+CL:0000623	natural killer cell	CL:0000623	natural killer cell
+CL:0000540	neuron	CL:0002319	neural cell
+CL:0000127	astrocyte	CL:0002319	neural cell
+CL:0000125	glial cell	CL:0002319	neural cell
+CL:0000126	macroglial cell	CL:0002319	neural cell
+CL:0000095	neuron associated cell	CL:0002319	neural cell
```
jaclyn-taroni, line 41: Is the problem your own knowledge of marker genes, or is it a problem with the sources (e.g., CellMarker)? I think we should try to be more specific if we can, but I also don't know how often these more granular categories are getting assigned, which may mean it is not worth it.

analyses/cell-type-consensus/references/consensus-validation-groups.tsv
```diff
@@ -0,0 +1,53 @@
...
+CL:0000723	somatic stem cell	CL:0000034	stem cell
+CL:0000034	stem cell	CL:0000034	stem cell
+CL:0000499	stromal cell	CL:0000499	stromal cell
+CL:0000327	extracellular matrix secreting cell	CL:0000499	stromal cell
+CL:0000084	T cell	CL:0000084	T cell
+CL:0000624	"CD4-positive, alpha-beta T cell"	CL:0000084	T cell
+CL:0000791	mature alpha-beta T cell	CL:0000084	T cell
+CL:0002419	mature T cell	CL:0000084	T cell
+CL:0000815	regulatory T cell	CL:0000084	T cell
+CL:0000813	memory T cell	CL:0000084	T cell
```
jaclyn-taroni, line 53: New line:
```suggestion
CL:0000813	memory T cell	CL:0000084	T cell

```

analyses/cell-type-consensus/references/consensus-validation-groups.tsv
```diff
@@ -0,0 +1,53 @@
...
+CL:0000126	macroglial cell	CL:0002319	neural cell
+CL:0000095	neuron associated cell	CL:0002319	neural cell
+CL:0000669	pericyte	CL:0000669	pericyte
+CL:0000786	plasma cell	CL:0000786	plasma cell
+CL:0000723	somatic stem cell	CL:0000034	stem cell
+CL:0000034	stem cell	CL:0000034	stem cell
+CL:0000499	stromal cell	CL:0000499	stromal cell
+CL:0000327	extracellular matrix secreting cell	CL:0000499	stromal cell
+CL:0000084	T cell	CL:0000084	T cell
+CL:0000624	"CD4-positive, alpha-beta T cell"	CL:0000084	T cell
```
jaclyn-taroni, line 49: I know these quotes are here because of the comma, but are they likely to cause problems downstream?

analyses/cell-type-consensus/scripts/05-generate-validation-markers.R
```diff
@@ -0,0 +1,98 @@
...
+  dplyr::top_n(10, percent_tissues) |> 
+  # arrange genes by cell type and percentage of tissues 
+  dplyr::arrange(validation_group_ontology, desc(percent_tissues)) |> 
+  # bring in ensembl ids
+  dplyr::left_join(mapped_ids, by = c("gene_symbol")) |>
+  # match to human readable name for ontology group 
+  dplyr::left_join(consensus_groups_df, by = c("validation_group_ontology")) |> 
+  dplyr::relocate(
+    validation_group_annotation, ensembl_gene_id, .after = validation_group_ontology
+  )
```
jaclyn-taroni, line 94: Should we be concerned about the specificity of marker genes, too? This only considers how often a gene shows up for a cell type across tissues, but what if the same gene shows up for many cell types? I don't know that we want that for our purposes. CellMarker might be built with this in mind (I don't recall), or we can ameliorate my concern by demonstrating this doesn't come up in practice.

jaclyn-taroni (commented):
The code looks good. I have mostly science comments.

- If the more granular neural cell types get assigned, I think we should try to do a bit better there. 
- What about the _specificity of marker genes_? I think this could be fully addressed by demonstrating your approach doesn't produce marker gene lists with unacceptable levels of overlap between cell types in practice.

analyses/cell-type-consensus/references/consensus-validation-groups.tsv
allyhawkins: I don't know how often they are getting assigned, but I did look at the possible combinations within the original consensus reference and how they are getting assigned and used that to make some updates. 

- Astrocytes are an available option in both panglao and celldex, and subsets of astrocytes are available in panglao so there are multiple ways to get this assignment. Because of that, I'm including those as their own category. 
- For glial cells, the celldex annotation is always going to be astrocyte, but the panglao annotation is another glial cell type (like microglia). There's only one way to get macroglial, which is to be labeled as astrocyte in celldex and oligodendrocyte in panglao. Because of that, I'm going to group both glial and macroglial together as just glial cell. 
- There are no markers for [neuron associated cell](https://www.ebi.ac.uk/ols4/ontologies/cl/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FCL_0000095?lang=en) in CellMarker. I looked at the possible combinations for that label and they are again all astrocytes in celldex and then one of [neuron associated cell (sense Vertebrata)](https://www.ebi.ac.uk/ols4/ontologies/cl/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FCL_0000123?lang=en), so I think we should group those with "glial cell". 

I updated the table in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1044/commits/7e5aa3676ab2fc19182f511f7459f5af4234ec47

analyses/cell-type-consensus/scripts/05-generate-validation-markers.R
allyhawkins: To address this I added a column for `gene_observed_count` which notes how many times that gene shows up in the table and therefore how many cell types it's found in. For now I didn't do any further filtering on this column, but we may want to in the future when we go to actually make the figure. 

For some additional context, most cell types have at least 2 marker genes that are specific to that cell type. I included a table with the total counts of unique marker genes per group below. 
Screenshot 2025-02-17 at 10 25 04 AM

jaclyn-taroni (approved):
This seems good for now. I say "for now" because I anticipate a fair bit of effort filtering genes and ordering genes (e.g., if immune cells share a gene) when it's time to make the plots these are intended to help with.

PR #1048, opened 2025-02-18
Fix cell-type-wilms-tumor-06 in CI, attempt 1

sjspielman:
Towards #1040 

The most recent full run of `cell-type-wilms-tumor-06` has been failing. I cannot reproduce the bug locally, including within the Docker container, but I caught some items that either _might_ be the culprit, or can help point us in the right direction which I'm addressing here:

- First, we had a seed bug in the inferCNV script where the seed was character instead of integer. 
- Second, we have a slightly more stringent threshold now for identifying normal cells to use in the inferCNV reference. I bumped this to a much less stringent threshold to use during testing only.
- Third, unrelated to bug, I noticed a script was accidentally missing from CI, so I added it!


sjspielman:
Ah, I know the problem. I'm going to open a fresh PR, since I was slightly wrong about the fix here.

PR #1049, opened 2025-02-18
Attempt 1: Fix wilms module CI

sjspielman:
Towards #1040 

After much trying to reproduce and inability to reproduce, I realized that we have a seed bug which is being treated as a character.... Making it be an integer is resulting in some reproducible runs locally, and hopefully this seed will do it. After this goes in, we'll be able to re-run the full module workflow and see 🤞 

While here, I also added some comments to emphasize that we are holding onto a deprecated script that isn't used anymore.


allyhawkins (approved):
🚀

PR #1050, opened 2025-02-18
Save gene expression for validation markers when assigning consensus labels

allyhawkins:

Towards #1033 

Before I can actually make any plots to validate marker gene expression in assigned cell types, I need to be able to get the gene expression for all of the genes in [validation-markers.tsv](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/validation-markers.tsv) across all samples. 

I thought the quickest way to do this was to modify the existing workflow for assigning consensus cell types to also output a file with the `logcounts` for all the marker genes. Then I can read in all of the individual files to create a larger data frame that I can use to create the validation plots just like we do for the consensus cell type assignments.

- I modified the existing script for assigning consensus cell types to also take as input a marker gene file. Now both the consensus output TSV is saved along with the gene expression TSV. I originally had this as a separate script, but the most time consuming part is reading in the SCE object and I felt it was silly to have to read in the SCE object twice for every library?
- The output TSV file has a columns for library ID, barcodes, ensemble id, gene symbol, and gene expression, where gene expression is from the `logcounts `of the SCE. Note that I do not include the other columns from the `validation-markers.tsv`, such as the validation annotation and ontology. This is because many marker genes are present multiple times and I don't think we need to store information for a given gene more than once. 
- All genes in `validation-markers.tsv` that are also expressed in the SCE object are included in the output. 

Yes. 
I plan on running this through a couple of projects and then moving to a notebook where I'll read in the gene expression TSVs and consensus TSVs and create the validation plots mentioned in #1033. 

I haven't actually run this through more than just one project, but once approved, I plan to run it through a few projects and use that to get started on making the plots. At that point, I'll add results to S3 when appropriate.  

TSV file with columns for library ID, barcodes, ensemble id, gene symbol, and gene expression, where gene expression is from the `logcounts `of the SCE. 

Coming next! 

Are we okay with this happening within the same script that saves the consensus labels?

### Author checklists

_Check all those that apply._

### DIFF
--- a/analyses/cell-type-consensus/renv.lock
+++ b/analyses/cell-type-consensus/renv.lock
@@ -76,6 +76,13 @@
       ],
       "Hash": "07a8b7f7a8a23998324a40eab02a44e7"
     },
+    "BH": {
+      "Package": "BH",
+      "Version": "1.87.0-1",
+      "Source": "Repository",
+      "Repository": "CRAN",
+      "Hash": "468d9a03ba57f22ebde50060fd13ba9f"
+    },
     "Biobase": {
       "Package": "Biobase",
       "Version": "2.66.0",
@@ -132,6 +139,25 @@
         "utils"
       ],
       "Hash": "3aec5928ca10897d7a0a1205aae64627"
+    },
+    "BiocParallel": {
+      "Package": "BiocParallel",
+      "Version": "1.38.0",
+      "Source": "Bioconductor",
+      "Repository": "Bioconductor 3.19",
+      "Requirements": [
+        "BH",
+        "R",
+        "codetools",
+        "cpp11",
+        "futile.logger",
+        "methods",
+        "parallel",
+        "snow",
+        "stats",
+        "utils"
+      ],
+      "Hash": "7b6e79f86e3d1c23f62c5e2052e848d4"
     },
     "BiocVersion": {
       "Package": "BiocVersion",
@@ -809,6 +835,21 @@
       ],
       "Hash": "73361f44874bfcecf54f7ba9238612d1"
     },
+    "beachmat": {
+      "Package": "beachmat",
+      "Version": "2.20.0",
+      "Source": "Bioconductor",
+      "Repository": "Bioconductor 3.19",
+      "Requirements": [
+        "BiocGenerics",
+        "DelayedArray",
+        "Matrix",
+        "Rcpp",
+        "SparseArray",
+        "methods"
+      ],
+      "Hash": "10e94b1bce9070632a40c6b873f8b2d4"
+    },
     "bit": {
       "Package": "bit",
       "Version": "4.5.0.1",
@@ -936,6 +977,16 @@
         "utils"
       ],
       "Hash": "3f038e5ac7f41d4ac41ce658c85e3042"
+    },
+    "codetools": {
+      "Package": "codetools",
+      "Version": "0.2-20",
+      "Source": "Repository",
+      "Repository": "CRAN",
+      "Requirements": [
+        "R"
+      ],
+      "Hash": "61e097f35917d342622f21cdc79c256e"
     },
     "colorspace": {
       "Package": "colorspace",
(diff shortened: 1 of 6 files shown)


jashapiro (commented):
This looks good overall, and the approach seems reasonable to me. 

My suggestions are mostly small, with the biggest one designed to push the effort of matrix manipulation onto the package developers, rather than doing all of it manually.

allyhawkins:
> My suggestions are mostly small, with the biggest one designed to push the effort of matrix manipulation onto the package developers, rather than doing all of it manually.

Note that I took this suggestion but had to make some small changes. I still had to add the `library_id` column since we don't have that in our `colData` and then I selected only the columns that we want to keep. 

I also noticed that we aren't actually saving `gene_symbol` so I removed that from the docs. I don't think we need it. 

This should be ready for another look.

allyhawkins:
> > My suggestions are mostly small, with the biggest one designed to push the effort of matrix manipulation onto the package developers, rather than doing all of it manually.
> 
> Note that I took this suggestion but had to make some small changes. I still had to add the `library_id` column since we don't have that in our `colData` and then I selected only the columns that we want to keep.
> 
> I also noticed that we aren't actually saving `gene_symbol` so I removed that from the docs. I don't think we need it.
> 
> This should be ready for another look.

Although doing this means I need to add `scuttle` and dependencies to the lock file and Docker image... I'll file a PR to do that

analyses/cell-type-consensus/scripts/04-assign-consensus-celltypes.R
```diff
@@ -169,4 +200,54 @@ all_assignments_df <- celltype_df |>
...
+    features = expressed_markers, 
+    assay.type = "logcounts",
+    use.dimred = FALSE
+  ) |>
+    tidyr::pivot_longer(starts_with("ENSG"), names_to = "ensembl_gene_id", values_to = "logcounts") |> 
+    # only pull out relevant columns
+    dplyr::select(barcodes, ensembl_gene_id, logcounts) |> 
+    # add library id column
+    dplyr::mutate(library_id = library_id) |> 
+    dplyr::relocate(library_id, .before = 0)
```
jashapiro, line 248: Mutate has it's own `.before` arg.
```suggestion
    dplyr::mutate(library_id = library_id, .before = 0)
```

analyses/cell-type-consensus/scripts/04-assign-consensus-celltypes.R
```diff
@@ -169,4 +200,54 @@ all_assignments_df <- celltype_df |>
...
+  # get logcounts from sce for expressed genes 
+  gene_exp_df <- scuttle::makePerCellDF(
+    sce, 
+    features = expressed_markers, 
+    assay.type = "logcounts",
+    use.dimred = FALSE
+  ) |>
+    tidyr::pivot_longer(starts_with("ENSG"), names_to = "ensembl_gene_id", values_to = "logcounts") |> 
+    # only pull out relevant columns
+    dplyr::select(barcodes, ensembl_gene_id, logcounts) |>
```
jashapiro, line 245: To save a bit of memory, I think it is probably better to get just the barcodes when making the data frame and save yourself the select. 
```suggestion
    features = expressed_markers, 
    assay.type = "logcounts",
    use.coldata = "barcodes",
    use.dimred = FALSE
  ) |>
    tidyr::pivot_longer(starts_with("ENSG"), names_to = "ensembl_gene_id", values_to = "logcounts") |> 
```

jashapiro (approved):
LGTM, with just small suggestions, though I suspect that merging in `main` may require conflict resolution in the `renv.lock` file (which should be to just take the version from `main`).

analyses/cell-type-consensus/renv.lock
jashapiro, line 1: All of the additions here look like they are duplicates.

analyses/cell-type-consensus/scripts/04-assign-consensus-celltypes.R
allyhawkins: Ah sorry, I misread the docs and thought it was TRUE/FALSE for the use.coldata option, but I'll update it!

PR #1052, opened 2025-02-19
Testing 1040 fixes

sjspielman:
This is a draft PR I am opening for investigating #1040 further, since I cannot locally reproduce the error. Hoping comparing CI here to the local log can help me find where to fix this.

I updated the workflow to run only the sample throwing the error, and some print statements to explore. More changes expected in this PR as a testing playground, and importantly _this PR is not expected to undergo review_.


sjspielman:
This is now passing CI, so we have a fix. I will open a separate PR to get this in more cleanly, after doing #1053.

PR #1054, opened 2025-02-19
Add scuttle to consensus cell typing Docker

allyhawkins:
In #1050 we want to be able to use `scuttle`. Before we can do that, we need to add `scuttle` to `renv` and the Docker image. This PR does that.

### DIFF
--- a/analyses/cell-type-consensus/Dockerfile
+++ b/analyses/cell-type-consensus/Dockerfile
@@ -27,9 +27,6 @@ RUN Rscript -e "install.packages('renv')"
 # Copy the renv.lock file from the host environment to the image
 COPY renv.lock renv.lock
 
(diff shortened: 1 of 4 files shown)


jashapiro (approved):
I guess I was expecting this was already installed. I didn't mean to make work!

jashapiro:
There was a mismatch in the versions of Bioconductor for the new packages. I added the corrected versions in [db182f4](https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1054/commits/db182f4d083afd82050c34bf174f3eef3b5b643f), and also updated `renv` to get off the version which had the strange `object` bug. Though that meant a big change to the `renv.lock` file due to version changes.

allyhawkins:
@jashapiro thanks for fixing that!

PR #1055, opened 2025-02-20
Patches for cell-type-wilms-tumor-06

### LINKED ISSUE #1040: Analysis module failing in CI

### LINKED ISSUE #1053: Update wilms-tumor-06 Docker to have awscli
Recently, we added code in `cell-type-wilms-tumor-06` to sync an ScPCA reference file from S3 to use when preparing inferCNV input. This aspect of the code runs fine in CI, but it turns out that's only because we actually are downloading `awscli` in the GHA! When run standalone in a docker container, we don't have this tool.
https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/e4dc422dd8ddcdc89d031c289fbd123748db500c/.github/workflows/run_cell-type-wilms-tumor-06.yml#L46-L50

This issue therefore tracks updating the Dockerfile for this module to include a step to install `awscli`.

sjspielman:
Closes #1040 
Closes #1053 

This PR implements 1 bug fix and 1 GHA fix in the cell-type-wilms-tumor-06 module.

### Bug fix (#1040)

inferCNV was failing on a test data sample because the Bugs MCMC was not detecting any CNV regions. There is no way to check for this, since its a fully internal step within a single command to run inferCNV and the information can't be known _a priori_. To get around this in CI, I updated the chromosome regions file that we use for inferCNV when in CI. We recently updated this chromosome file to also consider `p` and `q` arms to nail down some CNVs more carefully, but this wasn't playing nicely with test data. So, in CI, we should just not consider the arms since the sample sizes get to small to run inferCNV reliably through.

### Other fix (#1053)

Recently, we had updated the reference files used to create that very chromosome regions file to ensure we're using a genome reference that matches scpca-nf. Originally in my reviews, I suggested to just grab this from our S3 bucket. But, I realized while diagnosing the other bug, this runs in CI only by accident since the GHA installed `awscli` to run `download-data.py`, but the code would fail in a standalone Docker container. So, the plan for #1053 was to add this awscli dependency into Docker. When I went to start that though, I thought that ehhh maybe that's not the move since fewer dependencies are better, and I just swapped in the Ensembl link for the same file so we avoid `awscli` altogether.

### DIFF
--- a/.github/workflows/run_cell-type-wilms-tumor-06.yml
+++ b/.github/workflows/run_cell-type-wilms-tumor-06.yml
@@ -55,7 +55,8 @@ jobs:
           # These samples were selected either because:
           # - they are the subset of samples explored when evaluating CNV methods, so they must be present to run the module workflow in full if that is ever specified
           # - they do not use a reference with inferCNV so we would like to test that they are properly handled
-          DOWNLOAD_FLAG: ${{ github.event_name != 'pull_request' && '--projects SCPCP000006' || '--samples SCPCS000179,SCPCS000184,SCPCS000194,SCPCS000205,SCPCS000208,SCPCS000177,SCPCS000190' }}
+          # - they have been known to cause CI failures in edge cases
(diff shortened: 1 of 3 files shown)


sjspielman:
The sample causing the error in #1040 passed CI here!

allyhawkins (approved):
LGTM!

PR #1056, opened 2025-02-20
Add exploration notebook looking at validating consensus cell types

allyhawkins:

Towards #1033 

The goal of this PR is to look at using the marker genes added in #1044 to validate the assignment of consensus cell types in ScPCA. I originally was going to try and look at all ScPCA samples together, but I thought it might be better to first start with identifying some sort of approach to doing this with a single project and then apply that across all projects. 

So here I am adding an exploratory notebook that creates a few different versions of validation plots for a project. I set it up so that we could easily render the notebook across all projects and then save the results to the repo. For any additional plots where we look at all cells in a given "diagnosis group", I think we can do separately in the paper figures repo. There we can finalize color assignments for each cell type and decide on how many cell types we should show. 

- First I took the `consensus_annotation` for all cells in SCPCP000001 and assigned them to a broader cell type group to use for plotting. These correspond to the values in `validation_group_annotation` in https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/consensus-validation-groups.tsv. This is the `Broad cell type annotations` found on the y-axis for all plots. 
- I am only including any cell types that have > 50 cells assigned to a given annotation. For this project there are ~28000 cells, so that's less than 1% of cells. 
- I looked at expression of the validation markers across all cells in a broad group in two ways: 
  - Mean expression of a single marker gene across all cells in the same group. For this I filter to only look at marker genes that are specific to that validation group before plotting.
  - Mean expression of all marker genes for a given validation group. Here I first calculate the mean expression of each marker gene across all cells and then take the mean of that value across all marker genes. 
-  The dot plots only show dots if the mean expression is > 0 or the percent of cells that gene is expressed in is greater than 5%. The heatmaps show all genes, with no filtering. 

Yes, I think we should render this report across all projects once it's finalized and then save the results for posterity. 

HTML report: 
[05-marker-gene-validation.nb.html.zip](https://github.com/user-attachments/files/18894981/05-marker-gene-validation.nb.html.zip)

I think we did a pretty good job!! 
At least for this project, I see distinct groups of normal cells that express the genes they are supposed to express. 

### Author checklists

_Check all those that apply._


jashapiro (commented):
Overall this looks good. 

The one major thing I might change would be to be a bit more deliberate about the order of the cell type annotations; having annotations that are expected to be more similar near each other (e.g., neurons & astrocytes; monocytes, macrophages, & myeloid cells) seems like it would make some the of the shared expression patterns a bit easier to see/interpret. 

I do think the dot plots are generally sufficient; I like that they show two axes in one with both the percent expressed and level of expression. I did suggest that if you do want to add in the "all genes" plot, that might work better as a heatmap, so you could have both...

Code-wise, there is a lot of data frame manipulation going on, which can get a bit confusing to try to figure out without a good knowledge of the formats. I suggested a couple places where the adding in some comments about expected formats might be helpful for interpretation. 

Random additional thought: I wonder if it makes sense to split the `unknown` by diagnosis? It probably doesn't matter for this project as most are glioblastoma, but maybe for some of the others it could be informative.

jaclyn-taroni (commented):
I agree with Josh about being more deliberate about the ordering of cell types and prefer the dot plot to the heatmaps.

allyhawkins:
In short, I updated the cell type order. Here's a new copy of the notebook: 
[05-marker-gene-validation.nb.html.zip](https://github.com/user-attachments/files/18948102/05-marker-gene-validation.nb.html.zip)


> The one major thing I might change would be to be a bit more deliberate about the order of the cell type annotations; having annotations that are expected to be more similar near each other (e.g., neurons & astrocytes; monocytes, macrophages, & myeloid cells) seems like it would make some the of the shared expression patterns a bit easier to see/interpret.

Totally agree about this! Since the ordering is determined based on the input marker gene table, I updated the script to make that table to output it using a specific order where cell types belonging to similar groups are grouped together. This should make it easy so that no matter what project we are looking at and which cell types are included, similar ones will always be together. 

> I do think the dot plots are generally sufficient; I like that they show two axes in one with both the percent expressed and level of expression. I did suggest that if you do want to add in the "all genes" plot, that might work better as a heatmap, so you could have both...

I was mostly including both just to see two different options, but I agree the dot plots are sufficient! I ended up updating this to show the dot plot with unique genes only, the heatmap with all genes, and then the dot plot with mean expression of all genes. 

Just another note that the unique genes here were determined when creating the marker genes table using Cell Marker. The `gene_observed_count` tells us how many cell types that gene is a marker gene for, so we only show genes where that number is 1. 

> Code-wise, there is a lot of data frame manipulation going on, which can get a bit confusing to try to figure out without a good knowledge of the formats. I suggested a couple places where the adding in some comments about expected formats might be helpful for interpretation.

I'm so sorry about this. I tried my best to make things simple since yes it's a lot. The biggest thing that's confusing is that we have to account for both the assigned cell type groups (which I've named `broad_celltype_group`) and the validation cell type group for all marker genes (`validation_group_annotation`). I added some comments throughout that will hopefully help clear things up, but let me know if any of the data manipulation is still confusing.

> Random additional thought: I wonder if it makes sense to split the `unknown` by diagnosis? It probably doesn't matter for this project as most are glioblastoma, but maybe for some of the others it could be informative.

Hmm... I do like this idea but I do feel like generally the diagnosis are very similar in a project, so I'm not sure it would be all that informative. I don't expect a lot of variation in the marker gene expression for normal cell types between tumor types that are similar.

jaclyn-taroni (commented):
This looks good from my perspective, but I am going to let @jashapiro have the final say.

jashapiro (approved):
This all looks good to me. I am only mildly annoyed by the some of the legend positions for the heatmap, but that level of detail is not worth worrying about at this time (and doing it "right" looks like it is pretty annoying).

allyhawkins:
> This all looks good to me. I am only mildly annoyed by the some of the legend positions for the heatmap, but that level of detail is not worth worrying about at this time (and doing it "right" looks like it is pretty annoying).

If it makes you feel any better, I too am annoyed but felt it was not worth the time to fix, especially if we really prefer the dot plot anyways.

allyhawkins:
No changes to the workflow here, so canceling the run.

PR #1065, opened 2025-02-26
Add data.table and complexHeatmap to renv for cell-type-consensus

allyhawkins:
When running CI in #1064, I'm seeing failures because we don't have `data.table` in `renv`. This adds that and some other packages that were missing from when I added the exploratory notebook, like `complexHeatmap`.


jashapiro (approved):
LGTM

PR #1068, opened 2025-03-03
Update readme for consensus cell type references

### LINKED ISSUE #1067: Update references/README in consensus cell type module
Currently, the [`references/README`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/README.md) for the consensus cell type module is missing some column documentation.

In particular, the column `original_panglao_name` appears in [cell type reference table](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv) but is not documented. 

In addition, I would recommend using maybe L4 headers instead of numbers in the readme, since the tables are wider than the tabbing for the numbered list and it's a bit hard to see where each list item starts and stops.

allyhawkins:
Closes #1067 

Here I'm updating the README file in the `cell-type-consensus/references` folder to add the missing column described in #1067 and remove the numbering. I added L4 headers as suggested and fixed some other small typos.


sjspielman (approved):
Thanks!!

PR #1071, opened 2025-03-10
CI updates for cell-type-wilms-tumor-06

### LINKED ISSUE #1070: Update code to pass CI: annotation notebook in wilms-tumor-06
Following https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1030, the `cell-type-wilms-tumor-06` notebook which accomplishes the final module cell type annotation will have been updated to include a refined annotation strategy.

Because of test data limitations, however, this notebook does not run in CI. This is because we previously had to turn off use of chromosome arm information for inferCNV to get _that_ running in CI, which in turn prevents _this_ from running in CI.

To deal with this, we could turn off chunks which need chromosome arm information, _unless_ this effectively means no aspect of the notebook is run in CI since such code is too pervasive. An alternative approach could be to create some fake chromosome arm data columns to join into the data when testing is true. That fake data would be created and added within this notebook for maximum transparency, as well as a description in the notebook of why this step is needed.

sjspielman:
Closes #1070 

This PR aims to get the cell-type-wilms-tumor-06 module passing. For this initial fix, I'm opening as a draft PR to test it. 

The issue is chromosome arms are not used in testing, but they are integrated into plots. When looking on my own with fresh eyes here, turns out that fixing this may have indeed been easier than I thought and we just need to use an empty string for the "arm" plotting when testing. Hopefully this does the trick!

**Edit**: Passed CI, ready for review! Here is some additional context to support review:

- Back in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1055, we had to update this module to pass CI because of inferCNV failures on Ubuntu. The failure was caused by module changes to use a genome file for inferCNV that considered chromosome arms in addition to chromosomes themselves. To pass CI, we turned off arms in testing to only consider chromosome number. 
- Since then, the `07` notebook which produces the final annotations was updated to use the chromosome arm information: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1055. But, this did not pass in CI, since we turned off arms in CI. I therefore let this PR go in with that notebook commented out and planned to come back on my own and fix the CI issue. 
- This PR gets that notebook running in CI! Turns out the fix was pretty quick - there are a few plots which directly reference gain/loss columns for chromosome arms. In testing, I basically just remove the p/q so it becomes a chromosome again. I had to do this four times; here's one example on its own:
https://github.com/sjspielman/OpenScPCA-analysis/blob/85faa99e3d575b45234cdc7c8a854e4ab936e17b/analyses/cell-type-wilms-tumor-06/notebook/07_combined_annotation_across_samples_exploration.Rmd#L530-L536
- The additional changes to the notebook include:
  - many spacing changes from the `pre-commit` styler. i recommend turning off whitespace for review!
  - i removed a definition for `data_dir`, which isn't used. It was left over from the OpenScPCA RMarkdown template.


sjspielman:
@maud-p FYI, this is the PR which fixes the `07` annotation notebook, so everything will pass in CI :)

PR #1084, opened 2025-03-20
Add CNV references and Ewings cell type exploration

### LINKED ISSUE #1081: Add ewings CNV reference information

Part of #1080 

This issue tracks adding information about known Ewings sarcoma CNVs which we can use for inferCNV validation to the module that will be created in #1078. The goal is just to get this information all in one place in this module for future reference. 

Some known CNVs are listed here to start: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/5a605070b0808ff3a1c70d7ec8a5cb8468dd82ea/analyses/cell-type-ewings/exploratory_analysis/04-infercnv.Rmd#L418

TSV and/or README of CNV information. 

No

No

_No response_

### LINKED ISSUE #1082: Explore/tabulate ewings consensus cell type distribution

Part of #1080

The goal of this analysis module is to do some initial exploratory analysis on the immune consensus cell types present for Ewings (SCPCP000015) samples. We'd like get a sense of how many immune cells there are per sample (both count and fraction of total cells), and we would also like to get a sense of which immune cell types are most prevalent. 

This information will help 

An exploratory notebook to explore the immune cell type distribution in ewings samples. The notebook should ideally conclude with recommendations for approaches we can test for pooling cells in next steps to run inferCNV.

No

No

_No response_

sjspielman:

Closes #1081 
Closes #1082

This PR addresses two issues:

- For #1081, I add some documentation for the CNV reference validation into the README - any other recommendations where this might go? Nothing else really made sense so me since it's such a brief blurb, but still one we want in the module somewhere
- For #1082, I added a notebook to look at immune cell counts.
  - It does look like we have sufficient immune cells to use, and could possibly subset to macrophages or T cells for a more cohesive normal reference
  - But, there are indeed some cells (not many, but some!!) across a few projects that consensus labeled as immune and your ewings module labeled as tumor - good call on checking this! 
- I made a few other updates as well:
  - I added a module bash script and activated the workflow
  - I updated into about input data in the README:
    - I had been hopeful we might get a single run out of `inferCNV` with the merged object, but indeed `inferCNV` [doesn't want that](https://github.com/broadinstitute/infercnv/issues/170#issuecomment-522641397). So, we need the individual processed SCE.
      - Edit: Now I have found this other comment which suggests it could be worth a go with merged - https://github.com/broadinstitute/infercnv/issues/330#issuecomment-919276952. For near-future us, it might be worth seeing how different results are for merged vs individual samples.
    - Since the consensus cell types are also in the ewings cell type results, we can just pull in the ewings results for this project and the consensus TSVs specifically aren't needed

There are no S3 results, here is the rendered HTML: 
[01_ewings-consensus-cell-types.nb.html.zip](https://github.com/user-attachments/files/19369169/01_ewings-consensus-cell-types.nb.html.zip)


analyses/infercnv-consensus-cell-type/run-analysis.sh
```diff
@@ -0,0 +1,21 @@
...
+module_dir=$(dirname "${BASH_SOURCE[0]}")
+cd ${module_dir}
+
+# Define directories
+notebook_dir="notebooks"
+
+##### Analysis for SCPCP000015 #####
+
+# Explore distribution of cell types
+Rscript -e "rmarkdown::render('${notebook_dir}/01_ewings-consensus-cell-types.Rmd')"
```
allyhawkins, line 21: This is just an exploratory notebook, so in my opinion it does not need to be run in CI.

analyses/infercnv-consensus-cell-type/notebooks/01_ewings-consensus-cell-types.Rmd
```diff
@@ -0,0 +1,186 @@
...
+
+```{r}
+celltype_df |>
+  dplyr::filter(category == "immune") |>
+  dplyr::count(ewing_annotation) |>
+  dplyr::filter(stringr::str_detect(ewing_annotation, "tumor"))
+```
+
+Well, it appears there are some consensus immune cells which were classified as `tumor`. 
+Which samples are these in?
```
allyhawkins, line 158: Ugh, I should have checked this in the Ewings module. I think it would be good to check the expression of tumor markers vs. macrophage/T cell markers in these cells and ensure we have them classified correctly. I feel like that's something we should do in the Ewings module in a notebook rather than here. I would probably agree that these should be excluded from the normal reference that we use.

analyses/infercnv-consensus-cell-type/notebooks/01_ewings-consensus-cell-types.Rmd
```diff
@@ -0,0 +1,186 @@
...
+We'll just divide the total number of cells by this count, not taking into consideration how many immune cells are actually in that sample, to get a rough sense of how many immune cells there are relative to all.
+
+```{r}
+celltype_df |>
+  dplyr::count(library_id, name = "total_cells") |>
+  dplyr::mutate(fraction_immune = round(total_immune / total_cells, 2)) |>
+  knitr::kable()
+```
+
+For the low-quality sample `SCPCL001111` this fraction isn't very meaningful since there are fewer cells in the sample than there are pooled immune cells.
```
allyhawkins, line 107: We should probably just exclude this sample from the analysis?

analyses/infercnv-consensus-cell-type/notebooks/01_ewings-consensus-cell-types.Rmd
```diff
@@ -0,0 +1,186 @@
...
+      stringr::str_detect(consensus_annotation, "T cell"),
+      "T cell type",
+      consensus_annotation
+    )
+  ) |>
+  dplyr::filter(consensus_annotation %in% c("T cell type", "macrophage")) |>
+  dplyr::count(consensus_annotation)
+```
+These constitute the vast majority of cells, so it would also be an option to use just these groups of cell types.
+If we were to want a single immune cell type, then macrophage is the only suitable option to get more than 100 cells.
```
allyhawkins, line 142: I don't think we want to use a single immune cell type, but choosing the most dominant immune cell types and comparing that to using all immune cells makes sense to me.

analyses/infercnv-consensus-cell-type/README.md
```diff
@@ -33,3 +38,18 @@ This module uses `renv` and `docker` to manage software dependencies.
...
+### Ewing sarcoma
+
+* Gain of chr1q [1,2]
+* Gain of chr8 [1,2]
+* Loss of chr16q [1]
+* Gain of chr12 [1]
+
+Sources:
+1. https://doi.org/10.1158/2159-8290.CD-14-0622
+2. https://doi.org/10.1158/2159-8290.CD-13-1037
```
allyhawkins, line 55: I might turn this into a table instead. That will make it easier to read into a notebook/script. I imagine we are going to look specifically at the CNV scores for these chromosomes.

allyhawkins (commented):
I think this is generally good for what you are trying to do here. My biggest comment is to move the CNV information into a table. You could have columns for `diagnosis`, `chromosome`, `gain_or_loss`, `arm`, and `source`. 

I think once this goes in, I would create two normal references - one with all immune and one with only macrophages and T cells and then run inferCNV using both of those references. Then we can explore those results.

analyses/infercnv-consensus-cell-type/notebooks/01_ewings-consensus-cell-types.Rmd
sjspielman: >  I feel like that's something we should do in the Ewings module in a notebook rather than here. I would probably agree that these should be excluded from the normal reference that we use.

Ok, so I will leave this as some exploration/updates for you to potentially do in the Ewings module, and for now just proceed to exclude these.

analyses/infercnv-consensus-cell-type/notebooks/01_ewings-consensus-cell-types.Rmd
sjspielman: 💯

sjspielman:
> I think once this goes in, I would create two normal references - one with all immune and one with only macrophages and T cells and then run inferCNV using both of those references. Then we can explore those results.

I agree with this, and I have a couple other thoughts here for things we might like to try out - will open an issue to discuss! 

I believe I've implemented all the reviews here, so should now be ready for another look. The changes I made to the notebook are all text, so I didn't attach a new html here since review should be the same looking at the diff.

analyses/infercnv-consensus-cell-type/exploratory-notebooks/01_ewings-consensus-cell-types.Rmd
```diff
@@ -0,0 +1,188 @@
...
+celltype_df |>
+  dplyr::filter(
+    stringr::str_detect(ewing_annotation, "tumor"),
+    category == "immune"
+  ) |>
+  dplyr::count(library_id, ewing_annotation, consensus_annotation)
+```
+
+These results show macrophage and T cells, which are the two most prevalent categories immune cell categories.
+We might want to investigate these cells further, but at least we should probably exclude them as "normal" cells for `inferCNV` since their true status is bit unclear.
```
allyhawkins, line 170:
```suggestion
We might want to investigate these cells further, but at least we should probably exclude them as "normal" cells for `inferCNV` since their true status is a bit unclear.
```

allyhawkins (approved):
LGTM

Issue #1091, opened 2025-03-20
Add gene order file


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

Part of #1080

### Describe the goals of the changes to the analysis module.

In order to run `inferCNV`, we will need to add a gene order file. Similar to the `cell-type-wilms-tumor-06` module, we might consider making one of these for at the level of chromosomes, and one with more fine-grained version that has chromosome arms. But, let's start with just chromosome for now and add arms later if it seems necessary after the first go around.

We already have a script for this in the `cell-type-ewings` module: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/f8186b1fc63a596c3416011c1aca2287d28abade/analyses/cell-type-ewings/scripts/cnv-workflow/00-make-gene-order-file.R

So, here we can just port that script over, including a comment with this permalink as a source.

### What will your pull request contain?

A script to get chromosome info for `inferCNV`, and a line to call this script in the module bash script

### Will you require additional software beyond what is already in the analysis module?

No

### Will you require different computational resources beyond what the analysis module already uses?

No

Issue #1090, opened 2025-03-20
Add script to run inferCNV with a normal reference


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

Part of #1080 

### Describe the goals of the changes to the analysis module.

Following #1088, we are ready to start running `inferCNV`. This issue tracks adding a script to run `inferCNV` on a single library, and we can loop over this script in the module bash script. This script should include an argument to specify the path to the normal reference matrix.
(Edit) - this script should also allow for running in HMM mode with `i3`, which can help detect specific CNVs including gain, loss, and neutral without being as parameter rich as the default `i6` HMM.

### What will your pull request contain?

A script to run inferCNV on a library and save results to `results`.

### Will you require additional software beyond what is already in the analysis module?

More packages will get added to renv

### Will you require different computational resources beyond what the analysis module already uses?

No

PR #1093, opened 2025-03-24
Add script to create ewings normal references

### LINKED ISSUE #1088: Create ewings inferCNV normal references

Part of #1080 

Following #1084, we plan to proceed using two normal references:

- All immune cells
- Only macrophages and types of T cells, which represent the dominant cell types

From either normal reference, we should exclude the immune cells which were labeled as `tumor` by `cell-type-ewings`, since we are not confident of their status. This information is indicated in `exploratory-notebooks/01_ewings-consensus-cell-types.nb.html`. 

This issue tracks adding these two references. Each reference should be a gene x cell matrix of raw counts, with ensembl ids as row names and barcodes as column names. We might consider adding a prefix like `normal-` to the barcodes too for confident later bookkeeping. 

A script to create the two reference matrices. These will be saved to `results` as RDS files.

I will probably include a script for ewings specifically, and a `utils` script with a function to more generally create a reference. We can use this for future projects after ewings. 

Some R/bioc packages, so renv will be updated

No

_No response_

sjspielman:

Closes #1088 

This PR adds a script to create normal references for inferCNV for Ewings, based on our conclusions from the previous exploratory notebook added in #1084 

- I ended up using the merged SCE to construct the reference since this object makes it very simple to subset cells! 
- The references are exported as RDS files containing raw count matrices to `results/normal-references/SCPCP000015`.
- renv updated to include `SingleCellExperiment` and dependencies
- READMEs generally updated
- I added a step for this into `run-analysis.sh`    

Results, which includes the reference files, are available from `s3://researcher-654654257431-us-east-2/infercnv-consensus-cell-type/results/`

All in `renv`

N/A

Note, this is being opened as a Draft since it will conflict a bit with #1091, so I will mark this for review once #1091 is merged in and this branch/PR can be updated.

### DIFF
--- a/.github/workflows/run_infercnv-consensus-cell-type.yml
+++ b/.github/workflows/run_infercnv-consensus-cell-type.yml
@@ -54,6 +54,7 @@ jobs:
       - name: Download test data and results
         run: |
           ./download-results.py --module cell-type-ewings --test-data
+          ./download-results.py --module merge-sce --project SCPCP000015 --test-data
 
       - name: Run analysis module
         run: |
--- a/.github/workflows/run_infercnv-consensus-cell-type.yml
+++ b/.github/workflows/run_infercnv-consensus-cell-type.yml
@@ -61,12 +61,6 @@ jobs:
         with:
           working-directory: ${{ env.MODULE_PATH }}
 
-      #- name: Install aws-cli
-      #  run: |
-      #    curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
-      #    unzip awscliv2.zip
-      #    sudo ./aws/install
-
       - name: Download test data and results
         run: |
           ./download-results.py --module cell-type-ewings --test-data
--- a/analyses/infercnv-consensus-cell-type/.gitignore
+++ b/analyses/infercnv-consensus-cell-type/.gitignore
@@ -7,4 +7,5 @@
 !/scratch/.gitkeep
 
 # Ignore any large reference files
-references/*gene_order.txt
+references/Homo_sapiens.GRCh38.104.gene_order.txt
+references/normal-references/
--- a/analyses/infercnv-consensus-cell-type/README.md
+++ b/analyses/infercnv-consensus-cell-type/README.md
@@ -15,13 +15,16 @@ To run the analysis, use the following command:
 ## Input files
 
 * Results from the `cell-type-ewings` module
-* Processed SCEs for project `SCPCP000015`
+* Processed SCEs and the merged SCE for project `SCPCP000015`
 
 To obtain these files, run the following commands from the top-level of the repository after logging into your AWS account.
 
 ```sh
 # Download the cell-type-ewings results
 ./download-results.py --module cell-type-ewings
+
+# Download the merged SCE object
+./download-results.py --module merge-sce --project SCPCP000015
 
 # Download the processed SCE objects
 ./download-data.py --project SCPCP000015
--- a/analyses/infercnv-consensus-cell-type/references/README.md
+++ b/analyses/infercnv-consensus-cell-type/references/README.md
@@ -2,4 +2,7 @@ This directory contains reference information needed for analysis.
 
 * `cnv-validation.tsv`: Lists specific CNVs which can be used for validation for diagnoses explored in the module
 * `Homo_sapiens.GRCh38.104.gene_order.txt` gene order file for input to `inferCNV`
-  * This file is created with `../scripts/00-make-gene-order-file.R` when running the analysis pipeline, but it is ignored from the repository due to size
+  * This file is created with `../scripts/00-make-gene-order-file.R` when running the analysis pipeline, but it is ignored from the repository due to its large size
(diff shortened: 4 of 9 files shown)


sjspielman:
I had started to work on next steps for #1090 to add a script to run inferCNV, and in doing so, realized there's a different approach here that might be cleaner so I will hold off on next steps until we can discuss this....

- Currently, the script in this PR saves a _matrix_ of cells to include in the reference. The inferCNV (subsequent PR) script would would be to combine this matrix with a given individual library's counts matrix for input to inferCNV, making sure not to duplicate cells.
- Instead of that we could do something like this:
  - Rather than saving a reference matrix, we can save a table of reference cell labels (formatted `<library_id>-<barcode>`) 
  - Then, in the inferCNV script, we could actually read in the _merged object_ rather than an individual library, and then subset the merged counts matrix to just be the union of a given library & the reference cells for input to inferCNV. This means the inferCNV script would need to have, as input, the merged object and a library ID of interest, rather than a `_processed.rds` file.

@allyhawkins Let me know (after training!) if we like this other approach more, or what I can clarify about it!

jashapiro:
> I had started to work on next steps for #1090 to add a script to run inferCNV, and in doing so, realized there's a different approach here that might be cleaner so I will hold off on next steps until we can discuss this....
> 
> * Currently, the script in this PR saves a _matrix_ of cells to include in the reference. The inferCNV (subsequent PR) script would would be to combine this matrix with a given individual library's counts matrix for input to inferCNV, making sure not to duplicate cells.
> * Instead of that we could do something like this:
>   
>   * Rather than saving a reference matrix, we can save a table of reference cell labels (formatted `<library_id>-<barcode>`)
>   * Then, in the inferCNV script, we could actually read in the _merged object_ rather than an individual library, and then subset the merged counts matrix to just be the union of a given library & the reference cells for input to inferCNV. This means the inferCNV script would need to have, as input, the merged object and a library ID of interest, rather than a `_processed.rds` file.
> 
> @allyhawkins Let me know (after training!) if we like this other approach more, or what I can clarify about it!

If the ultimate plan is to incorporate this into the workflow, I would avoid using the merged objects, for memory and file copying overhead reasons.

allyhawkins:
I think my preference is to have the first step be used to create a reference of normal cells, which contains the barcode, library ID, and counts matrix. I think I would favor saving this as an SCE in an RDS file to make it easy to work with? But up to you on that. 

Then the next step would take as input the normal cell reference and a single library to run inferCNV. I don't want to work with the merged objects because it's not going to be very practical when we get to other projects. Also I think if we are going to run inferCNV on each library (which we want to do) then it's a waste of time and memory to read in the merged object.

sjspielman:
>  I think I would favor saving this as an SCE in an RDS file to make it easy to work with? But up to you on that.

Right now it's just a matrix (in this PR), which to me feels more straightforward to work than an SCE with since we have to pull out the matrix anyways for input to inferCNV? To prep the inferCNV object, we can just bind up the reference and library counts matrices (again, making sure to not duplicate cells!). But that's a pretty straightforward change to make if it turns out SCE is preferable!

allyhawkins:
> > I think I would favor saving this as an SCE in an RDS file to make it easy to work with? But up to you on that.
> 
> Right now it's just a matrix (in this PR), which to me feels more straightforward to work than an SCE with since we have to pull out the matrix anyways for input to inferCNV? To prep the inferCNV object, we can just bind up the reference and library counts matrices (again, making sure to not duplicate cells!). But that's a pretty straightforward change to make if it turns out SCE is preferable!

I think the matrix is fine, but the SCE would allow us to store other information like the cell type and library ID, which might be helpful when evaluating results? And make it easier to make additional references like one with just macrophages and T cells.

sjspielman:
> I think the matrix is fine, but the SCE would allow us to store other information like the cell type and library ID, which might be helpful when evaluating results? 

This is a good point! I'll note the library ids are implicitly stored as part of the cell labels since it's from the merged object, but I'm sold on the argument that cell type labels may be helpful later. Will update to SCEs.

sjspielman:
This is now ready for review! Again, the reference files are now SCEs (not matrices as indicated in the opening comment) and are available in S3.

allyhawkins (commented):
I think overall this looks good, but I wonder if we should actually save these normal references to `references/normal-references` instead of `results`? To me, these aren't really results and `references` would be more appropriate. Obviously we don't need to add it to the repo, so I would include that folder in the `.gitignore`, but it would allow you to describe all the reference files in the `references/README.md` file.

sjspielman:
> I think overall this looks good, but I wonder if we should actually save these normal references to references/normal-references instead of results? To me, these aren't really results and references would be more appropriate

This makes sense to me as well, will do & `.gitignore`

sjspielman:
RTD had a stochastic failure here and needs to be retriggered, but instead of pushing an empty commit or so I'll re-request review now in case there are other changes to make. Since last time:

- references are exported (but ignored) to `references/normal-references` and documented in `references/README`
- I simplified the R script a bit - rather than creating 2 SCE objects and NULL'ifying each, instead let's just make 1 SCE object, NULLify some slots, and _then_ subset the second. This doesn't need the helper function.

allyhawkins (approved):
LGTM

Issue #1094, opened 2025-03-27
Run inferCNV on ewings and explore results


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

Part of #1080 

### Describe the goals of the changes to the analysis module.

Following #1090, it's time to start actually running `inferCNV`. This issue tracks running `inferCNV` across ewings samples (except for the one low-quality library we don't want to include) with both the immune and immune-subset references, and then writing an exploratory notebook to explore the results and begin to determine whether we are recovering expected CNV patterns.

Following this exploration, we can determine whether additional approaches to running inferCNV as outlined in #1089 are worth pursuing.

### What will your pull request contain?

An exploratory notebook for the first round of ewings inferCNV results.

### Will you require additional software beyond what is already in the analysis module?

Probably more tidyverse packages, but we're still in `renv` land

### Will you require different computational resources beyond what the analysis module already uses?

No

sjspielman:
Following some discussion we have decided to do the following for this issue:
- Using inferCNV defaults, run across _all_ (except the low-quality one) libraries
- Add an exploratory notebook to assess the following:
  - Are we identifying CNVs that match up with our validation CNVs? https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/infercnv-consensus-cell-type/references/cnv-validation.tsv
  - Is there a relationship between inferred CNVs and putative tumor cells as determined by the `cell-type-ewings` module?

sjspielman:
We have run these and generated reports available in `s3://researcher-654654257431-us-east-2/infercnv-consensus-cell-type/ `

PR #1095, opened 2025-03-28
Add infercnv script

### LINKED ISSUE #1090: Add script to run inferCNV with a normal reference

sjspielman:

Closes #1090

This PR makes the following changes:

- I added a script to run inferCNV on a given SCE, by default using the `i3` HMM model
  - Note there's a TODO in there if we want to include other arguments about clustering for use maybe later?
  - This code was adapted from the infercnv script in the ewings module
  - I ended up classifying cells just as unknown & reference, but we could divvy up the reference too if we wanted to ensure cell types cluster together. This might be worth exploring?
- I added a step in the bash script to run inferCNV, as well as variables needed to do so. I run inferCNV with both references across all but the bad library
- I ended up also modifying the script to create the reference to make sure we have consensus cell types included in the reference files. This seemed like it might be helpful to have downstream. In a future data release once those cell types are in the sce objects, we won't have to add these into the object, but for now we do so I put it in an `if`.
  - I also added some more docs about reference files based on how I've been making/using them 

### Results

I have not run this yet over the libraries so there are no results in S3 yet; I am planning to run them as part of #1094. That said, I did locally run a library (real data, not test) through after I saw a weird result from a test library. The real library gave the same weird result, screenshot below... The values for `has_cnv_chr1` are all `unknown_sX` (presumably `s` for "subcluster`), but I've never seen such a column _not be numeric_ before... So, hm. Not sure what it means or what this bodes for data analysis? Thoughts?

Screenshot 2025-03-28 at 2 24 39 PM

### DIFF
--- a/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
+++ b/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
@@ -1,5 +1,17 @@
 #!/usr/bin/env Rscript
 # This script runs inferCNV on an SCE object using the provided normal reference
+#
+# Usage:
+#
+# Rscript 01_run-infercnv.R \
+#   --sce_file <input sce file> \
+#   --reference_file <input normal reference file> \
+#   --output_dir <output directory for infercnv results> \
+#
+# For information on additional arguments, run:
+#
+# Rscript 01_run-infercnv.R --help
+#
 
 project_root <- here::here()
 
@@ -20,10 +32,18 @@ option_list <- list(
     help = "Path to the normal reference SCE"
   ),
   make_option(
-    opt_str = "--annotation_file",
+    opt_str = c("--output_dir"),
     type = "character",
-    default = file.path(project_root, "scratch", "infercnv_annotation.txt"),
-    help = "Path to annotation file for inferCNV input"
+    help = "Folder to save final infercnv results"
+  ),
+  # TODO: we may want some clustering options here later:
+  # https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1089
+  # https://github.com/broadinstitute/infercnv/wiki/infercnv-tumor-subclusters#tumor-subclustering-by-leiden-clustering-preferred
+  make_option(
+    opt_str = c("--hmm_model"),
+    type = "character",
+    default = "i3",
+    help = "The HMM model to use with inferCNV, either 'i3' or 'i6'. Default is i3."
   ),
   make_option(
     opt_str = c("--gene_order_file"),
@@ -32,25 +52,17 @@ option_list <- list(
     help = "Path to gene order file as tab delimited .txt file with no column headers.
       Columns are: Ensembl gene id, chr, start, stop."
   ),
-  # TODO: we may want some clustering options here later:
-  # https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1089
-  # https://github.com/broadinstitute/infercnv/wiki/infercnv-tumor-subclusters#tumor-subclustering-by-leiden-clustering-preferred
   make_option(
-    opt_str = c("--output_dir"),
+    opt_str = "--annotation_file",
     type = "character",
-    help = "Folder to save final infercnv results"
+    default = file.path(project_root, "scratch", "infercnv_annotation.txt"),
+    help = "Path to annotation file for inferCNV input"
   ),
   make_option(
     opt_str = c("--scratch_dir"),
     type = "character",
     default = file.path(project_root, "scratch", "infercnv"),
     help = "path to scratch directory to store intermediate files from infercnv"
-  ),
-  make_option(
-    opt_str = c("--hmm_model"),
-    type = "character",
-    default = "i3",
-    help = "The HMM model to use with inferCNV, either 'i3' or 'i6'. Default is i3."
   ),
   make_option(
     opt_str = c("-t", "--threads"),
--- a/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
+++ b/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
@@ -232,8 +232,10 @@ if (!opts$skip_hmm) {
     readr::write_tsv(cnv_metadata_file)
 }
 
-# copy final object to output directory
-fs::file_copy(scratch_obj_file, output_obj_file, overwrite = TRUE)
+# add reference information the options slot and save to output directory
+dat <- readRDS(scratch_obj_file)
+dat@options$normal_reference <- ref_name
+readr::write_rds(dat, output_obj_file, overwrite = TRUE)
 
 # copy png file to output directory
 fs::file_copy(scratch_png, output_png, overwrite = TRUE)
--- a/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
+++ b/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
@@ -88,9 +88,8 @@ stopifnot(
   "sce_file does not exist" = file.exists(opts$sce_file),
   "reference_file does not exist" = file.exists(opts$reference_file),
   "gene_order_file does not exist" = file.exists(opts$gene_order_file),
-  "annotation_file was not specified" = !is.null(opts$annotation_file),
-  "output_dir was not specified" = !is.null(opts$output_dir),
-  "scratch_dir was not specified" = !is.null(opts$scratch_dir),
+  "annotation_file does not exist" = file.exists(opts$annotation_file),
+  "output_dir was not specified" = !is.null(opts$output_dir)
   "hmm_model not properly specified" = opts$hmm_model %in% c("i3", "i6")
 )
 
@@ -157,15 +156,9 @@ reference_group_name <- "reference"
 
 # Create and export annotation table for inferCNV
 data.frame(
-  cell_id = colnames(sce),
-  annotation = "unknown" # TODO: Should we make this a consensus annotation once that information is in the SCEs? Is this overkill?
-) |>
-  dplyr::bind_rows(
-    data.frame(
-      cell_id = colnames(ref_sce),
-      annotation = reference_group_name
-    )
+  cell_id = colnames(sce)
   ) |>
+  dplyr::mutate(annotations = dplyr::if_else(barcodes %in% normal_cells, "reference", "unknown")
   readr::write_tsv(opts$annotation_file, col_names = FALSE)
 
 
--- a/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
+++ b/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
@@ -45,8 +45,8 @@ option_list <- list(
   ),
   make_option(
     opt_str = c("--skip_hmm"),
-    action = "store_false",
-    default = TRUE,
+    action = "store_true",
+    default = FALSE,
     help = "Use this flag to turn off fitting the HMM."
   ),
   # clustering options. context:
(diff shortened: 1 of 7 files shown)


sjspielman:
Phew, passing CI! Now ready for review, but can be saved for a future sprint for review (after current sprint which ends 4/4).

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -0,0 +1,211 @@
...
+scratch_obj_file <- file.path(scratch_dir, "run.final.infercnv_obj")
+output_obj_file <- file.path(opts$output_dir, glue::glue("{library_id}_reference-{ref_name}_cnv-obj.rds"))
+
+# png file to save
+scratch_png <- file.path(scratch_dir, "infercnv.png")
+output_png <- file.path(opts$output_dir, glue::glue("{library_id}_reference-{ref_name}_infercnv.png"))
+
+
+# Prepare input data  ---------------------------------------------
+ref_sce <- readRDS(opts$reference_file)
```
allyhawkins, line 135: Can you read this in with the other SCE at the top of the script?

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -0,0 +1,211 @@
...
+data.frame(
+  cell_id = colnames(sce),
+  annotation = "unknown" # TODO: Should we make this a consensus annotation once that information is in the SCEs? Is this overkill?
+) |>
+  dplyr::bind_rows(
+    data.frame(
+      cell_id = colnames(ref_sce),
+      annotation = reference_group_name
+    )
+  ) |>
```
allyhawkins, line 168:
```suggestion
data.frame(
  cell_id = colnames(sce)
  ) |>
  dplyr::mutate(annotations = dplyr::if_else(barcodes %in% normal_cells, "reference", "unknown")
```
I think this is a little clearer to follow. Also I don't think we need to include the annotations since we want to consider all non-reference cells together for CNV inference.

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -0,0 +1,211 @@
...
+  ref_group_names = reference_group_name
+)
+
+
+# run infercnv
+infercnv_obj <- infercnv::run(
+  infercnv_obj,
+  cutoff = 0.1, # use 1 for smart-seq, 0.1 for 10x-genomics
+  out_dir = scratch_dir, # save all intermediate files to scratch dir
+  HMM = TRUE,
```
allyhawkins, line 189: It looks like in the Ewing script we set `denoise = TRUE` which we do want. Can you just confirm that is the default behavior of `infercnv::run`, otherwise we should specify that.

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -0,0 +1,211 @@
...
+    help = "Path to the normal reference SCE"
+  ),
+  make_option(
+    opt_str = c("--output_dir"),
+    type = "character",
+    help = "Folder to save final infercnv results"
+  ),
+  # TODO: we may want some clustering options here later:
+  # https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1089
+  # https://github.com/broadinstitute/infercnv/wiki/infercnv-tumor-subclusters#tumor-subclustering-by-leiden-clustering-preferred
```
allyhawkins, line 41: I think we want to be able to use the `leiden_function` and `leiden_resolution`. I watched the tutorial they have on their GitHub and they talk about the CNV prediction being dependent on the clusters and that clusters with mixed populations or small numbers of cells are going to have less accurate predictions. So I think we do want to be able to adjust that. 

One thought I had, which is potentially crazy is to first run without the HMM and use that information to pick the appropriate clustering parameters and then run a second time with the HMM. Alternatively, we could just use the clustering parameters we have chosen from the gene expression data, which we have for the Ewing data but do not have for all the other projects. So at minimum I would add in options to specify the parameters and I might even add in an option to run without the HMM.

allyhawkins (commented):
I took a look at this and have a few comments, the biggest one being to add in some clustering params which you had already suggested. 

I don't have a solution for what you are seeing regarding "Unknown" in the CNV column, but my guess is that it was unable to call anything with confidence for that sample and that chromosome. Do you see the same thing when you try a different sample?

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
sjspielman: can't hurt to be explicit either way

sjspielman:
Well, indeed, there is no bug with the output... It was me misreading a file and thinking rownames were a column 🥴 . So that's good!

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
sjspielman: > clusters with mixed populations or small numbers of cells are going to have less accurate predictions. So I think we do want to be able to adjust that.

What's good here though is we can at least assess this. We know from the output which cells are in which subcluster, so we can get a sense if some are less reliable. 

Noting that this is section of infercnv code which seems to set the resolution parameter. No idea where this calculation comes from..... https://github.com/broadinstitute/infercnv/blob/624feae9727dff74926aecd0d8945a21d61b572b/R/inferCNV_tumor_subclusters.R#L586-L593

> . Alternatively, we could just use the clustering parameters we have chosen from the gene expression data, which we have for the Ewing data but do not have for all the other projects

I had thought about this at one point as well, and want to note that in order to do this, we'd need to use the `leiden_method="PCA"` argument to ensure a (closer at least) apples-to-apples situation. Without this argument, clustering would be performed on the expression directly, but the parameters we have are definitely from clustering the PCA.

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -0,0 +1,211 @@
...
+
+# read SCE file so we can set up paths with its library id
+sce <- readRDS(opts$sce_file)
+
+# define reference name to set up output files and paths
+ref_name <- stringr::str_remove(
+  basename(opts$reference_file),
+  "^ref-"
+) |>
+  stringr::str_remove("\\.rds$")
```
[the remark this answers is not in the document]
sjspielman: I now use distinct output directories as requested, but I did keep this ref_name variable. I don't use it for file names, but I do add it into the infercnv outputs directly just so we are sure to keep track!

- I had it to the `options` slot of the infercnv object itself
- If we run an HMM, I add it as a tsv column

Let me know what you think!

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
allyhawkins: > I had thought about this at one point as well, and want to note that in order to do this, we'd need to use the leiden_method="PCA" argument to ensure a (closer at least) apples-to-apples situation. Without this argument, clustering would be performed on the expression directly, but the parameters we have are definitely from clustering the PCA.

I think if we are going to explore different clustering options here then this is something we will want to explore more and won't be able to decide in this PR. So with this information, I would update the script to add a few new arguments: `k_nn`, `leiden_method`, `leiden_function`, `leiden_resolution`, and `run_hmm` which is a flag to decide if `inferCNV` is run with or without the HMM. That way we have the option to use these parameters when we actually run the script and look at the results in an exploratory notebook(s).

allyhawkins: It looks like you actually already did that! So you can ignore me.

sjspielman: Don't have `k_nn` though, will add! Fyi was waiting for formally request review once this passes CI, but it does look now like it's going to this time :)

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -0,0 +1,241 @@
...
+    help = "The HMM model to use with inferCNV, either 'i3' or 'i6'. Default is i3."
+  ),
+  make_option(
+    opt_str = c("--skip_hmm"),
+    action = "store_true",
+    default = FALSE,
+    help = "Use this flag to turn off fitting the HMM."
+  ),
+  # clustering options. context:
+  # https://github.com/broadinstitute/infercnv/wiki/infercnv-tumor-subclusters#tumor-subclustering-by-leiden-clustering-preferred
```
allyhawkins, line 53: Just a reminder to add the `k_nn`

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -0,0 +1,241 @@
...
+# Parse options ----------------
+opts <- parse_args(OptionParser(option_list = option_list))
+
+
+# check input directories
+stopifnot(
+  "sce_file does not exist" = file.exists(opts$sce_file),
+  "reference_file does not exist" = file.exists(opts$reference_file),
+  "gene_order_file does not exist" = file.exists(opts$gene_order_file),
+  "annotation_file not provided" = !is.null(opts$annotation_file),
```
allyhawkins, line 115:
```suggestion
  "annotation_file not provided" = file.exists(opts$annotation_file),
```
Since we set a default for this, we want to just make sure it exists.

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -0,0 +1,241 @@
...
+
+# Run infercnv ------------------------
+
+# create the infercnv object
+infercnv_obj <- infercnv::CreateInfercnvObject(
+  raw_counts_matrix = raw_counts_matrix,
+  annotations_file = opts$annotation_file,
+  delim = "\t",
+  gene_order_file = opts$gene_order_file,
+  ref_group_names = reference_group_name
```
allyhawkins, line 198:
```suggestion
  ref_group_name = reference_group_name
```
I think this should probably be singular?

allyhawkins (commented):
This looks good, just need to add that `k_nn` argument!

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
sjspielman: This was a revision you previously left, but when I tested it I realized doesn't actually work as the code is written - we do have a default for it, but the file itself doesn't exist because it's written out as part of the script itself. But, what I can also check for is that its `dirname` exists?

sjspielman: noting i added that additional check here! https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1095/commits/d4be9c673c1389ab05baae47363e831d9269188b

allyhawkins: Ah that makes sense! I do think rather than check that the directory exists you could just make the directory to ensure that it exists before writing out.

allyhawkins (approved):
LGTM. I had one comment about the directory check but up to you on how you want to handle it.

PR #1102, opened 2025-04-15
Add template to explore ewings inferCNV results

### LINKED ISSUE #1100: Add template notebook to explore initial ewings infercnv results

Broken out of #1094

I have now run `inferCNV` over ewings samples, and we are ready to begin exploring the results. I think the most convenient way to look through all these results is to use a template notebook. This notebook should contain the following aspects/ask these questions for a given library and each of the two normal references (all immune cells & subset of immune cells):

- (Show the inferCNV pngs)
- What are the per cell CNV distributions for reference vs query cells? We expect a bimodal distribution here, with normal cells and some amount of query cells near zero, and a group of putative tumor cells with higher values
- What is the relationship between per cell CNV distributions and cell types? We'll consider both comparisons to consensus cell types and `cell-type-ewings` cell types
- What is the CNV landscape across chromosomes? How does this compare to the CNVs we have identified for validation?
- How do the clusters that inferCNV calculated look? We hope to see that these are reasonably sized and sane assignments.

A template notebook for exploring the initial inferCNV results, containing items that address the questions listed above.

Just renv stuff

No

1-2 days

sjspielman:

Closes #1100

The goal of this PR is to add a template Rmd that can be used to explore the inferCNV results generated as part of #1094. 

Note that I ended up running those with the `i3` HMM since it was really helpful with runtime. But, after starting to look at some results here compared to what we observed in the `cell-type-ewings` module (albeit with different reference cells!), I do think we might want to just got for `i6 since I am not necessarily seeing the expected CNVs we'd want to see...

That said, I have not formally run all libraries through the template, since let's get it reviewed first! :)

I added a new directory `notebook-templates` with this notebook in it. It contains visualizations of the following (it's a lot, but now we have a bunch to choose from for next steps!)

- the actual heatmaps
- distributions of CNVs per cell
- distributions of CNVs across cell types, considering both ewings & consensus
- UMAP colored by CNVs , using the ewings-colored UMAP as a guide
- A direct comparison of the number of CNVs inferred between the two references
- Some poking around at clusters: how big are they? any library batch effects? correspondance with cell types?
- A look at the actual CNV distributions themselves - here, I used the `proportion_scaled_` column because I borrowed a bit from the `cell-type-ewings` module, but I'm really open to discussion about the right column to use there?

I am very much aware that the code could be more modular since I am duplicating much of the plotting code when separate ggplot code is needed for the two references. We can absolutely firm these up if we want to, but I wanted to see what's going to make it through review first! Also, some can always move into a new utils file.  

Here is a copy of the template with `SCPCL000822`:
[ewings-infercnv-results.nb.html.zip](https://github.com/user-attachments/files/19763668/ewings-infercnv-results.nb.html.zip)

There are also these smaller changes:
- I updated the inferCNV script to use i6 as the default HMM, since it's their default. 
- I therefore updated the analysis script to specific i3, and added threads to be explicit as well

`s3://researcher-654654257431-us-east-2`

I have uploaded the inferCNV results into the bucket: `s3://researcher-654654257431-us-east-2/infercnv-consensus-cell-type/results/SCPCP000015/`. Again, these are all with `i3`. As part of review here, let me know if I should run with the `i6` HMM too or instead.

Haven't looked too in depth yet, only a little - that's the next step to be able to close out #1094!

Just renv

Nothing beyond what I've noted above.

N/A

### DIFF
--- a/analyses/infercnv-consensus-cell-type/template-notebooks/ewings-infercnv-results.Rmd
+++ b/analyses/infercnv-consensus-cell-type/template-notebooks/ewings-infercnv-results.Rmd
@@ -14,11 +14,20 @@ params:
 
 ## Introduction
 
-Here, we have run `inferCNV` with the following settings:
+This notebook explores `inferCNV` results run on library `params$library_id` from `SCPCP000015`.
+`inferCNV` was run with two different normal references:
 
-* With two references, `all-immune` and `subset-immune`
-  * The former contains all consensus immune cells, _excluding_ any which were also identified as tumor in the `cell-type-ewings` analysis module
-  * The latter is a subset of those immune cells considering only macrophage and T-cell cell types.
+* `all-immune`: All consensus immune cells across `SCPCP000015` libraries, _excluding_ any which were also identified as tumor in the `cell-type-ewings` analysis module
+* `subset-immnune`: A subset of `all-immune`, considering only macrophage and T-cell cell types
+
+`inferCNV` was run with the following settings:
+
+* We used the `i3` HMM which has improved run times relative to their `i6` HMM
+* We used the default `analysis_mode = "subclusters"`, which allows the HMM to call CNVs at the level of clusters rather than whole samples or individual cells
+* We used the following clustering parameters:
+  * Leiden clustering run on PCA (not the expression matrix)
+  * `k=20` nearest neighbors with a `resolution` parameter of 1 and using the `modularity` objective function 
+
 
 ## Setup
 
@@ -32,14 +41,14 @@ library(patchwork)
 # Set default ggplot theme, customized for UMAPs
 theme_set(theme_bw())
 
-
-#   theme(
-#     plot.margin = margin(rep(20, 4)),
-#     axis.text = element_blank(),
-#     axis.ticks = element_blank(),
-#     panel.grid = element_blank(),
-#     aspect.ratio = 1
-#   )
+umap_theme <- list(
+  coord_fixed(),
+  theme_classic(),
+  theme(
+    axis.ticks = element_blank(),
+    axis.text = element_blank()
+  )
+)
 ```
 
 ### Paths
@@ -63,13 +72,11 @@ data_dir <- file.path(repository_base, "data", "current")
 
 #### Input and output files
 
-Set paths to input and output directories and files in the chunk below.
-
 ```{r paths}
-# cnv validations
+# TSV with expected Ewing CNVs
 validation_tsv <- file.path(ref_dir, "cnv-validation.tsv")
 
-# references themselves so we know which cells belong to reference
+# References themselves so we know which cells belong to reference
 immune_ref_sce_file <- file.path(ref_dir, "normal-references", "SCPCP000015", "ref-all-immune.rds")
 subset_ref_sce_file <- file.path(ref_dir, "normal-references", "SCPCP000015", "ref-subset-immune.rds")
 
@@ -79,7 +86,7 @@ immune_png <- file.path(infercnv_immune_dir, glue::glue("{params$library_id}_inf
 subset_hmm_tsv <- file.path(infercnv_subset_dir, glue::glue("{params$library_id}_cnv-metadata.tsv"))
 subset_png <- file.path(infercnv_subset_dir, glue::glue("{params$library_id}_infercnv.png"))
 
-# cell types
+# ewings and consensus cell types
 ewings_tsv <- file.path(
   data_dir,
   "results",
@@ -98,17 +105,11 @@ sce_file <- file.path(
 )
 ```
 
-
-- we also need to read in the cnv-validations.tsv
-
-- section 1: what are the cnvs that they detected? do this from the hmm output at least
-- section 2: what is the relationship between consensus cell types and the cnvs? is there one?
-- section 3: what are the clusters? how big are they? what is their relationship to the cnv scores? do we want to show them on a UMAP?
-
 ### Read in data
 
 ```{r message = FALSE}
-validation_df <- readr::read_tsv(validation_tsv)
+validation_df <- readr::read_tsv(validation_tsv) |>
+  dplyr::filter(diagnosis == "Ewing sarcoma")
 
 celltype_df <- readr::read_tsv(ewings_tsv) |>
   dplyr::select(barcodes, consensus_annotation, ewing_annotation)
@@ -133,20 +134,8 @@ infercnv_df <- readr::read_tsv(immune_hmm_tsv) |>
     is_subset_reference = barcodes %in% subset_ref_barcodes
   )
 
-# Separate normal reference from query cells
-# cnv_query_df <- infercnv_df |>
-#  dplyr::filter(
-#    stringr::str_starts(cell_id, params$library_id)
-#  ) |>
-#  dplyr::mutate(
-#    barcodes = stringr::str_remove(cell_id, glue::glue("{params$library_id}-"))
-#  )
-# cnv_reference_df <- dplyr::anti_join(infercnv_df, cnv_query_df)
-
 # Read SCE and create data frame with full results
-sce <- readr::read_rds(sce_file)
-
-cnv_df <- sce |>
+cnv_df <- readr::read_rds(sce_file) |>
   scuttle::makePerCellDF(use.dimred = "UMAP") |>
   # replace UMAP.1 with UMAP1
   dplyr::rename_with(
@@ -155,18 +144,12 @@ cnv_df <- sce |>
   # add in hmm results
   dplyr::left_join(infercnv_df, by = "barcodes") |>
   # add in cell type results
-  dplyr::left_join(celltype_df)
-
-rm(sce)
-
-# remove MT and GL chromosomes
-chr_columns_to_keep <- which(
-  !stringr::str_detect(
-    names(cnv_df),
-    "chrMT|chrGL000219.1|chrGL000194.1"
+  dplyr::left_join(celltype_df) |>
+  # remove MT and GL chromosomes
+  dplyr::select(
+    -contains("chrMT"),
+    -contains("chrGL")
   )
-)
-cnv_df <- cnv_df[, chr_columns_to_keep]
 ```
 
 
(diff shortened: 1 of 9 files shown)


analyses/infercnv-consensus-cell-type/template-notebooks/ewings-infercnv-results.Rmd
```diff
@@ -0,0 +1,536 @@
...
+  dplyr::inner_join(
+    readr::read_tsv(ewings_tsv) |>
+      dplyr::select(barcodes, consensus_annotation, ewing_annotation)
+  )
+```
+
+
+## InferCNV Heatmaps
+
+Below, we show the `inferCNV` heatmaps from the `immune-all` and `immune-subset` normal references for this library.
```
allyhawkins, line 150: I actually think for this sample these heatmaps look pretty convincing in terms of seeing CNVs in the queried cells and not as much in the reference.

analyses/infercnv-consensus-cell-type/template-notebooks/ewings-infercnv-results.Rmd
```diff
@@ -0,0 +1,536 @@
...
+  labs(
+    title = "Immune-subset reference"
+  ) +
+  theme(legend.position = "bottom")
+
+wrap_plots(all_plot, subset_plot) +
+  plot_annotation(title = "Total CNVs per cell")
+```
+
+### CNVs across cell types
```
allyhawkins, line 237: For this plot I would lump tumor cells together. Alternatively, maybe we should make a category that's either reference, non-reference normal, and tumor and use that to facet these plots.

analyses/infercnv-consensus-cell-type/template-notebooks/ewings-infercnv-results.Rmd
```diff
@@ -0,0 +1,536 @@
...
+wrap_plots(consensus_violin, ewings_violin, ncol = 1, guides = "collect") +
+  plot_annotation(title = "CNV scores across cell types") &
+  theme(legend.position = "bottom")
+```
+
+
+
+## UMAPs
+
+Next, we'll visualize the total CNV counts as a UMAP.
```
allyhawkins, line 281: I think same thing here about showing tumor vs non-normal reference vs reference in the colors rather than the original cell type.

analyses/infercnv-consensus-cell-type/template-notebooks/ewings-infercnv-results.Rmd
```diff
@@ -0,0 +1,536 @@
...
+  geom_point(size = 0.1, alpha = 0.3) +
+  scale_color_viridis_c() +
+  facet_wrap(vars(normal_reference)) +
+  labs(title = "Total CNV per cell") +
+  umap_theme
+```
+
+## Compare references directly
+
+We can also look at CNV predictions directly between references using a point-density scatterplot, with the Spearman correlation shown:
```
allyhawkins, line 310: I don't know what I get from this plot? I do think we might want to compare the two references directly, but I'm not sure this is the way to go. I'm interested in what the CNV levels look like in tumor cells across references which you already have shown me above. I also want to know if we use the CNV profiles to classify tumor cells how accurate are we and which reference does a better job. That requires some figuring out how we actually use inferCNV to classify tumor cells and I think might be outside the scope of this notebook.

analyses/infercnv-consensus-cell-type/template-notebooks/ewings-infercnv-results.Rmd
```diff
@@ -0,0 +1,536 @@
...
+  scale_y_continuous(limits = c(0, max_axis), breaks = 1:max_axis) +
+  labs(
+    title = "Per-cell CNVs between normal references",
+    subtitle = glue::glue("rho = {rho}"),
+    color = "Number of cells"
+  ) +
+  theme(legend.position = "bottom")
+```
+
+## Clusters
```
allyhawkins, line 353: For all the plots where you have clusters as the x-axis can you order the clusters so the numbers are in the right order? right now s1, s10, and s11 are all in order.

analyses/infercnv-consensus-cell-type/template-notebooks/ewings-infercnv-results.Rmd
```diff
@@ -0,0 +1,536 @@
...
+wrap_plots(all_plot, subset_plot, guides = "collect") &
+  theme(legend.position = "bottom")
+```
+
+
+How many CNVs were inferred per cluster?
+We expect a single value here, because with `analysis_mode = "subclusters"` (the `inferCNV` default), CNV parameters are estimated per cluster.
+
+```{r}
+# How many CNV per cell per cluster?
```
allyhawkins, line 451: I think this plot is really helpful! It looks like `unknown` clusters dominated by tumor cells have higher total CNV than clusters dominated by normal cells, which is a good thing! I think we might be able to use this information to help classify tumor cell clusters.

analyses/infercnv-consensus-cell-type/template-notebooks/ewings-infercnv-results.Rmd
```diff
@@ -0,0 +1,536 @@
...
+
+```{r fig.height=8, fig.width=8}
+cnv_validation_df |>
+  dplyr::filter(normal_reference == "all-immune") |>
+  ggplot() +
+  aes(
+    x = proportion,
+    color = cnv_type
+  ) +
+  geom_density() +
```
allyhawkins, line 514:
```suggestion
  geom_density(bounds = c(0,1), bw = 0.05) +
```
Something is off with these plots since they aren't showing any gains that I would expect to be there based on the heatmap. This is what I used when plotting this value before so maybe this will help? 
Additionally, I might just look at the proportion scaled regardless of up/down right now and color by normal, non-normal reference, and tumor. I want to see if the tumor cells show higher CNV proportions in the chromosomes we expect to be altered. 

Also note that those CNVs are not present in every patient and most are in less than 50% so no guarantee that we'll see the ones listed in the table. This patient should have chr8 though and it looks like there is some slight amplification in the heatmap.

allyhawkins (commented):
Wow this is a lot!  
I went through the notebook and have some overall thoughts/comments I wanted to provide before I go through the code line by line. 

- I do think we should use the default of i6 rather than i3. It looks like i3 operates under the assumption that most cells are "normal" and outliers are CNVs. Since we already expect low signal, then I don't think we want to use that. Also I used the default i6 when I originally ran inferCNV on this library and I actually saw pretty strong signal, including in chr8. The only difference was that I used endothelial cells as the normal cells there, so that could have been a factor. Maybe we want to use those in our reference and not just immune? 
Here's a screenshot of the heatmap from this sample when run with endothelial cells as reference. I also used the default clustering params in that run too. 
Screenshot 2025-04-16 at 1 25 47 PM
- Can you make a column that lumps all tumor cells together regardless of High/low state? I think for most of the plots we want to use that since here we are mainly interested in whether or not the CNVs are higher in tumor vs normal. I think looking at the individual subtypes of tumor cells is going to make things more complicated at this point. I would probably show the high/low states in the bar charts looking at composition of each cluster, but for all other plots just lump all tumor together. 
- For the plots showing the distribution of cell types per cluster, can you make sure the colors match for the labels that are the same in both plots? I like these plots and think they are really helpful but would be easier to interpret with the colors! 
- I don't know that we need to look at everything with consensus cell types and with ewing annotations. I feel pretty good about the Ewing annotations and if a cell isn't labeled as tumor it's going to have the same label as the consensus annotation so these annotations are quite similar. Since there's a lot in this report, I think we could pick one annotation to compare too. 
- I think an important question that we want to answer is can we use the CNV profiles to distinguish tumor cells from other normal cells (like endothelial cells). It looks like endothelial cells tend to be in their own cluster which is good. I might plot the total CNV stratified by reference cells, non-reference normal cells, and then tumor cells.

analyses/infercnv-consensus-cell-type/template-notebooks/ewings-infercnv-results.Rmd
sjspielman: I can remove this plot! There's plenty else anyways 😬

allyhawkins:
> That said, I think we want a separate category/color for cells the consensus module called as normal but the ewings called as tumor. It might be nice to keep these visually separate in case we can tell they have something funky going on - and hey it could even be informative for nailing down their classification!

I think that makes sense! Let's keep in mind though that the numbers are going to be really small for that population and generally the tumor cells are always going to have the most cells. It might be worth adding the number of cells in each group to the plot labels for some of the plots.

sjspielman:
@allyhawkins I began responding to feedback here starting with cell types, and along the way ran the report through with the `i6` result for `SCPCS000490`, and the heatmap looks pretty similar to what I was getting before and less like yours with the endothelial reference. I'm attaching this (WIP, but still some updates!) HTML here:

Looking at yours and my UMAP, a major difference seems to be the number of clusters - I am getting way fewer clusters than you got with the default settings, so the heatmap looks a lot less noisy as a consequence. So, I'm wondering if we should try cluster parameters at their defaults here?

allyhawkins:
> Looking at yours and my UMAP, a major difference seems to be the number of clusters - I am getting way fewer clusters than you got with the default settings, so the heatmap looks a lot less noisy as a consequence. So, I'm wondering if we should try cluster parameters at their defaults here? Edit - or we'd really want to lower the resolution or similar to get more fine-grained clusters, perhaps.

I think starting with the defaults for the clustering is probably the right move. If we see the results are really bad then I think we can look at different clustering params.

sjspielman:
Attaching here a WIP report with default clustering (...so many clusters!), and I do think it's worth trying endothelial too for comparison. The heatmap and density plots do show an increased signal, sometimes even where we expect signal to be, but it's still muted compared to previous observations in `cell-type-ewings`. In addition, this plot suggests there is a subset of reference cells which are incorrectly classified and may actually be tumor based on CNV patterns, so the reference may not be serving us great here. 

Screenshot 2025-04-18 at 12 21 55 PM

[ewings-infercnv-results.nb.html.zip](https://github.com/user-attachments/files/19814282/ewings-infercnv-results.nb.html.zip)

allyhawkins:
> Attaching here a WIP report with default clustering (...so many clusters!)

I do remember there being a lot of clusters. Maybe that's a bad thing though and it's causing more noise and increasing the CNV scores. 

> In addition, this plot suggests there is a subset of reference cells which are incorrectly classified and may actually be tumor based on CNV patterns, so the reference may not be serving us great here.

I'm actually comforted by the fact that your density plot shows almost a trimodal distribution where we have reference, a middle group which is probably normal cells that aren't in the reference, and then tumor. Since we are not pulling out all normal cells as the reference, I would expect some overlap between CNV scores for reference and query cells. 

I do agree that it would be helpful to use endothelial cells as the reference and compare to using immune cells. That being said I want to caution you from making too many conclusions from this sample alone. This is the outlier sample in this project so I don't expect what's true here to always be true for the other samples.

allyhawkins:
I've been thinking about this and I think for this PR we want to focus on making a template notebook that can be used to evaluate results for each library across a single reference. Then we can easily render the notebook for each reference/ set of parameters we choose to test. This will make it easy to test different reference combinations and parameters and then look at all the results. From there, we can choose to directly compare any combinations that we think might be worth pursuing. 

So I might focus this PR on just looking at the results from a single run of inferCNV in a single library. The next step would be to create a script that runs inferCNV and renders the notebook using the set of parameters/ refs we are interested in testing. To start, I would just use two different refs - all immune and endothelial (since that's what we've had success with in the past). After that we can decide how to proceed and what might seem reasonable.

sjspielman:
> So I might focus this PR on just looking at the results from a single run of inferCNV in a single library. The next step would be to create a script that runs inferCNV and renders the notebook using the set of parameters/ refs we are interested in testing. To start, I would just use two different refs - all immune and endothelial (since that's what we've had success with in the past). After that we can decide how to proceed and what might seem reasonable.

This sounds like a good plan - I'm going to file issues and proceed in this direction accordingly, and probably open a new PR because of forks to file after I do #1105.

sjspielman:
Closing this in favor of updated PR where the template notebook will focus on a _single_ reference.

PR #1103, opened 2025-04-18
Update infercnv script, including clustering parameters

sjspielman:
This is a quick PR to update some aspects of the `inferCNV` script for running ewings with the normal references. Note that some of these changes are in #1102, but can't really stack from a fork so sending them in here instead, will merge into that open PR!

- I updated the HMM default to `i6` instead of `i3`
- We need to clear out the scratch directory between runs to avoid conflict previous results from using the other normal reference
- For any clustering parameters that updated the defaults for clustering parameters when running inferCNV, based on this discussion: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1102#issuecomment-2814045014
  - `leiden_function` default has been updated to CPM
  - `leiden_resolution` default has been updated to "auto" but if a number is provided the script will update this to a numeric

Using this updated script, I will run `inferCNV` through with our immune references and see how results are looking!

### DIFF
--- a/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
+++ b/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
@@ -40,8 +40,8 @@ option_list <- list(
   make_option(
     opt_str = c("--hmm_model"),
     type = "character",
-    default = "i3",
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
LGTM

Issue #1105, opened 2025-04-18
Add endothelial normal reference


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

Related to this comment: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1102#issuecomment-2815779360

### Describe the goals of the changes to the analysis module.

This issue tracks adding an endothelial reference for ewings samples in the `infer-consensus-cell-type` module. Preliminary results with immune cells as reference suggest that this population may not be identifying all CNV events, so we'd like to compare to endothelial which had some success in the `cell-type-ewings` module.

### What will your pull request contain?

I expect to update 2 scripts:

- The script that prepares ewing references
- The analysis run script to add a step to generate the endo reference

### Will you require additional software beyond what is already in the analysis module?

No

### Will you require different computational resources beyond what the analysis module already uses?

No

PR #1106, opened 2025-04-18
Update infercnv-consensus GHA

### LINKED ISSUE #1104: Update infercnv-consensus-cell-type module GHA

_No response_

Currently, the `infercnv-consensus-cell-type` module GHA runs the same workflow on PRs and on merge/when the "run all" workflow is run. Similar to what we do for some other long-running modules, we should make the process smaller when run during PRs.

I'm specifically thinking we should use the i3, not i6, model in testing when run on PRs.

An updated workflow file

N/A

N/A

_No response_

sjspielman:
Closes #1104 

This PR updates the `run` GHA for this module to move a little faster through PRs! Initially I planned to set this up to run `i3` on PRs and `i6` otherwise, but I decided when implementing that was maybe overkill and it seems fine to just have `i3` always run during _all_ testing, but run with `i6`  as default otherwise. So, I added a `TESTING` variable instead of a variable to control the HMM specifically.

Let me know if you think I should take the more fine-grained strategy where the HMM depends on the GHA context, rather than always running `i3` in GHAs.

### DIFF
--- a/.github/workflows/run_infercnv-consensus-cell-type.yml
+++ b/.github/workflows/run_infercnv-consensus-cell-type.yml
@@ -74,4 +74,4 @@ jobs:
       - name: Run analysis module
         run: |
           cd ${MODULE_PATH}
-          bash run-analysis.sh
+          TESTING=1 bash run-analysis.sh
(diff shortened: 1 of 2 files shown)


sjspielman:
Drafting this until the inferCNV code is updated more generally, as it was removed from the GHA in #1105.

PR #1123, opened 2025-05-09
Add ewings palette file

### LINKED ISSUE #1122: Add color scheme to ewings template
> We also may want to assign a color scheme so that when comparing across reports and references the same cell type is the same color.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1118#discussion_r2078367070_
            
--- 

This issue tracks adding a palette for coloring ewings cell types in the `infercnv-consensus-cell-type` module.

sjspielman:

Closes #1122 

This PR adds a palette file to `template_notebooks/palettes` (and documents in the README there) for ewings cell types. To determine which cell types to include, I identified the top 10 cell types across Ewings samples (collapsing all tumor into a single "tumor" category) and pooled together into a set of cell types we need for palettes. I assigned colors (mostly) from our unique colors from the consensus cell type validation palette.

Then, I updated the notebook to use this palette in the bar plot showing cell types across clusters, and simplified some of the plot code for setting factor order along the way; here is a screenshot of how the plot now looks for one sample, for example:

Screenshot 2025-05-09 at 12 46 28 PM

**In addition,** I updated the `inferCNV` script to help with memory since I was having some problems on our RStudio Server running it. In particular (h/t @jashapiro) I sparsified the input matrix, which really does help to drop the size of the interCNV object and that certainly doesn't _hurt_! I also did a bit of `rm()`/`gc()`.

### Results

N/A 

renv

Let me know if, now that we have this palette, it would be helpful to incorporate further? For example, we currently have a UMAP showing cell type categories of only tumor, reference, and other query cell. We could have a UMAP with these colors (but honestly I don't think it's necessary since we get a decent amount of insight from what we have!).

### DIFF
--- a/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
+++ b/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
@@ -193,7 +193,9 @@ sce <- sce[, !(colnames(sce) %in% duplicated_cells)]
 raw_counts_matrix <- cbind(
   as.matrix(counts(sce)),
   as.matrix(counts(ref_sce))
-)
+) |>
+  # sparse for memory
+  as("CsparseMatrix")
 
 # Check that we removed the duplicates:
 stopifnot(
@@ -220,6 +222,11 @@ infercnv_obj <- infercnv::CreateInfercnvObject(
   gene_order_file = opts$gene_order_file,
   ref_group_name = reference_group_name
 )
+
+# Clean up some memory before running inferCNV
+rm(sce, ref_sce, raw_counts_matrix)
+gc()
+
 
 # run infercnv
 infercnv_obj <- infercnv::run(
--- a/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
+++ b/analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
@@ -227,7 +227,6 @@ infercnv_obj <- infercnv::CreateInfercnvObject(
 rm(sce, ref_sce, raw_counts_matrix)
 gc()
 
-
 # run infercnv
 infercnv_obj <- infercnv::run(
   infercnv_obj,
--- a/analyses/infercnv-consensus-cell-type/template-notebooks/README.md
+++ b/analyses/infercnv-consensus-cell-type/template-notebooks/README.md
@@ -1,3 +1,4 @@
 This directory contains template notebooks used to analyze results from running `inferCNV` on an individual library.
 
 * `SCPCP000015_explore-infercnv-results.Rmd` explores `inferCNV` results run on a library from `SCPCP000015`
+* The `palettes/` directory contains TSV files with celltype palettes used in project-specific notebooks
--- a/analyses/infercnv-consensus-cell-type/template-notebooks/SCPCP000015_explore-infercnv-results.Rmd
+++ b/analyses/infercnv-consensus-cell-type/template-notebooks/SCPCP000015_explore-infercnv-results.Rmd
@@ -332,35 +332,20 @@ query_only_df <- library_has_cnv_df |>
       stringr::str_detect(ewing_annotation, "tumor"),
       "tumor",
       ewing_annotation
-    )
+    ),
+    # top 10 only - lump here to assist in color setup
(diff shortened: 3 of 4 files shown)


sjspielman:
CI has run one reference/library through successfully, so I'm going to request review before it's completed since it seems to be fine.

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -193,7 +193,9 @@ sce <- sce[, !(colnames(sce) %in% duplicated_cells)]
 raw_counts_matrix <- cbind(
   as.matrix(counts(sce)),
   as.matrix(counts(ref_sce))
```
allyhawkins, line 195:
```suggestion
  counts(sce),
  counts(ref_sce)
```
I actually don't think you need the `as.matrix` at all and then you don't need to add the lines later. When I was modifying this script for `ews-nf`, I was noticing large memory warnings so removed this and everything was fine.

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -193,7 +193,9 @@ sce <- sce[, !(colnames(sce) %in% duplicated_cells)]
...
-)
+) |>
+  # sparse for memory
+  as("CsparseMatrix")
```
allyhawkins, line 198:
```suggestion
) 
```

allyhawkins (approved):
Color palette looks good to me, but I have a suggestion about the infer CNV script that I think you should fix before merging this, but I don't need to see it again.

sjspielman:
Hm, I guess `cbind` doesn't like it, which maybe is why I had the `as.matrix()` there in the first place... But @allyhawkins it's working for you in `ews-nf`?

jashapiro:
> Hm, I guess `cbind` doesn't like it, which maybe is why I had the `as.matrix()` there in the first place... But @allyhawkins it's working for you in `ews-nf`?

You probably need `Matrix::cbind()` since you have not loaded `SingleCellExperiment` or any other library that knows sparse matrices.

sjspielman:
> since you have not loaded SingleCellExperiment or any other library that knows sparse matrices.

Nope, we're blind, the script _does_ load `SingleCellExperiment`, and `cbind()` didn't like the namespacing I did either, so something else is happening here. Looking more.

sjspielman:
So, turns out the matrices I have here are of different types (`DelayedArray` vs `dgCMatrix`) and that's causing the issue. Will solidify!

analyses/infercnv-consensus-cell-type/scripts/build-normal-reference/build-reference-SCPCP000015.R
```diff
@@ -114,6 +114,9 @@ clean_sce <- function(sce) {
   reducedDim(sce, "PCA") <- NULL
   reducedDim(sce, "UMAP") <- NULL
 
+  # ensure the counts matrix is sparse
+  counts(sce) <- as(counts(sce), "dgCMatrix")
```
allyhawkins, line 118: Are you using the merged objects? Maybe we should re-process the merged objects in the workflow results bucket so that they match the formatting of what's available on the Portal which should now be sparse.

jashapiro, line 118: Update to preferred syntax.
```suggestion
  counts(sce) <- as(counts(sce), "CsparseMatrix")
```

sjspielman: thanks!

jashapiro:
> > since you have not loaded SingleCellExperiment or any other library that knows sparse matrices.
> 
> Nope, we're blind, the script _does_ load `SingleCellExperiment`, and `cbind()` didn't like the namespacing I did either, so something else is happening here. Looking more.

Yep, I'm blind: But also I recommend not loading SingleCellExperiment until after parsing options, for much faster failing.

sjspielman:
> Yep, I'm blind: But also I recommend not loading SingleCellExperiment until after parsing options, for much faster failing.

Ah interesting, I hadn't thought to do that, but it makes sense

PR #1124, opened 2025-05-12
Template notebook fix

sjspielman:
While looking over the ewings inferCNV results, I noticed that there was a warning about many matches coming out of the some of the parsing when creating the data frame with `proportion_scaled` information. I wanted to confirm that this wasn't a bug, and it doesn't seem to be - we are joining a long with a wide data frame, so we do expect that many rows will match barcodes between data frames. So, I added a little `relationship = "many-to-many"` with an associated comment to clarify this. In addition, I realized I was filtering to remove cells outside the library, which we don't actually want because this will remove some of the reference cells! I removed that line too.


allyhawkins (approved):
LGTM

Issue #1125, opened 2025-05-12
Run inferCNV on libraries with "internal" normal references


sjspielman:
This issue is broken out of #1089.

We plan to pursue an modified version of this bullet from the parent issue as a control to assess the pooling strategy:

> We should consider comparing results with the pooled reference(s) to results from using a single library's immune cells, for any libraries with sufficient immune counts. This can help tell us if the pooling strategy itself is adding or detracting signal. 

At this point, we are also using endothelial cells for a reference, so I have retabulated which libraries can be used for this. The table below shows counts in each library for cells that fall into each category of immune or endo only, excluding any cells which are annotated as `"tumor"` by the ewings module:

```
# A tibble: 10 × 3
   library_id  group      n
   <chr>       <chr>  <int>
 1 SCPCL000822 endo     449
 2 SCPCL000822 immune   165
 3 SCPCL000824 endo      71
 4 SCPCL000824 immune    92
 5 SCPCL000825 endo      19
 6 SCPCL000825 immune     5
 7 SCPCL000826 immune     3
 8 SCPCL000828 immune    44
 9 SCPCL001112 endo     154
10 SCPCL001115 immune     5
```

We therefore have the following conditions we can run as control; we'd like to have close (ideally more than) 100 cells in the reference as well:

| library | reference cell type | number of cells in reference | 
|------|----------------------|----------------------------|
| `SCPCL000822` | endo | 449 |
| `SCPCL000822` | immune | 165 |
| `SCPCL000822` | endo_immune | 614 |
| `SCPCL000824` | endo  |    71 |
| `SCPCL000824` | immune  |    92 |
| `SCPCL000824` | endo_immune  |   163 |
| `SCPCL001112` | endo | 154 |

I don't expect running all seven through `inferCNV` to take more than about a day, so I don't think there is much of a drawback to running all these flavors through. To enable this, we'll need to add a separate R script to run with a normal reference from the same library, given a cell type category (here: immune, endo, combined) to use in the reference. This should be a general script which could be used with any ScPCA project in the future. We'll also need to add a step to run this script through a set of libraries of interest in the project-specific bash script.

sjspielman:
Noting the code to run this has been completed with #1126, but I will leave this open to leave comments next on insights from actually running inferCNV

sjspielman:
[...]

PR #1126, opened 2025-05-13
Add script to run inferCNV with an "internal" reference

sjspielman:

This adds the code needed for #1125, but I'll close that out after I actually run all the inferences!

This PR makes two main changes:
- I added an additional script (I am _not at all_ sold on the script name, any other ideas??) to run inferCNV but with a reference of cells within the same library, which I am designating "internal." This script is pretty similar to the existing inferCNV script but the way the reference is set up (as well as some input args accordingly!) are different. 
  - Now, we _could_ have only one inferCNV script with some `ifs` to handle the reference creation (pooled vs internal) and we'd have less repeated code, but this was a bit faster for me to write and given the goals of this module seems ok? Let me know if you prefer that I consolidate!
- I added steps to the project workflow bash script to run libraries & internal references as listed in #1125 

A couple implementation notes:
- I added a `testing` flag here which will ensure that there are enough reference cells created when using test data. Instead of basing reference creation on cell types when `testing` is TRUE, I use the first 20% of cells. This will keep things running in CI!
- Even though it's not strictly necessary for bookkeeping, I do keep the code to convert sce colnames to include the library id. This is because of how things are parsed in the template notebook that explores inferCNV results where it assumes there's a column `libraryid-barcode` to split up. Rather than adding conditions into the notebook for this parsing, I just made this consistent with what the notebook expects.
   

### Results

Nothing new is in my bucket yet since real data has not yet been processed (although I did kick off some runs!)

renv

N/A

N/A


analyses/infercnv-consensus-cell-type/scripts/01b_run-infercnv-internal-reference.R
```diff
@@ -0,0 +1,305 @@
+#!/usr/bin/env Rscript
```
allyhawkins, line 1: I think what I would do instead is modify the `run-infercnv.R` script that you already have to do the following: 

- Make the option for `reference_file` optional. 
- Add two new options for `reference_celltype_group` and `celltype_file` that are optional. 
- If no `reference_file` is provided, then you need to provide a `reference_celltype_group` and `celltype_file`.
- Handle creating the `raw_counts_matrix` and annotations file with an if statement depending on using a pooled or internal reference.

analyses/infercnv-consensus-cell-type/scripts/01b_run-infercnv-internal-reference.R
```diff
@@ -0,0 +1,305 @@
...
+  "blood vessel endothelial cell",
+  "microvascular endothelial cell"
+)
+if (opts$reference_celltype_group == "immune") {
+  reference_celltypes <- immune_celltypes
+} else if (opts$reference_celltype_group == "endo") {
+  reference_celltypes <- endo_celltypes
+} else {
+  reference_celltypes <- c(immune_celltypes, endo_celltypes)
+}
```
allyhawkins, line 184: Maybe we pull this out into its own little prep script that creates a tsv file with the assigned reference group and consensus annotation. Then that becomes an input to the build reference and run inferCNV script to match reference group names to consensus annotations?

allyhawkins (commented):
I see why doing it this way was originally faster, but I think it might be in our best interest to update the main `01_run-infercnv.R` script to accommodate running with either a pooled pre-built reference or an internal reference. It's really a lot of duplicated code... I commented with an overview of the approach I would take to modify the existing script. 

Along those same lines, we may want to pull out the specific reference group definitions to a separate setup script and keep the ref group and matching consensus cell types in a TSV file. I think these two changes should make this easier to use with more than just Ewing samples in the future. 

I do see why you chose to do this though, so if you really don't want to combine the scripts into one then at minimum you should update the names for both scripts so that we have `01a_run-pooled-infercnv.R` and `01b_run-internal-infercnv.R`, but again, I favor the approach of a combined script.

sjspielman:
I've gone back and forth on this, and I do think that we want two scripts named `01a` and `01b`, and maybe we can put a lot of the shared code in a `utils` file. I'll see if I can get that going!

analyses/infercnv-consensus-cell-type/scripts/01b_run-infercnv-internal-reference.R
sjspielman: Yes, I'll do that separately!

sjspielman:
> I've gone back and forth on this, and I do think that we want two scripts named 01a and 01b, and maybe we can put a lot of the shared code in a utils file. I'll see if I can get that going!

Noting while writing this up I changed my mind since code was still not as modular as I wanted, So, I'm now aiming for one script but with functions to handle data/reference prep for internal vs pooled to keep things modular. I'm also using a argument `reference_type` to control logic, with other arguments being used for additional aspects of reference information.

analyses/infercnv-consensus-cell-type/project-workflows/run-SCPCP000015.sh
```diff
@@ -106,3 +109,47 @@ for sample_id in $sample_ids; do
             output_file = '${html_name}')"
     done
 done
+
+# Next, we run a set of control inferences where the reference is taken from the same library itself
+# Libraries and references to run were determined here: https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1125
+for sample_id in SCPCS000490 SCPCS000492 SCPCS000750; do
```
allyhawkins, line 115: An alternative to this would be to add an `if` statement within the other `for` loop that checks if any of the sample IDs are one of these IDs. If that's the case then run the script with the internal reference and generate a new notebook.

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -108,22 +129,46 @@ option_list <- list(
...
   "hmm_model not properly specified" = opts$hmm_model %in% c("i3", "i6")
 )
 
 
+# check the reference information for both types of references, as different arguments are required for either type
+# also, define the ref_name while here
+stopifnot("reference_type must be one of 'pooled' or 'internal'." = opts$reference_type %in% c("pooled", "internal"))
+if (opts$reference_type == "pooled") {
+  stopifnot("pooled_reference_sce was not specified. This is required if reference_type is 'pooled'." = !is.null(opts$pooled_reference_sce))
+  stopifnot("pooled_reference_sce does not exist." = file.exists(opts$pooled_reference_sce))
```
allyhawkins, line 150: You could simplify this by setting the default to `""` and then just have the file exists check.

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -108,22 +129,46 @@ option_list <- list(
...
+  ref_name <- stringr::str_remove(
+    basename(opts$pooled_reference_sce),
+    "^ref_"
+  ) |>
+    stringr::str_remove("\\.rds$")
+} else {
+  stopifnot(
+    "internal_reference_group was not specified. This is required if reference_type is 'internal'." = !is.null(opts$internal_reference_group),
+    "celltype_tsv was not specified. This is required if reference_type is 'internal'." = !is.null(opts$celltype_tsv),
+    "reference_celltype_tsv was not specified. This is required if reference_type is 'internal'." = !is.null(opts$reference_celltype_tsv)
```
allyhawkins, line 161: Same thing with updating the defaults and simplifying these checks.

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -108,22 +129,46 @@ option_list <- list(
...
+# check the reference information for both types of references, as different arguments are required for either type
+# also, define the ref_name while here
+stopifnot("reference_type must be one of 'pooled' or 'internal'." = opts$reference_type %in% c("pooled", "internal"))
+if (opts$reference_type == "pooled") {
+  stopifnot("pooled_reference_sce was not specified. This is required if reference_type is 'pooled'." = !is.null(opts$pooled_reference_sce))
+  stopifnot("pooled_reference_sce does not exist." = file.exists(opts$pooled_reference_sce))
+
+  ref_name <- stringr::str_remove(
+    basename(opts$pooled_reference_sce),
+    "^ref_"
```
allyhawkins, line 154: Do we use this `ref_` anywhere else? Could we make the change to use `pooled` and `internal` as prefixes throughout?

analyses/infercnv-consensus-cell-type/scripts/utils.R
```diff
@@ -0,0 +1,72 @@
...
+
+#' Prepare and export an internal reference inferCNV annotation file
+#'
+#' @param reference_name Reference name to create, as recorded in the reference_celltype_tsv
+#' @param reference_celltype_tsv TSV with reference names and associated consensus cell types
+#' @param celltype_tsv TSV with per-cell cell types to consider when creating the reference
+#' @param annotation_file Path to export inferCNV annotations file
+#' @param library_id Library id for the SCE being processed
+#' @param testing Logical for whether we are running with test data
+prepare_internal_reference_annotations <- function(
```
allyhawkins, line 11: Since this is Ewing specific, maybe we note that in the function name? 
```suggestion
prepare_internal_reference_annotations_ewing <- function(
```

analyses/infercnv-consensus-cell-type/template-notebooks/SCPCP000015_explore-infercnv-results.Rmd
```diff
@@ -24,9 +24,15 @@ ref_text <- list(
   "ref_immune" = "immune",
   "ref_endo-immune" = "endothelial and immune"
 )
+if (stringr::str_detect(params$reference_name, "internal")) {
+  reference_type_text <- glue::glue("within the `SCPCL000822` library")
```
allyhawkins, line 28:
```suggestion
  reference_type_text <- glue::glue("within the {params$library_id} library")
```
I think this is what you mean?

sjspielman: sure is 😂

analyses/infercnv-consensus-cell-type/scripts/utils.R
sjspielman: We'd need a wrapper function to put this in to call from from the infercnv script itself, is why I figured to punt this to later! I was planning to explain when re-requesting review once CI finished but you beat my to the review :)

So, I'd prefer to either:
- keep this as is, or
- actually make a real wrapper function

do you have a preference?

allyhawkins: Ah I see what you're saying... I think I might punt this for later but then make a comment in the main inferCNV script where you call this function and file an issue about a future TODO.

analyses/infercnv-consensus-cell-type/project-workflows/run-SCPCP000015.sh
sjspielman: Huh, TIL there is no real quick way in bash to check if a value is in an array. So, while I don't disagree with this in principle, implementing it won't be as pretty as it sounds it seems.

allyhawkins: Huh? Why would you need to check if it's in an array? You could just do:

```
if [[ $sample_id == "SCPCS000490" || $sample_id == "SCPCS000492" || $sample_id == "SCPCS000750" ]]
```

sjspielman: Ah, I was only thinking of defining an array with those sample ids!! This is reasonable/small enough to use or's though!

analyses/infercnv-consensus-cell-type/scripts/utils.R
sjspielman: I left another TODO in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1126/commits/f37c1fb5b4353df2b24185d4312b5c401cc8f7f1, and I will open the issue after this goes in so that I can link the TODO spots we have to come back to

sjspielman:
CI still running but it has made it past some critical spots so I think we're good! Here are the changes I've made:

- I now ensure that the result directories are named with `_pooled` or `_internal` and killed the `ref_` prefix
- I made arg checks less ridiculous 
- I also updated the inferCNV script so that both reference types ("pooled" and "internal") now require the `reference_celltype_group` argument (e.g., "endo" or "immune"). This ends up getting used 1) for "internal" to create the reference, and 2) for both "internal" and "pooled" to create the reference name (e.g. "endo_pooled" or "endo_internal") to save into inferCNV output objects
  - Note: the reference name is also defined separately in the bash script, but I decided not to directly pass it in, and instead pass the reference cell type group and type in separately. In other words, I pass in separately "endo" and "pooled" rather than "endo_pooled", even though I could! The reason for this is I felt a clear argument with _just_ "pooled" or "internal" was important, and the "endo" on its own is important for filtering cell types when creating internal references. 
- Docs updates accordingly
- Plus, bonus update here to match what we're doing in `ews-nf` which will save us some `inferCNV` compute: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1126/commits/ab16e687dee159f1a07b428a4aefc72dd3767da6

analyses/infercnv-consensus-cell-type/scripts/01_run-infercnv.R
```diff
@@ -108,22 +134,31 @@ option_list <- list(
...
+# check the reference information for both types of references, as different arguments are required for either type
+# also, define the reference_name while here
+if (opts$reference_type == "pooled") {
+  stopifnot("pooled_reference_sce does not exist." = file.exists(opts$pooled_reference_sce))
+} else {
+  stopifnot(
+    "celltype_tsv does not exist." = file.exists(opts$celltype_tsv),
+    "reference_celltype_tsv does not exist." = file.exists(opts$reference_celltype_tsv)
+  )
+}
```
allyhawkins, line 159: Wow so much simpler!

analyses/infercnv-consensus-cell-type/scripts/README.md
```diff
@@ -1,13 +1,14 @@
...
-
+* `01_run-infercnv.R` runs `inferCNV` on a given ScPCA library using a specified normal reference
+  * The normal reference can either be "pooled" (created from combining cells across project libraries) or "internal" (created from cells in the given library only)
+* `utils.R` contains helper functions used by other scripts in this directory
 
 ## Normal references
 
-The directory `build-normal-reference/` contains scripts that prepare normal SCE references for use for `inferCNV` for individual projects and are named `build-reference-{project id}.R`.
-Normal reference files should be stored in `../references/normal-references/` and named as `ref_{reference name}.rds`.
+The directory `build-normal-references/` contains scripts that prepare pooled normal SCE references for use with `inferCNV` for individual projects and are named `build-reference-{project id}.R`.
```
allyhawkins, line 10:
```suggestion
The directory `build-normal-reference/` contains scripts that prepare pooled normal SCE references for use with `inferCNV` for individual projects and are named `build-reference-{project id}.R`.
```
Directory name is singular.

allyhawkins (approved):
Assuming this passes CI, LGTM

PR #1128, opened 2025-05-14
Add celltype reference file

### LINKED ISSUE #1127: Establish ewings normal reference TSV
> Maybe we pull this out into its own little prep script that creates a tsv file with the assigned reference group and consensus annotation. Then that becomes an input to the build reference and run inferCNV script to match reference group names to consensus annotations?

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1126#discussion_r2087566603_

---

This issue tracks writing a script to export a TSV file which contains cell types used for normal references. We can have two columns, `reference_name` and `consensus_celltype`, giving the celltypes that belong in each normal reference.

sjspielman:
Closes #1127 

As discussed in review for #1125, this PR adds a TSV file with normal reference cell types for SCPCP000015. I added the code to do this into the same script where we build the reference, and the TSV is exported to the same directory where the normal references are held. I also tossed in `pericyte` while I was here, can't hurt even though it won't actually do anything since there aren't any of those right now (but hey you never know what a future release brings!).

I also figured this is a reasonable file to keep under version control, so I modified the `.gitignore` so that these files are now present in the repo:
- `references/normal-references/README.md`: overall docs/description of normal references folder
- `references/normal-references/SCPCP000015/reference-celltypes.tsv`: the tsv established in this PR

### DIFF
--- a/analyses/infercnv-consensus-cell-type/.gitignore
+++ b/analyses/infercnv-consensus-cell-type/.gitignore
@@ -8,6 +8,5 @@
 
 # Ignore any large reference files
 references/Homo_sapiens.GRCh38.104.gene_order.txt
-references/normal-references/*
-!references/normal-references/README.md
+references/normal-references/SCPCP000015/*
 !references/normal-references/SCPCP000015/reference-celltypes.tsv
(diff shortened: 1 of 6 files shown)


allyhawkins (approved):
LGTM

PR #1131, opened 2025-05-15
Notebook text patch

sjspielman:
While running inferCNV for #1125, I caught a tiny bug in the notebook that we missed for text printing - 

For example, it was wrongly printing as (see the array!):
Screenshot 2025-05-15 at 11 22 02 AM

It's now fixed: 
Screenshot 2025-05-15 at 11 21 53 AM

I also did a quick organizational update to the workflow script to move some general definitions including the testing check up in the file to _before_ the section that performs the analysis.


allyhawkins (approved):
🚀

PR #1140, opened 2025-05-28
Add brief exploratory notebook for normal references

### LINKED ISSUE #1133: Investigate contents of pooled references further

Part of #1080

While reviewing results from running `inferCNV` on Ewings samples, we have observed that the pooled references sometimes show peaks of cells with higher CNVs that the internal references do not. For example, from the `SCPCS000490` sample:

Image

We should investigate whether there is any particular (problematic?) sample that might be contributing to this signal. Note that we also plan to ensure PDX's aren't included in the reference (#1132), and that may end up explaining this pattern entirely.

If the problem is not PDXs (#1132), there will be an exploratory notebook to try and nail down the discrepancies. 

No

No

This sprint

sjspielman:

Closes #1133

The aim of this notebook was to explore the CNV distributions for pooled reference cells to determine if we still see a bimodal distribution after re-running, and then to try and track down its source. After playing around with the data for a bit, I kind of convinced myself this signal is mostly gone now that we have debugged (removed PDX from reference, and updated the OpenScPCA-nf results). It's still somewhat there for the `immune` pooled reference but not to an extent I'm worried. So, this notebook imo is mostly for posterity? 

At one point, I did look around at cell types too but it was hard to plot all the colors and since not much signal was there in the first place, I ended up leaving out. But, I can add some cell type specific exploration if we think it will help.

Please let me know if you disagree and you think there are more trends worth digging into.

Here's the notebook, which really is just some ridge plots and a table! [02_ewings-reference-cnv.nb.html.zip](https://github.com/user-attachments/files/20491781/02_ewings-reference-cnv.nb.html.zip)

Results are in my bucket `researcher-654654257431-us-east-2` but relevant figures are reproduced in this notebook

inferCNV output

### Author checklists

_Check all those that apply._


analyses/infercnv-consensus-cell-type/exploratory-notebooks/02_ewings-reference-cnv.Rmd
```diff
@@ -0,0 +1,234 @@
...
+  dplyr::count(reference_library_id)
+```
+
+The table displayed below shows for each inference (`infercnv_library_id`), how many of top 25\% CNV cells in the pooled immune reference are from each library.
+For example, of the reference cells with the highest CNVs in the `SCPCL000822` inference, 28 cells came from `SCPCL000822`, 30 cells came from `SCPCL000824`, etc.
+Generally, these numbers are consistent across inferences although cells from `SCPCL000824` do seem to be more prevalent in certain samples.
+Even so, there doesn't seem to be any major trend here.
+
+```{r}
+top25 <- ref_immune_df |>
```
allyhawkins, line 207: Maybe rather than showing the total number of cells it would make sense to show the proportion of all cells from that library compared to cells that are in the top 25%? I would expect that generally this should be a small number and if we see any glaring numbers then that might be concerning. That being said, I think the signal is pretty low and we do see separation between tumor and normal for the most part.

allyhawkins (commented):
I think the level of detail here is good. I had one suggestion to improve the table. I also think having these summary plots across all libraries is really helpful and I think we're doing a pretty good job of creating a reference with any of these options (although I think endo might be the cleanest).

PR #1159, opened 2025-06-04
Add neuroblastoma CNVs to reference tsv

### LINKED ISSUE #1145: Add validation CNVs for neuroblastoma

Part of #1143 

This issue tracks adding neuroblastoma-specific CNVs we can use for validation into our [`cnv-validation.tsv`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/infercnv-consensus-cell-type/references/cnv-validation.tsv) file.

Updated TSV file and associated documentation

No

No

This sprint

sjspielman:

Closes #1145

This PR updates `references/cnv-validation.tsv` with a bunch of CNVs in neuroblastoma. I drew from two main papers for this, and it's worth noting that some of these are "more canonical" than others. We could consider adding another column to record that type of metadata, we could also pare this down to just the higher-frequency CNVs, or we could just record it all as I've done here. Let me know if you have a preference! 

I _think_ I like the idea of adding another column called `notes` or so where we can include info like `more common` or `less common`? Thoughts?


sjspielman:
> Maybe we could be general and put expected_frequency and use high and low? I'm open to other ideas though since I'm definitely not glued to that.

What about `common_cnv` and then `TRUE`/`FALSE`? I feel a bit uncertain about making frequency statements I think, and also I like the idea of T/F for consistent parsing.

sjspielman:
Changes include:
- I've updated sources here to reflect cnvs in multiple pubs, including adding the NBAtlas doi to the two really canonical ones
- I also caught 1 cnv that I missed from the oncoplot
- I updated the gain -> loss typo you caught
- I added the common column
  - For ewings, I only set chr8 to true
  - for NB, I set 1p, 11q, and 17q. This is more conservative than it could be, but I tend to be more conservative... If you think others should be TRUE here let me know! Notably also, 11q is only common in high-risk according to the primer paper, so that one could maybe be swapped to FALSE?

PR #1161, opened 2025-06-05
Exploratory notebook of NB normal reference options

### LINKED ISSUE #1146: Explore normal reference options for neuroblastoma

Part of #1143 

This issue tracks adding an exploratory notebook to look at the overall distribution of cell types in `SCPCP000004` (NB) samples to determine what normal references might be possible. Similar to the analysis we ended up doing for `SCPCP000015`, we'll want to look both at immune and endothelial cell types. We should also note whether any samples are lower quality and should be excluded from any pooled reference.

This notebook should explore these distributions both pooled across all samples in this project, and for individual samples. The former exploration will inform our choices for making a pooled reference, and the latter exploration will tell us which samples, if any, we can use with an internal reference as a control to test the effect of a pooled reference.

This PR will contain an exploratory notebook and associated documentation.

Not expecting anything new

No

This sprint

sjspielman:

Closes #1146

This PR adds an exploratory notebook to look at the distribution of immune and endothelial cells (from non-PDX samples!) in `SCPCP000004` to determine our options for normal references. 
In addition, I organized the `exploratory-notebooks` directory to have nested project-specific directories to reduce forthcoming clutter. 

### Results

Nothing in S3 for this PR, but the notebooks has "results"! There's a _ton_ of normal cells in this project, so we are definitely good to have an immune and endothelial reference. I don't think we need to consider a combined immune + endo given how many cells there are. We could use a subset of immune cells, but I think it will make the most sense to just start with all immune and see how it goes, especially since we might hypothesize that endothelial will be our preferred reference in the end anyways. The notebook also identifies 11 libraries we can try with an internal reference for comparison.

Here is the notebook: [01_nb-consensus-cell-types.nb.html.zip](https://github.com/user-attachments/files/20611385/01_nb-consensus-cell-types.nb.html.zip)


allyhawkins (commented):
This mostly looks good, but I had a few suggestions about the tables/plots. 

The other bigger comment I have is that we should definitely use a pooled reference with all normal non-cancer like cells. Based on the literature in NB and other tumors, the consensus seems to be either use all immune cells or all normal cells. So at minimum those are the references we should be using. I think the fact that we have enough endo cells is great and if that works as a reference that is awesome, but I think the Ewing tumors were the exception to the rule and not the standard. 

The other comment I have is about making a shared TSV with consensus cell types and groupings. This should include all normal consensus cell types and the assigned normal "group" (like immune, endothelial, neuronal, stromal). Then when working on each project we can pick the groups we want to use from that central TSV. That might be a separate PR that we should do before merging this in?

analyses/infercnv-consensus-cell-type/exploratory-notebooks/SCPCP000004/01_nb-consensus-cell-types.Rmd
```diff
@@ -0,0 +1,236 @@
...
+ggplot(category_prop_df) +
+  aes(x = library_id, y = prop_total_ref_cells, fill = category) +
+  geom_col(position = "dodge") +
+  theme(axis.text.x = element_text(angle = 30, hjust = 1, size = 8))
+```
+The plurality of endothelial cells are from `SCPCL000125`, and the plurality of immune cells are from both `SCPCL000126` and `SCPCL000127`.
+These libraries dominate their respective reference cell categories, but cells are still reasonably spread across the other libraries.
+
+
+Next, let's look more closely at the counts for cell types within these broad categories of `immune` and `endo`, since we may have enough data here to narrow the reference down further:
```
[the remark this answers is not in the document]
sjspielman: Yeah, I wasn't sure if this was worth doing while I was doing it. Now that I have you're review that we want to include _all normal cells_ in one of our reference options, this doesn't seem necessary to keep at all.

sjspielman:
This notebook has been roundly overhauled to incorporate our new reference cell group TSV and incorporate review comments including scope of which reference options to look at. I also excluded stem cells, neuronal cells, and stromal cells from any normal reference calculation, which leaves only normal cell types immune, endothelial, epithelial (1 library has 1 cell lol), and adipocyte (should this be excluded?) in an "all normal" reference. Let me know what you think now! Here's the notebook:  
[01_nb-consensus-cell-types.nb.html.zip](https://github.com/user-attachments/files/20631081/01_nb-consensus-cell-types.nb.html.zip)

sjspielman:
(Clicked wrong button, didn't mean to close!)

analyses/infercnv-consensus-cell-type/exploratory-notebooks/SCPCP000004/01_nb-consensus-cell-types.Rmd
```diff
@@ -0,0 +1,238 @@
...
+
+These cell groupings are present in the neuroblastoma samples we're considering:
+
+```{r}
+unique(celltype_df$reference_cell_group)
+```
+
+We should remove any cell groups which might correspond to tumor cells, including `stromal`, `neuronal`, and `stem cell`.
+The `NA` values here correspond to `Unknown` cells, which we'll also remove.
+This means we'll keep groups `endothelial`, `epithelial`, and `immune`.
```
allyhawkins, line 120:
```suggestion
This means we'll keep groups `endothelial`, `epithelial`, `adipocyte`, and `immune`.
```
I don't know that we have a specific reason to exclude `adipocyte` here, so I would keep them in the "all normal" reference.

analyses/infercnv-consensus-cell-type/exploratory-notebooks/SCPCP000004/01_nb-consensus-cell-types.Rmd
```diff
@@ -0,0 +1,238 @@
...
+  # arrange by total normal cells
+  dplyr::arrange(desc(all_normal))
+
+# this way we can print the full table at once; RMarkdown will only show 10 rows at once
+knitr::kable(celltype_count_wide_df)
+```
+
+Virtually all of the normal cell types are `endothelial` and `immune`, with very few `adipocyte` and literally only one `epithelial` cell.
+
+We'll look at this data in barplot form to see, per-library, the relative proportions of each normal cell type group out of all normal cells in the library.
```
allyhawkins, line 159: I think we want to include a color for "other" to show us what percentage of each library is made up of the cells that we are interested in.

analyses/infercnv-consensus-cell-type/exploratory-notebooks/SCPCP000004/01_nb-consensus-cell-types.Rmd
```diff
@@ -0,0 +1,238 @@
...
+  theme(
+    legend.position = "bottom",
+    axis.text.x = element_text(angle = 30, hjust = 1, size = 7)
+  )
+```
+
+Based on these numbers, we can consider three normal reference options:
+
+* All normal cells
+* ALl immune cells
```
allyhawkins, line 185:
```suggestion
* All immune cells
```

analyses/infercnv-consensus-cell-type/exploratory-notebooks/SCPCP000004/01_nb-consensus-cell-types.Rmd
```diff
@@ -0,0 +1,238 @@
...
+* All endothelial cells
+
+How many cells would be in each of those pooled references?
+
+```{r}
+# update the count data frame without categories we don't need
+celltype_count_df <- celltype_count_wide_df |>
+  dplyr::select(-adipocyte, -epithelial) |>
```
allyhawkins, line 193: To be clear, these are still included in the `all_normal` count but just not being printed out as rows in the below table?

allyhawkins (commented):
Thanks for making these updates! It's almost there, but I think the stacked bar plot needs one more tweak before it goes in. We should include all other cells in the library as one color so we can see how much of each library is from the reference cell types.

analyses/infercnv-consensus-cell-type/exploratory-notebooks/SCPCP000004/01_nb-consensus-cell-types.Rmd
sjspielman: Yes - they are still included in `all_normal`! But we don't need them as standalone categories. I'll update the comment to make this clear.

sjspielman:
I updated the barplot, but I didn't end up including an "other" color. Instead, the total height is the fraction of normal cells itself. So, if bar is 0.2 tall, than 20% of its cells are normal. Screenshot below.

I felt this was a bit easier to see the total normal fraction since it has a truncated Y-axis (which I add a note about in the text!) Let me know if that's ok with you or if you prefer to see this ranging from `[0,1]`.
Screenshot 2025-06-09 at 11 17 16 AM

allyhawkins (approved):
I'm good with the new plot! LGTM

PR #1162, opened 2025-06-05
Add reference cell group table

sjspielman:
As discussed in #1161, this PR adds a TSV of cell type groupings to support building normal references. In addition to some of the categories we've discussed on Slack, I also made these choices:

- A `stem cell` grouping which contains `somatic stem cell` and `stem cell`
- I put `melanocyte` in its own grouping
- I made an `epithelial` category

Let me know how these groupings look to you!


analyses/infercnv-consensus-cell-type/references/normal-references/README.md
```diff
@@ -5,3 +5,6 @@ The RDS files are not stored in version control due to their large size.
 Each project directory will also have a TSV tile listing all of the consensus cell types included in each reference, which is included in version control.
 
 All files are created by project-specific scripts found in `../../scripts/build-normal-reference/`.
+
+The TSV file `reference-cell-groups.tsv` contains consensus cell types organized into reference cell group categories, and it is used to help create references across projects.
```
allyhawkins, line 9: I think I would move this file to be in `infercnv-consensus-cell-type/references` alongside the CNV reference file and keep the `normal-references` folder for storing the actual references themselves? I don't feel strongly, it's just a lot of nested directories and I think we could decrease that.

analyses/infercnv-consensus-cell-type/references/normal-references/reference-cell-groups.tsv
```diff
@@ -0,0 +1,53 @@
...
+endothelial cell	endothelial
+blood vessel endothelial cell	endothelial
+microvascular endothelial cell	endothelial
+epithelial cell	epithelial
+ecto-epithelial cell	epithelial
+keratinocyte	epithelial
+squamous epithelial cell	epithelial
+epidermal cell	epithelial
+fibroblast	stromal
+hematopoietic precursor cell	HSPC
```
allyhawkins, line 19: I'm torn on if we should combine this with the other stem cell group, but I think it's probably fine since realistically anything that's a stem cell could be a cancer cell so I would never use those in a normal reference.

allyhawkins (commented):
The assignments look good, I just might move where the file is? Let me know if you disagree.

analyses/infercnv-consensus-cell-type/references/normal-references/reference-cell-groups.tsv
sjspielman: Yeah, that's a good point. While they are super closely related, for reference building in this context, we'd want to keep them separate.

allyhawkins (approved):
LGTM

sjspielman:
I'm going to merge this before checks complete since there are no code changes, only docs and a new TSV.

Issue #1166, opened 2025-06-09
Create SingleR trained models from NBAtlas


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

Part of #1151 

### Describe the goals of the changes to the analysis module.

We plan to use `SingleR` to annotate cell types in `SCPCP000004` samples using `NBAtlas` as a reference. This issue tracks adding a script to create a `SingleR`-trained reference to use for this step.

The [available `NBAtlas` data](https://data.mendeley.com/datasets/yhcf6787yp/1) contains only cell type metadata, not tumor labels or ploidy inferences, but we have requested this additional metadata in https://github.com/AlexsLemonade/OpenScPCA-admin/issues/362. If/when we are able to obtain this metadata, we may want to incorporate it into the cell type labels in the trained reference. For example, we might try two levels of annotation: one with cell type only, and one with combined cell type and tumor information. The specific details here can be fleshed out once we have that info. 

### What will your pull request contain?

A script to create the `SingleR` reference objects

### Will you require additional software beyond what is already in the analysis module?

I imagine this is when `SingleR` gets added to `renv`!

### Will you require different computational resources beyond what the analysis module already uses?

This should be fine to run on a laptop

sjspielman:
> We may want to do some clustering/pseudobulking before creating the SingleR reference

Related to #1167 too to the extent we want to try doing all this with the _full_ `NBAtlas` reference, but I again I do think it's most likely we just end up using the subsetted version only and use of the full reference might be a bit "recreational" as it were!

sjspielman:
I started looking into this, and we probably want to just pseudobulk ourselves rather than using `SingleR::aggregateReference()`. This function performs k-means clustering on the logcounts, and calculates a pseudobulk profile for each cluster by _averaging_ (not summing!) values, as a way to preserve some heterogeneity when collapsing the data.

But, we already have cooked-in heterogeneity in NBAtlas since data comes from several different studies in the first place. Also, I'm not necessarily enthusiastic about clustering such batchy logcounts (though it may just recapitulate the studies anyways, what with all the batchiness!). 

Therefore, I suggest we pseudobulk across studies & cell types directly with `scuttle::aggregateAcrossCells()`. We could also provide the argument `statistics = "mean"` to recapitulate the behavior of SingleR, but since this will operate on raw and not log counts, I think the default of summing is probably right.

jashapiro:
[...]

PR #1173, opened 2025-06-13
Reformat NBAtlas to SCE and AnnData objects

### LINKED ISSUE #1165: Add script to reformat NBAtlas object

Part of #1151 

The `NBAtlas` reference object is disseminated as a Seurat object. We will need to use this object in different formats: an SCE object for use with `SingleR`, and an AnnData object for use with `scArches/scANVI`. 

Since we'll need both these formats, we might as well tackle it in a single conversion script that reads in the Seurat object and export SCE and AnnData versions to scratch. We can then use this script to convert both the full and subsetted versions of `NBAtlas`; the specific one we'll use will depend on some runtime investigation (issue forthcoming), but we can still convert them to be prepared.

Script to convert the `NBAtlas` object from Seurat to both SCE and AnnData and associated docs.

Dependencies will get added to both renv and conda

This can be run on a laptop, but will need quite a few GB of memory. I will indicate the usage it ends up needing in docs as part of the PR.

_No response_

sjspielman:
_Noting I have edited this opening comment to reflect where this code has landed after updates along the road to de-drafting!_

Closes #1165

This PR adds a script to reformat the `NBAltas` object into both an SCE and AnnData object. 
I additionally added the module bash script to call this script after downloading the `NBAtlas` file. 

The script `scripts/00_convert-nbatlas.R` exports an SCE and AnnData object from a given `NBAtlas` reference file. As part of this, I also incorporate the `neuroendocrine-tumor` label for any cells in the atlas that also appear in the paper's "tumor zoom" subset. That information is now directly available from Mendeley so it's curl'd along with the atlas directly. The final exported objects contain all original cell metadata (including cell types, sample ids, and more).

Additional script notes:
- I ended up only retaining the raw counts matrix in these objects, since we won't need their normalized versions for either `SingleR` or `scANVI`, which both take raw counts matrices. So this saves us some space.
  - EDIT: We're keeping logcounts it turns out, see here: https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1166#issuecomment-3001346721 
- I additionally included a `testing` flag to make a smaller version of `NBAtlas` in testing scenarios that only contains 5% of the total cells in the given atlas, but retains the same proportions of cell types. This should help with efficiency of CI testing!

I additionally added the module bash script, which currently sets up paths, downloads the both the subsetted (which contains 50K cells for computational tractability) and full `NBAtlas` objects, and converts each of them to SCE+AnnData formats. 

renv

Noting besides what I noted above

There's a couple caveats I want to highlight for review:

- There are two versions of the full NBAtlas on available to download, where the newer version (v2) has 3198 fewer cells which were removed at some point during analysis for being putative doublets or otherwise low-quality
  - However, there is only one _subsetted_  version of the NBAtlas, and it was created from v1 which contained all those cells. There are 434 cells which are present in the subset which are _not present_ in full v2 atlas. Currently I'm not doing anything about these 434 cells, but I could remove them from the subsetted atlas if we really want things to match up. Or, I could go ahead and use the v1 full atlas which does contain all those cells.
  - Ultimately even though I do like the idea of removing those 434 cells from the subset, I really don't think it will end up mattering much, so it doesn't seem necessary? If I were to do it though, I'd approach it by adding another optional argument to the conversion script - a text file of cell ids to remove from any object being created. I'd create that text file separately in a standalone little helper script. With this approach,  we wouldn't have to read in two seurat atlas objects at the same time. 
- From correspondence with NBAtlas authors: they expect that the difference between the tumor-subset of neuroendocrine cells and other neuroendocrine cells may be more of a technical artifact rather than a specific biological signal. I'm going to open a separate issue to track some of that discussion/info shortly, along with some updated planning based on that corrspondence. Still, I do think it's worth creating those categories here, and we can always collapse them back later.
- Noting I have also opened this additional issue to deal with ID conversion, so this PR doesn't get too big: https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1180. Please feel free to leave any comments (or questions, not sure if I was super clear there) in that issue if you have thoughts about ensembl/symbols.


analyses/cell-type-neuroblastoma-04/scripts/00_convert-nbatlas.R
```diff
@@ -0,0 +1,114 @@
...
+    logcounts = nbatlas_seurat[["RNA"]]$data
+  )
+)
+
+# add in colData, including updated cell labels with `neuroendocrine-tumor` label for tumor cells
+colData(nbatlas_sce) <- nbatlas_seurat@meta.data |>
+  dplyr::mutate(
+    cell_id = rownames(colData(nbatlas_sce)),
+    NBAtlas_label = ifelse(
+      cell_id %in% tumor_cells,
```
allyhawkins, line 80: Are there any cases where tumor cells are not neuroendocrine? I think that they shouldn't be based on the methods, but we should double check if you haven't already.

analyses/cell-type-neuroblastoma-04/scripts/00_convert-nbatlas.R
```diff
@@ -0,0 +1,114 @@
...
+})
+set.seed(opts$seed)
+
+# read input files and determine tumor cell ids
+nbatlas_seurat <- readRDS(opts$seurat_file)
+tumor_cells <- readRDS(opts$tumor_metadata_file) |>
+  rownames()
+
+# convert Seurat to SCE object directly, to save space in the final object
+nbatlas_sce <- SingleCellExperiment(
```
allyhawkins, line 68: Is there any `rowData` in these objects? I think it's in `nbatlas_seurat[["RNA"]]@meta.data`? You should check that it doesn't have ensemble IDs in there since sometimes they hide them in that data frame.

allyhawkins (commented):
> There are two versions of the full NBAtlas on available to download, where the newer version (v2) has 3198 fewer cells which were removed at some point during analysis for being putative doublets or otherwise low-quality

I think we either want to use the unfiltered version for both or the filtered version for both. I don't like using the filtered version for one and not the other. Since we want this to be a good reference that we use with SingleR and scArches, probably to be more stringent on what we include and if there are cells that might be doublets that they identified, we should remove them. So I would use v2 and remove the cells that are in v1 but not v2 in the subset. This probably requires a separate setup script to generate a list of cell barcodes that are doublets and should be removed. 

As for ensemble IDs, I think you should do that conversion when setting up the objects before running any downstream analysis. But I agree that should be a separate PR.

sjspielman:
> So I would use v2 and remove the cells that are in v1 but not v2 in the subset. This probably requires a separate setup script to generate a list of cell barcodes that are doublets and should be removed.

will do!

> As for ensemble IDs, I think you should do that conversion when setting up the objects before running any downstream analysis. But I agree that should be a separate PR.

👍  - and to confirm, you're on board with making the atlas seurat vs making our objects symbols, right?

allyhawkins:
> 👍 - and to confirm, you're on board with making the atlas seurat vs making our objects symbols, right?

I think you mean making the atlas ensemble ids vs making our objects use symbols. If so, then yes!

analyses/cell-type-neuroblastoma-04/scripts/00_convert-nbatlas.R
sjspielman: No, I checked this before and just confirmed - there isn't any feature metadata at all :/

sjspielman:
This should be ready for another look! The main update is a helper script which identifies the cells which _can_ be included in reference, which comprises all ids in the new atlas version with droplets removed. Any cells that don't appear in this list now get removed before converting.

Issue #1171, opened 2025-06-12
Add list of marker genes to validate NB annotations


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

Part of #1151

### Describe the goals of the changes to the analysis module.

The goal of this PR is to add a list of marker genes to the `cell-type-neuroblastoma-04` module that can be used to validate cell type annotations.

(Edited, pressed enter too soon)

The overall cell types in `NBAtlas` are:

```
[1] "Neuroendocrine" "NK cell"        "T cell"         "Fibroblast"    
[5] "Myeloid"        "Schwann"        "B cell"         "Stromal other" 
[9] "Endothelial"    "pDC"            "RBCs"           "Plasma"  
```

For many of these, we can grab marker genes from [our `cell-type-consensus` validation markers](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/validation-markers.tsv).

At a minimum for tumor genes, we probably want `MYCN`, `ALK`, and [`AF1q`](https://www.nature.com/articles/s41388-024-02980-y). We'll also want to select marker genes here that are independent from marker identified in `NBAtlas` since this will be a more robust check. 

### What will your pull request contain?

TSV and associated docs with marker genes to use for cell type annotation validation.

### Will you require additional software beyond what is already in the analysis module?

No

### Will you require different computational resources beyond what the analysis module already uses?

No

sjspielman:
Following some discussion in DSTM, we do want to consider the marker genes that came out of `NBAtlas`, in particular tumor and tumor neuroendocrine, since this will be a good check that we are both identifying similar cell groupings.

PR #1181, opened 2025-06-25
Add marker genes for NB cell type validation

sjspielman:

Towards #1171 but not sure if this _closes_

The goal of this PR is to start getting marker genes into the repo for validating NB cell type annotation. 

To begin I am adding two TSVs with marker genes from either NBAtlas itself, or our consensus validation markers. It's worth noting the additional genes I indicated for inclusion in #1171 are all already part of the NBAtlas marker gene list in the first place, _except_ for `ALK` and `NTRK1` (which is also a common marker gene for lower risk NB). Looking more through the literature however, these genes seem more immediately relevant for a variant-based, not expression-based, analysis so I didn't consider them (yet).

To achieve this, I wrote a helper script which I stored in `scripts/setup`; it currently is lacking README details since it'll conflict with README changes in #1173, but I suspect that'll work itself out during review the course of review here... This script is not run in CI but has to be run directly, and I include docs for this in the script itself. It produces two TSVs which I've stored in the `references` directory that contain marker genes of potential interest. 

Some notes to be aware of:

- There are both up- and down-regulated genes in the NBAtlas, which is why I include the `direction` column throughout the TSVs
- The genes included in the NBAtlas tsv are from Supp Table 2, which contains cell-type marker genes. There is another spreadsheet which contains marker genes for the "tumor zoom" - aka marker genes for each subcluster of the set of tumor cells. In that data, there are 8 subclusters with on average ~700 marker genes per cluster. Because that's SO many genes, I'm not currently considering them. Do you think we should record those for any reason?

renv

When looking through the literature to find marker genes, this is the main other paper I found with good lists: https://www.nature.com/articles/ng.3899 - it has markers for both mesenchymal and adrenergic states, but determined from cell line data not patient data. And again, there are hundreds (see supp table 2). I didn't include any of these genes, but let me know if you think this might merit a TSV as well. 

I do think the genes lists I made are probably sufficient for now, and we can reconsider if validation comes back looking rough. Please let me know if you disagree with this as part of review!


analyses/cell-type-neuroblastoma-04/references/consensus-marker-genes.tsv
```diff
@@ -0,0 +1,101 @@
+NBAtlas_label	ensembl_gene_id	gene_symbol	direction	source
```
allyhawkins, line 1: To be honest I'm not super convinced that we need this and think we could just read in the TSV that already exists in the `cell-type-consensus` module when we grab marker genes? Is there another reason you think having this separate file will be beneficial?

allyhawkins (commented):
This mostly looks good, but I went back and forth on if we needed the consensus cell type TSV. I think I've decided we should keep it but with some modifications to keep the original annotation, the `NBAtlas_label`, and the `gene_observed_count`. 

I also think this the NBAtlas is a comprehensive list and good to start with.

analyses/cell-type-neuroblastoma-04/references/consensus-marker-genes.tsv
sjspielman: > Is there another reason you think having this separate file will be beneficial?

Mostly I like that the labels are recoded to match what's in NBAtlas to save some us wrangling this every time we'd have to use them

> Another related note is that you might want to keep only genes that have gene_observed_count == 1

👍

sjspielman:
This has now been updated as requested!

I wasn't sure if you _definitely_ wanted me to filter down to `gene_observed_count` since you also said to keep the column (and if we filter to all==1, then maybe no need to keep?). So for now no filtering, but keeping the column. Let me know if I should do more :)

analyses/cell-type-neuroblastoma-04/scripts/setup/prepare-gene-lists.R
```diff
@@ -0,0 +1,121 @@
...
+  "stromal cell" = "Stromal other", # may not be a great match, but can't hurt
+  "T cell" = "T cell"
+)
+
+validation_df <- readr::read_tsv(opts$consensus_marker_genes_url)
+
+consensus_markers_df <- validation_df |>
+  dplyr::filter(
+    validation_group_annotation %in% names(celltype_map),
+    gene_observed_count == 1
```
allyhawkins, line 104:
```suggestion
```
Sorry for the confusion, since we are keeping the column let's not filter.

allyhawkins (approved):
LGTM, just remove the filtering and we should be good!

sjspielman:
Noting the code changes here aren't run in CI, and the potentially CI-relevant changes to `renv.lock` passed at https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1181/commits/649e4c275460aa8bd0fb537cd5876a45db5a48ba. So this can be merged before the module check completes.

PR #1182, opened 2025-06-26
Convert NBAtlas to ensembl ids

### LINKED ISSUE #1180: Symbol/ensembl id conversion for NBAtlas

Part of #1151 

Since the `NBAtlas` data comes out of Seurat land, it doesn't have ensembl ids (including hiding out in the metadata), only symbols. We will need to reconcile this with our ScPCA data which uses ensembl ids primarily.

Since we plan to run both `SingleR` and `scArches` here - aka R and Python methods - we'll want to do this conversion in a way that can be consistent between platforms. Ideally then, this would be done at the initial NBAtlas conversion stage (#1165) before we get into running the methods themselves.

It looks like we can retain about half the genes from the atlas (about 20k down from about 40k), based on a quick comparison with genes we have in SCEs. We do have some functionality in `rOpenScPCA` to potentially help us out here in different ways:

- We can use `rOpenScPCA::sce_to_symbols()` in the script that actually runs `SingleR`
  - Caveats: This means the references stays coded with gene symbols. We would need to call this conversion when we actually run `SingleR`, but we'd also want to keep this consistent with scArches next. This means we'd need to export some kind of ID map to help out in python code. That map could be generated once from a single SCE at the beginning of the module workflow, using `rOpenScPCA::ensembl_to_symbols()` 
- We can convert the NBAtlas to ensembl ids as part of the initial conversion to SCE/AnnData objects. 
  - Caveats: We don't have a function in `rOpenScPCA` to wrap this, but we can of course just match ids up ourselves. 
  - Benefits: Get into ensembl land right off the bat!

I think the move is probably to convert to ensembl off the bat, _but_ we do also want to make sure that marker genes identified in the `NBAtlas` are still present when we do this. The procedure to convert would be:

- Directly convert gene symbols in NBAtlas using our ScPCA symbol mapping to the extent possible. 
  - There are 42,413 genes in NBAtlas. 
  - 23,998 are represented directly in our ScPCA objects (e.g. we can map them using the the `gene_symbol` rowData column)
  - I used the `rOpenScPCA::scpca_gene_reference` table to peek at other conversion possibilities too, since the NBAtlas data comes from a variety of originating studies
    - There are 4 genes which only appear in the older ensembl version in `gene_symbol_10x2024` which we can also toss in the mix:
Image
    - There are 7493 genes which only appear in the older ensembl version in `gene_symbol_10x2020`, some of which we can bring over too it the corresponding scpca symbol for that gene is NOT present in the NBAtlas already. 

(unless further discussion changes the plan), an updated version of the script being added for #1165 (PR #1173) that additionally maps over to ensembl IDs. This will necessarily make the NBAtlas smaller of course, but given the id differences that will happen no matter what. After this, we'll be able to live in ensembl land for the rest of the module.

I'll likely need to add `rOpenScPCA` to renv

No

_No response_

sjspielman:

Closes #1180

The goal of this PR is to use ensembl ids, not gene symbols, in our converted NBAtlas objects.

I added a section to the script `00b_convert-nbatlas.R` to convert ensembl ids using `rOpenScpCA::scpca_gene_reference` for assistance. This data frame holds ScPCA ensembl ids and symbols, as well as several other symbol versions - from the 10x reference used in 2020, and from the 10x reference used in 2024. 

I used the following logic to flip to ensembl:

- Use ScPCA mapping as a baseline, filling in ids from 10x2020 where ScPCA ids are unknown.
  - Based on initial exploration, there were over 7000 of these. Some of them are duplicates though, keep reading!
- As described in #1180, for the few ids which can only be found in 10x2024, toggle those in
- Remove the duplicates! 
  - After all this there are 128 duplicate ensembl ids that arose from circumstances like this: 
    - The ensembl id `ENSG00000235790` is associated with the scpca gene symbol `PEF1-AS1` and the 10x2020 gene symbol `AC114488.2`. _Both_ of those gene symbols are in NBAtlas. 
    - Ideally this would have previously been consolidated, but since it wasn't and there are relatively few of these genes, I took the conservative route and just removed them all. 
 - --> In the end we have 31370 (confirmed unique!) genes in the converted reference. This is down from an original 42413 genes in the raw NBAtlas.

Note that `renv.lock` was also updated since we now have the `rOpenScPCA` dependency to support the id mapping. I did a lil module README updating too.

renv, laptop

Mostly my logic - does it make sense to you? I feel like code is reasonably concise, do you? Basically, have a close look at...my code ;) 
In particular though, let me know if you think there are any choices I made here which might benefit from broader (group) discussion.

### DIFF
--- a/analyses/cell-type-neuroblastoma-04/README.md
+++ b/analyses/cell-type-neuroblastoma-04/README.md
@@ -8,7 +8,17 @@ TBD.
 
 ## Usage
 
-TBD.
+To run this module, use the `run-analysis.sh` script as follows:
+
+```
+bash run-analysis.sh
+```
+
+To run this module with test data, run the script as:
+
+```
+testing=1 bash run-analysis.sh
+```
 
 ## Input files
 
@@ -16,7 +26,6 @@ This module requires the following input files:
 
 * The processed `SCE` and `AnnData` objects for project `SCPCP000004`
 * The merged `SCE` object for `SCPCP000004`
-* Consensus cell types for `SCPCP000004`
 
 These files can be obtained with the following code run from the root of the `OpenScPCA-analysis` repository.
 You must be logged into your AWS account to download these files.
@@ -27,9 +36,6 @@ You must be logged into your AWS account to download these files.
 
 # Download merged SCE objects
 ./download-results.py --modules merge-sce --projects SCPCP000004
-
-# Download consensus cell types
-./download-results.py --modules cell-type-consensus --projects SCPCP000004
 ```
 
 
--- a/analyses/cell-type-neuroblastoma-04/renv.lock
+++ b/analyses/cell-type-neuroblastoma-04/renv.lock
@@ -173,6 +173,112 @@
       "Maintainer": "Marcel Ramos <marcel.ramos@sph.cuny.edu>",
       "Repository": "CRAN"
     },
+    "BiocNeighbors": {
+      "Package": "BiocNeighbors",
+      "Version": "1.22.0",
+      "Source": "Bioconductor",
+      "Date": "2022-12-17",
+      "Title": "Nearest Neighbor Detection for Bioconductor Packages",
+      "Authors@R": "c(person(\"Aaron\", \"Lun\", role=c(\"aut\", \"cre\", \"cph\"), email=\"infinite.monkeys.with.keyboards@gmail.com\"))",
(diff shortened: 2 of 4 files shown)


sjspielman:
Ok, my plan is to do a bit of exploration on some of our open questions here and come prepared to chat about it for next sprint!

Issue #1185, opened 2025-06-30
Add helper script to convert portal objects to use symbols


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

Part of #1151 

### Describe the goals of the changes to the analysis module.

Since `NBAtlas` uses gene symbols we plan to live in symbol land as well. To facilitate analyses, we will benefit from a simple helper script that converts SCE row names to gene symbols. While this could be done on the fly when running `SingleR`, we'll also need AnnData versions of these objects, and we'll be using an `rOpenScPCA` function to perform the conversion. By having this as a standalone script, we can do it all at once.

### What will your pull request contain?

Script to export a symbol-ified SCE and AnnData object from a given Ensembl-y SCE

### Will you require additional software beyond what is already in the analysis module?

No

### Will you require different computational resources beyond what the analysis module already uses?

No

### If known, when do you expect to file the pull request?

This sprint

PR #1187, opened 2025-07-01
Script to train SingleR model with NBAtlas

### LINKED ISSUE #1166: Create SingleR trained models from NBAtlas

### LINKED ISSUE #1184: Update nbatlas_label name

Part of #1151 

In this script, I named the label _that includes `-tumor_ `NBAtlas_label`. This is a bit misleading, since _we_ added that designation based on their "tumor zoom" subset.

Edit: forgot to paste "this script": https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/0a1595b42be2196e7712c66b682337d645f99471/analyses/cell-type-neuroblastoma-04/scripts/00b_convert-nbatlas.R#L92

To make things clearer, I plan to keep this column as the original labels, and instead add a logical column `in_tumor_zoom` and indicate `TRUE` for the relevant cells. 

This is a super quick code change of labeling, so I will probably do this simultaneously with building the SingleR model. It will involve a couple line changes to existing scripts.

No

No

This sprint

sjspielman:

Closes #1166 
Closes #1184 

This PR adds a script to aggregate NBAtlas and train a SingleR model.
While I was here, I also took care of some quick renaming as described in #1184.

Changes include:

- New script to train model
  - The script takes the atlas object and an SCE (to grab restrict genes) and trains the model with the setting `aggr.ref = TRUE`. I confirmed that this is equivalent (both in runtime and output) to a 2-step process where you first run `SingleR::aggregateReference()` and then train the model on that aggregated reference, so one step it is!
  - The script includes a flag that we might use in the future `--separate_tumor` to make a separate `Neuroendocrine-tumor` category
- Updates to `run-analysis.sh`
  - To save some compute time, I added an argument to this script `nbatlas_version` where we can specify "full" or "subset." Only the specified version is downloaded and used. In CI, I am now calling this script with the "subset" nbaltas only. 
  - I added the new script to train the model as the last step in the script
 - Docs updates to both the module README and the scripts/README
   - re-module README, we don't need AnnData anymore since we're making it ourselves, forthcoming in #1185   
 - SingleR added to renv 

renv; the new script takes a couple minutes but runs fine locally

N/A

N/A

### DIFF
--- a/analyses/cell-type-neuroblastoma-04/run-analysis.sh
+++ b/analyses/cell-type-neuroblastoma-04/run-analysis.sh
@@ -18,6 +18,8 @@ set -euo pipefail
 module_dir=$(dirname "${BASH_SOURCE[0]}")
 cd ${module_dir}
 
+# Define the nbatlas_version
+nbatlas_version=${nbatlas_version:-"full"}
 
 # Set up the testing flag
 testing=${testing:-0}
@@ -35,13 +37,8 @@ scratch_dir="scratch"
 mkdir -p $ref_dir
 mkdir -p $scratch_dir
 
-# Atlas file names
-nbatlas_sce="${ref_dir}/NBAtlas_${nbatlas_version}_sce.rds"
-nbatlas_anndata="${ref_dir}/NBAtlas_${nbatlas_version}_anndata.h5ad"
-
 # Define URL and file names for the specified NBAtlas
 # We ensure local files have the same name as on Mendeley: https://data.mendeley.com/datasets/yhcf6787yp/3
-nbatlas_version=${nbatlas_version:-"full"}
 if [[ $nbatlas_version == "full" ]]; then
     nbatlas_url="https://data.mendeley.com/public-files/datasets/yhcf6787yp/files/f5969395-5f6e-4c5d-a61a-5894773d0fee/file_downloaded"
     nbatlas_seurat="${scratch_dir}/seuratObj_NBAtlas_share_v20241203.rds"
@@ -52,6 +49,13 @@ fi
 nbatlas_tumor_url="https://data.mendeley.com/public-files/datasets/yhcf6787yp/files/78cad1b4-7425-4073-ba09-362ef73c9ab9/file_downloaded"
 nbatlas_tumor_metadata_file="${scratch_dir}/SeuratMeta_Share_TumorZoom_NBAtlas_v20250228.rds"
 
+# Reformatted atlas file names
+nbatlas_sce="${ref_dir}/NBAtlas_${nbatlas_version}_sce.rds"
+nbatlas_anndata="${ref_dir}/NBAtlas_${nbatlas_version}_anndata.h5ad"
+
+###################################################################
+######################## Prepare NBAtlas ##########################
+###################################################################
 
 # First, download the NBAtlas Seurat objects from Mendeley with a helper function
 # This function takes two arguments in order, the URL and the filename to save to
--- a/analyses/cell-type-neuroblastoma-04/run-analysis.sh
+++ b/analyses/cell-type-neuroblastoma-04/run-analysis.sh
@@ -28,6 +28,7 @@ else
 fi
 
 # Define and create directories
+data_dir="../../data/current/SCPCP000004"
 script_dir="scripts"
 ref_dir="references"
 scratch_dir="scratch"
@@ -83,11 +84,13 @@ Rscript ${script_dir}/00b_convert-nbatlas.R \
 
 
 ###################################################################
-############### SingleR annotation of NBAtlas data ################
+######################## SingleR annotation #######################
 ###################################################################
 
 # Train the SingleR model
+singler_model_file="${scratch_dir}/singler_model_NBAtlas.rds"
 Rscript ${script_dir}/01_train-singler-model.R \
     --nbatlas_sce "${nbatlas_sce}" \
-    --sce_file ../../data/test/SCPCP000004/SCPCS000101/SCPCL000118_processed.rds \
-    --singler_model_file scratch/singler_model_NBAtlas.rds
+    # arbitrary SCE; chosen because this is a very small one
+    --sce_file "${data_dir}/SCPCS000696/SCPCL001058_processed.rds" \
+    --singler_model_file "${singler_model_file}"
--- a/analyses/cell-type-neuroblastoma-04/run-analysis.sh
+++ b/analyses/cell-type-neuroblastoma-04/run-analysis.sh
@@ -6,6 +6,9 @@
 #
 # ./run-analysis.sh
 #
+# By default, this will run with the full NBAtlas dataset. To use the 50k subset instead, use:
+# nbatlas_version="subset" ./run-analysis.sh
+#
 # When running in CI or with test data, use:
 # testing=1 ./run-analysis.sh
 
@@ -14,6 +17,7 @@ set -euo pipefail
 # Ensure script is being run from its directory
 module_dir=$(dirname "${BASH_SOURCE[0]}")
 cd ${module_dir}
+
 
 # Set up the testing flag
 testing=${testing:-0}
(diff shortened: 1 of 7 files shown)


analyses/cell-type-neuroblastoma-04/run-analysis.sh
```diff
@@ -24,22 +30,40 @@ else
...
+
+# Define and create directories
+script_dir="scripts"
+ref_dir="references"
+scratch_dir="scratch"
+mkdir -p $ref_dir
+mkdir -p $scratch_dir
+
+nbatlas_sce="${ref_dir}/NBAtlas_${nbatlas_version}_sce.rds"
+nbatlas_anndata="${ref_dir}/NBAtlas_${nbatlas_version}_anndata.h5ad"
```
allyhawkins, line 62: So... if we aren't using the AnnData anymore, why don't we just get rid of this and any of the script that performs the conversion?

analyses/cell-type-neuroblastoma-04/run-analysis.sh
```diff
@@ -51,32 +75,58 @@ download_file() {
...
 
-nbatlas_subset_url="https://data.mendeley.com/public-files/datasets/yhcf6787yp/files/0a569381-3a0c-4eec-863a-e20544b686ed/file_downloaded"
+# Define URLs; the first 2 will always be downloaded
 nbatlas_full_url="https://data.mendeley.com/public-files/datasets/yhcf6787yp/files/f5969395-5f6e-4c5d-a61a-5894773d0fee/file_downloaded"
 nbatlas_tumor_url="https://data.mendeley.com/public-files/datasets/yhcf6787yp/files/78cad1b4-7425-4073-ba09-362ef73c9ab9/file_downloaded"
+nbatlas_subset_url="https://data.mendeley.com/public-files/datasets/yhcf6787yp/files/0a569381-3a0c-4eec-863a-e20544b686ed/file_downloaded"
+
+# Define NBAtlas reference object files with the same name as on Mendeley: https://data.mendeley.com/datasets/yhcf6787yp/3
+nbatlas_full_seurat="${scratch_dir}/seuratObj_NBAtlas_share_v20241203.rds" # this is needed regardless of which nbatlas_version was specified since we need to subset ids
+nbatlas_tumor_metadata_file="${scratch_dir}/SeuratMeta_Share_TumorZoom_NBAtlas_v20250228.rds"
```
allyhawkins, line 85: I think you could still define the paths to the full and subset together here and then just download the full or the subset depending on what version is being used.

analyses/cell-type-neuroblastoma-04/run-analysis.sh
```diff
@@ -51,32 +75,58 @@ download_file() {
...
+# Define which NBAtlas version we're actually converting, and download subset if it's that one
+if [[ $nbatlas_version == "full" ]]; then
+  nbatlas_seurat=${nbatlas_full_seurat}
+else
+  nbatlas_seurat="${scratch_dir}/SeuratObj_Share_50kSubset_NBAtlas_v20240130.rds"
+  download_file $nbatlas_subset_url $nbatlas_seurat
+fi
+
+# grab remaining downloads
 download_file $nbatlas_full_url $nbatlas_full_seurat
```
allyhawkins, line 96:
```suggestion
  download_file $nbatlas_full_url $nbatlas_seurat
else
  nbatlas_seurat="${scratch_dir}/SeuratObj_Share_50kSubset_NBAtlas_v20240130.rds"
  download_file $nbatlas_subset_url $nbatlas_seurat
fi

# grab remaining downloads
```
With this logic you are downloading the full regardless, but I think we want to download the full _or_ the subset and we never need both. Unless you change this to run both the full and subset by default and only the subset when testing, which I would also be fine with.

analyses/cell-type-neuroblastoma-04/scripts/00b_convert-nbatlas.R
```diff
@@ -77,51 +75,54 @@ all_cell_ids <- readr::read_lines(opts$cell_id_file)
...
-      Cell_type
-    )
-  ) |>
-  DataFrame(row.names = rownames(colData(nbatlas_sce)))
+  nbatlas_seurat <- subset(nbatlas_seurat, cells = keep_cells)
+}
+
+# convert to SCE, using `as` to avoid CI error
+# note that Seurat gives some deprecation warnings, but this works
+nbatlas_sce <- as.SingleCellExperiment(nbatlas_seurat)
```
allyhawkins, line 91: Did you check to make sure we don't need to do any renaming of assays or anything after you do this? The docs don't note what the format is after conversion... https://satijalab.org/seurat/reference/as.singlecellexperiment

analyses/cell-type-neuroblastoma-04/run-analysis.sh
allyhawkins: Wait we are using this AnnData since we will need it for scArches. I think I would keep it in and not remove converting it as you did later on.

analyses/cell-type-neuroblastoma-04/scripts/01_train-singler-model.R
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env Rscript
+#
+# This script trains a SingleR model from a given NBAtlas object
+# The script additionally requires an input SCE object to determine genes to restrict to (TODO: make this a TSV?)
```
allyhawkins, line 4: One idea is you could just pull our tx2gene file from `scpca-references` and read that in to get the gene id list to restrict too rather than reading in an SCE? But I think either one is fine.

analyses/cell-type-neuroblastoma-04/run-analysis.sh
allyhawkins: Ah I see in your note now that you need to have the full to get the IDs.

analyses/cell-type-neuroblastoma-04/run-analysis.sh
```diff
@@ -15,6 +18,9 @@ set -euo pipefail
 module_dir=$(dirname "${BASH_SOURCE[0]}")
 cd ${module_dir}
 
+# Define the nbatlas_version
```
allyhawkins, line 21: I think this is confusing me. Don't we want to generate models for both versions and then you are just specifying subset for running through CI? Why don't we just use the `testing` flag to specify only getting the subset version.

analyses/cell-type-neuroblastoma-04/run-analysis.sh
```diff
@@ -51,32 +75,58 @@ download_file() {
...
+   # For now, we will not save the AnnData object
+   #--anndata_file "${nbatlas_anndata}"
 
-Rscript ${script_dir}/00b_convert-nbatlas.R \
-    --nbatlas_file "${nbatlas_full_seurat}" \
-    --tumor_metadata_file "${nbatlas_tumor_metadata_file}" \
-    --cell_id_file "${cell_id_file}" \
-    --sce_file "${nbatlas_full_sce}" \
-    --anndata_file "${nbatlas_full_anndata}" \
-    ${test_flag}
```
allyhawkins, line 82: So I think I would update this to just do the conversion for the full if testing is false and then always do the subset conversion. By default, we want to generate the data for both full and subset, which is not the case currently.

analyses/cell-type-neuroblastoma-04/run-analysis.sh
```diff
@@ -51,32 +75,58 @@ download_file() {
...
+
+# Train the SingleR model with an aggregated reference
+singler_model_file="${scratch_dir}/singler-model_aggregated_nbatlas-${nbatlas_version}.rds"
+
+# can pass in an arbitrary SCE here for sce_file; this is just the first sample in the project
+Rscript ${script_dir}/01_train-singler-model.R \
+    --nbatlas_sce "${nbatlas_sce}" \
+    --sce_file "${data_dir}/SCPCS000101/SCPCL000118_processed.rds" \
+    --singler_model_file "${singler_model_file}" \
+    --aggregate_reference
```
allyhawkins, line 132: The same logic should be applied here where we always do the subset and if testing is false we do the full. I would maybe group the if statement and first process the subset including conversion and creating the model and then check the testing flag and if it's false convert and train the full model.

allyhawkins (commented):
I think the actual script for training the model is good. My main comments are with the logic around the script that is running the module. By default we should be running the steps needed to reproduce the full analysis that we are trying to do, which would be converting and training the models for both the full and subset references. When we are testing in CI, we can decrease this to only run the minimal steps, but I don't like that only one version is being run in the script so we have to run the main workflow twice to get the models we need for this module.

analyses/cell-type-neuroblastoma-04/run-analysis.sh
sjspielman: I can definitely revert this, but to clarify I would plan to turn it back on once we get to scArches! It just takes a decent amount of time in CI, and we don't need that AnnData version yet. I could also open an issue to remind us to revert it back once we get to that stage? Or, I'll revert it now if you really think I should. :)

analyses/cell-type-neuroblastoma-04/run-analysis.sh
sjspielman: @allyhawkins I do actually want to confirm you seen this reply here, since it means we can _get rid of_ the `00a_extract-nbatlas-ids.R` script and behavior. I think I should formally remove that script in a separate PR though, but in this PR I'll get it _only downloading EITHER_ the full or subset atlas. Does this sound ok?

allyhawkins: > @allyhawkins I do actually want to confirm you seen this reply here, since it means we can _get rid of_ the `00a_extract-nbatlas-ids.R` script and behavior. I think I should formally remove that script in a separate PR though, but in this PR I'll get it _only downloading EITHER_ the full or subset atlas. Does this sound ok?

Yes I think your statement about only using the full atlas for actual analysis moving forward makes sense. So I agree that we should update this to download either the full (default) or the subset (for testing) only. I think you could still just use the `testing` flag that you have instead of specifying a version though.

Issue #1168, opened 2025-06-09
Run SingleR with full and subsetted NBAtlas reference options


sjspielman:
_Edited based on project updates since first filing this issue_

Once we add a script to run `SingleR` on NB samples in #1167, we'll want to run this script on maybe 2-3 libraries of different sizes (there's a decent amount of library size variation in `SCPCP000004`) to determine compute needs before running over all samples.

In particular, we'll have two versions of the `NBAtlas` we could consider using: The full atlas, or a subsetted version which contains fewer cells, but still with a representative cell type distribution, that is more computationally tractable. In #1166, we'll use the `SingleR::aggregateReference()` function to make these more computationally tractable, but it's possible the full will still be larger. 

The goal of this issue is therefore to determine both the compute needs and annotation differences between running with either the full or subsetted (both aggregated) reference. We can compare annotations both by comparing to each other, and by comparing to the consensus cell types, using heatmaps colored by Jaccard similarity as we have previously done. Based on these comparisons, we'll hopefully be able to determine if it makes more sense to proceed.
Or, if there are substantial differences, we would want to revisit how we are creating the reference in the first place.

The PR associated with this issue will contain an exploratory notebook to compare SingleR results between references (unless runtime is completely prohibitive, in which case we'd have an answer!).

jashapiro:
> I've completed some run time checks for a couple libraries of which have a diverse number of cells and cell types, using 4 cores to _classify_ using SingleR on an model built from the _full_ NBAtlas in either aggregated or non-aggregated (yes, the full thing) flavors:

Did you change the `de.method` for singleR to "wilcox" for the non-aggregated runs?

sjspielman:
Closed by #1194 with the conclusion to use the full NBAtlas & perform aggregation.

PR #1192, opened 2025-07-02
Add script to run SingleR

### LINKED ISSUE #1167: Add script to run SingleR on NB samples

Part of #1151 

This issue tracks adding a script to run `SingleR` on a single library for `SCPCP000004` using the `NBAtlas` trained reference to be prepared in #1166. This script should read in the SCE of interest and the reference, and export a TSV file with annotations.

An R script

No

No

_No response_

sjspielman:

Closes #1167

This PR adds a script to run SingleR with `classifySingleR()` on a given SCE sample + trained model. 

I also updated the `run-analysis.sh` script to run this over all samples. I generally did this as follows:
- I modified the script so that the _non-aggregated_ reference is turned off in testing. So, CI will only train and classify with aggregated reference (see #1168 for a run time table!)
- I export results into `results/singler/<sample id>/<aggregated or not-aggregated>/libraryid_singler-annotations.rds`. I had run a couple samples through but accidentally overwrote them with test data versions 😬, so I'm re-running them now to be able to upload to the results bucket. I will mark this as ready for review once they are there!
  - Regenerating these results will also allow us to make some conclusions for #1168, which I think will likely entail _not running non-aggregated at all_. Again though, this PR has non-aggregated running when not testing for now.

I also updated the GHA to only run three samples through when testing in PRs. All samples will still be run upon merge or with the "run all module" workflow.

`researcher-654654257431-us-east-2`

Output RDS files from running SingleR

I'm going to be sending in an exploratory notebook looking at these next in #1168, so interpretation results will be coming in the next PR! Preview - based on the sample I worked with during exploratory notebook development, aggregated vs non-aggregated are real similar. But, stay tuned still!

renv; no changes there

N/A

N/A

### DIFF
--- a/analyses/cell-type-neuroblastoma-04/run-analysis.sh
+++ b/analyses/cell-type-neuroblastoma-04/run-analysis.sh
@@ -13,13 +13,13 @@
 # - `aggregate_singler` (Default value: 1)
 #   - Use `aggregate_singler=0` to turn off reference aggregation before SingleR model training
 #   - Example usage: aggregate_singler=0 ./run-analysis.sh
-# - `threads` (Default value: 4)
-#   - Use `threads=X` to specify a different number of threads
-#   - Example usage: threads=2 ./run-analysis.sh # to request 2 threads
 # - `sample_ids` (Default value: "all")
 #   - By default, all samples in SCPCP000004 are processed. Specify a set of ids to only run a subset, which
 #     may be helpful in certain exploratory runs
 #   - Example usage: sample_ids="SCPCS000101 SCPCS000102" ./run-analysis.sh # to run only `SCPCS000101` and `SCPCS000102`
+# - `threads` (Default value: 4)
+#   - Use `threads=X` to specify a different number of threads
+#   - Example usage: threads=2 ./run-analysis.sh # to request 2 threads
 
 set -euo pipefail
 
@@ -43,10 +43,13 @@ mkdir -p $singler_results_dir
 ######################## Set up variables #########################
 ###################################################################
 
+# Define argument defaults
+testing=${testing:-0} # default is not testing
+aggregate_singler=${aggregate_singler:-1} # default is to perform aggregation
+sample_ids=${sample_ids:-"all"} # default is to run all samples
 threads=${threads:-4} # default 4 threads
-aggregate_singler=${aggregate_singler:-1} # default perform aggregation
 
-# Set up the singler aggregation accordingly
+# Set up singler aggregation
 if [[ $aggregate_singler -eq 1 ]]; then
     aggregate_flag="--aggregate_reference"
     singler_aggregate_type="aggregated" # sub-directory where results will be saved
@@ -56,11 +59,10 @@ else
 fi
 singler_model_file="${scratch_dir}/singler-model_nbatlas_${singler_aggregate_type}.rds"
 
-# Set up the testing flag:
+# Set up the testing flag and data
 # - If we are testing, we'll use the NBAtlas 50K subset. Otherwise, we'll use the full atlas.
 # - We'll also name the NBAtlas reference object files here with the same name as on Mendeley:
 #   https://data.mendeley.com/datasets/yhcf6787yp/3
-testing=${testing:-0}
 if [[ $testing -eq 1 ]]; then
     test_flag="--testing"
     # subset atlas


sjspielman:
Noting RDS files for three samples I ran to test for #1168 are now in my bucket: `s3://researcher-654654257431-us-east-2/cell-type-neuroblastoma-04/results/`.

sjspielman:
@allyhawkins please also have a look at https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1194 while leaving your review here. Some exploratory notebook results from singler classification are there which suggest we don't need to be working with the non-aggregated, which may change how you review this PR!

analyses/cell-type-neuroblastoma-04/scripts/02_classify-singler.R
```diff
@@ -0,0 +1,78 @@
...
+  rOpenScPCA::sce_to_symbols(reference = "sce")
+
+# Perform annotation and export RDS
+singler_result <- SingleR::classifySingleR(
+  sce,
+  singler_model,
+  BPPARAM = bp_param
+)
+
+readr::write_rds(
```
allyhawkins, line 75: Do we think we might also want just the TSV file with the assigned labels? I know when doing exploratory analysis everything is in the object so it's easy to grab it from there, but if you are ever just working with the SingleR assignments and not the scores, I think the TSV file would be easier to deal with.

analyses/cell-type-neuroblastoma-04/run-analysis.sh
```diff
@@ -85,31 +88,69 @@ Rscript ${script_dir}/00b_convert-nbatlas.R \
...
     --singler_model_file "${singler_model_aggregated}" \
     --aggregate_reference
 
-# Non-aggregated reference
-Rscript ${script_dir}/01_train-singler-model.R \
-    --nbatlas_sce "${nbatlas_sce}" \
-    --sce_file "${data_dir}/SCPCS000101/SCPCL000118_processed.rds" \
-    --singler_model_file "${singler_model_not_aggregated}"
+# If we are _not_ testing, we'll also train a model with a non-aggregated reference
+if [[ $testing == 0 ]]; then
```
allyhawkins, line 110: I'm a little torn here because moving forward we aren't going to be using non-aggregated, but for exploratory analysis you are using both versions of the reference. I think we've talked in the past that exploratory analysis doesn't need to be fully accounted for in CI, but we should have notes on how to reproduce that analysis if we ever need to. So I think my suggestion would be to update this to add a flag for aggregating the reference to this script and by default aggregating is turned on. That would change how the logic is here so you would define the model file name and the aggregated parameter based on that flag and then train and run SingleR? 

Then in the exploratory notebooks README (in #1194) you can make a note that this was run both with and without aggregating prior to generating the notebook. What do you think of that?

allyhawkins (commented):
This looks good, but my main comment is just about how we plan to deal with the logic around running non-aggregated vs. aggregated. I definitely see why you might want to included it in CI and in the workflow, but we've also discussed in the past that running things that are only present in exploratory notebooks doesn't need to be totally accounted for in CI. Especially since you aren't actually running the non-aggregated in CI, then it might make sense to adjust the script to just train and run one model and have a flag to specify if it should be aggregated or not.

analyses/cell-type-neuroblastoma-04/run-analysis.sh
sjspielman: This all makes sense to me, on it!

Thinking ahead, I will want to test the NE/NE-tumor split as well, and I suspect we'll want to pursue the same strategy for that next step too.

analyses/cell-type-neuroblastoma-04/run-analysis.sh
```diff
@@ -18,16 +28,39 @@ module_dir=$(dirname "${BASH_SOURCE[0]}")
...
+# Set up the singler aggregation accordingly
+if [[ $aggregate_singler -eq 1 ]]; then
+    aggregate_flag="--aggregate_reference"
+    singler_model_file="${scratch_dir}/singler-model_nbatlas_aggregated.rds"
+    singler_aggregate_type="aggregated" # sub-directory where results will be saved
+else
+    aggregate_flag=""
+    singler_model_file="${scratch_dir}/singler-model_nbatlas_not-aggregated.rds"
+    singler_aggregate_type="not-aggregated"
+fi
```
allyhawkins, line 58:
```suggestion
if [[ $aggregate_singler -eq 1 ]]; then
    aggregate_flag="--aggregate_reference"
    singler_aggregate_type="aggregated" # sub-directory where results will be saved
else
    aggregate_flag=""
    singler_aggregate_type="not-aggregated"
fi
singler_model_file="${scratch_dir}/singler-model_nbatlas_${singler_aggregate_type}.rds"
```
You can pull this out and use the aggregate type variable to define the file.

analyses/cell-type-neuroblastoma-04/run-analysis.sh
```diff
@@ -41,13 +74,29 @@ else
     nbatlas_seurat="${scratch_dir}/seuratObj_NBAtlas_share_v20241203.rds"
 fi
 
-nbatlas_sce="${ref_dir}/NBAtlas_sce.rds"
-nbatlas_anndata="${ref_dir}/NBAtlas_anndata.h5ad"
+# Set up `sample_ids` and check that all sample_ids exist
+sample_ids=${sample_ids:-"all"}
```
allyhawkins, line 78: I would move this up where you define all the defaults.

allyhawkins (approved):
This looks good, just two small things to clean up.

PR #1205, opened 2025-07-07
NB module: Docker and remove cell id script

### LINKED ISSUE #1189: Set up Docker for cell-type-neuroblastoma-04

Part of #1151 

It's currently taking about 15-20 minutes to set up the environment in CI for `cell-type-neuroblastoma-04`. Following #1187, we'll have the bulk of what we really need in the `renv` environment here, so we should consider moving over to Docker sooner rather than later for the sake of CI efficiency.

Updates that get Docker up and running

No

No

This sprint

### LINKED ISSUE #1191: Remove the 00a_extract-nbatlas-ids.R script
> @allyhawkins I do actually want to confirm you seen this reply here, since it means we can _get rid of_ the `00a_extract-nbatlas-ids.R` script and behavior. I think I should formally remove that script in a separate PR though, but in this PR I'll get it _only downloading EITHER_ the full or subset atlas. Does this sound ok?

Yes I think your statement about only using the full atlas for actual analysis moving forward makes sense. So I agree that we should update this to download either the full (default) or the subset (for testing) only. I think you could still just use the `testing` flag that you have instead of specifying a version though.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1187#discussion_r2180346728_

---

This comment comes from some discussion in the linked PR. Since we are going to move ahead with the full atlas in analysis and subset atlas in testing, we don't need to be taking the time to subset IDs. 

This issue tracks removing this script and its associated behavior.

sjspielman:
Closes #1189
Closes #1191 

This PR is a two-fer of module logistics:

- For #1189, I updated the Dockerfile and turned on the GHAs to get this going. Note that I do expect to get a conda environment in this Dockerfile eventually, but I'm still starting with the bioconductor image since we know we can get conda installed into there from experience with other modules. This _may_ need to change when GPUs happen, but I'm not really prepared to predict those changes quite yet so this seemed ok for now
- For #1191, I got rid of the `00a` script and renamed our current `00b` back to `00`. I also removed code in the (now-named) `00_convert_nbatlas.R` script that used the cell id file which we've booted.

In the next PR I file for this module, I'll change the GHA to use the Docker image. Note that I did test this locally by building the docker image and running the analysis script in the container, with test data.


allyhawkins (approved):
LGTM

PR #1206, opened 2025-07-09
Perform exploration of splitting NE cells

### LINKED ISSUE #1204: Explore whether Neuroendocrine cells should be split into two categories

Part of #1151 

In #1194, we have concluded to proceed with SingleR using an aggregated reference built from the full NBAtlas. The next item to explore is whether cells in the reference labeled `Neuroendocrine` (NE) should be split into two categories or not where one "subcategory" consists of the NE cells included in the dedicated "tumor zoom." The script to train a SingleR model includes an argument with associated functionality to facilitate this:
https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/feeb56198d6aae5c26b8448cbeb12cb70ddd9ffc/analyses/cell-type-neuroblastoma-04/scripts/01_train-singler-model.R#L36-L40

I will perform this exploration as follows:

- Add a boolean flag to `run-analysis.sh` to handle whether this group should be split or not
  - Note that we may need to probably update output directories/file names somehow here too; this may mean we also have to tweak input file paths in the aggregated/non-aggregated exploratory notebook
- Add a script to `exploratory-analysis` which can run a subset of samples through the pipeline. This can be the same samples as those used in #1194 (or a subset of those even; e.g. one where we saw lots of NE-annotated cells and one where we did not).
- Add an exploratory notebook to compare results with an emphasis on checking whether separating out the tumor zoom cells is "oversplitting". We'll include heatmaps & UMAPs at least to get at:
  - In the split reference, is the 2nd highest score for NE cells always NE-tumor, and vice versa? This would indicate oversplitting. For this, wrangle the scores & delta.next results.
  - In the "combined" reference do all NE cells get annotated as NE and NE-tumor in the split reference? For this, make a heatmap.
  - Are there _other_ differences in annotations, or just NE differences? E.g., do the same cells get annotated as T cells both times? For this, make a heatmap.
  - We can use the [`SingleR::plotMarkerHeatmap` function](https://bioconductor.org/books/release/SingleRBook/annotation-diagnostics.html#based-on-marker-gene-expression) to look at the marker genes SingleR picked out. One related question is, are the NE marker genes in the combined reference encompassed by 2 NE categories in the split reference?
  - Is the distribution of NE-specific marker genes different between NE and NE-tumor? There's quite a few marker genes we have and we don't want to plot them all most likely, so this may involve a bit of trial and error. This can be a violin plot or so, likely, but we may not need it.

- Updated `run-analysis.sh` with flag to control this behavior
- Script to run the exploratory analysis
- Notebook to explore the results (may be in a separate PR for size!)

No

No

_No response_

sjspielman:

Closes #1204 

This PR performs an exploratory analysis to determine if we should split `Neuroendocrine` into two categories with the "tumor zoom" cells in their own group. For this, I added another exploratory notebook and associated generation script. 

Here is one of the rendered notebooks: 
[SCPCL000129_separate-tumor.nb.html.zip](https://github.com/user-attachments/files/21144187/SCPCL000129_separate-tumor.nb.html.zip)

Edit: I also updated the GHA to use the Docker image following #1205; drafting this PR until I know we're passing there.

The notebook has the following parts:

- Compare marker genes in the trained model between NE/NE-tumor categories
  - This also includes a comparison of expression values for unique marker genes. Ideally I wanted to use `SingleR::plotMarkerHeatmap()` for this, but it's not available until bioc 3.21 and I'm not trying to get too out of sync with OpenScPCA versioning. So, I instead leveraged `scater::plotExpression` to make this code come together a bit more easily
    - Note that this means `scater` got added into `renv`, but since this notebook isn't run in CI it should be ok to have modified the renv environment even though Docker doesn't know.
- Compare NE marker genes in the combined model to NE/NE-tumor marker genes in the separated model
- Look directly at SingleR inferences: compare between models, assess extent of categorization. This includes a heatmap and some table counts to look behind the scenes of the heatmap
- Score heatmaps
- Plot of delta.median scores for NE & NE-tumor for the separated model

Note that I ended up removing the delta.next investigation I had started, as described here: https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1204#issuecomment-3052375433. Please let know if you have any thoughts on how else we might get at this information? (But, I do think we have enough in this report to go off of either way without needing to spend much more time here and we should focus on getting to actually running SingleR on all samples.)

SingleR results have been uploaded to my bucket `researcher-654654257431-us-east-2`, but I don't necessarily think you need to look at them.

Here I'll just describe notebook findings - 

First, here's a table of cell assignments: How many cells were assigned NE and NE-tumor in each when inferred with the separated model?

| sample id   | library id     | seq unit     | technology | NE cells    | NE-tumor cells |
| ----------- | ----------- | ----------- | ---------- | -----------|----------------|
| SCPCS000104 | SCPCL000121 | nucleus  | 10Xv2      | 23 | 6498                      |
| SCPCS000105 | SCPCL000122 | cell     | 10Xv2      | 5 | 207                        |
| SCPCS000108 | SCPCL000125 | cell     | 10Xv3      | 2 | 39                         |
| SCPCS000109 | SCPCL000126 | cell     | 10Xv3      | 464 | 6882                     |
| SCPCS000109 | SCPCL000127 | cell     | 10Xv3      | 480  | 6504                     |
| SCPCS000111 | SCPCL000129 | nucleus  | 10Xv3.1    | 5275  | 587                     |
| SCPCS000115 | SCPCL000133 | nucleus  | 10Xv3.1    | 3804  | 146                     |

We see strong imbalances in the number of NE and NE-tumor cells inferred in a given sample, where one category has 1-2 orders of magnitude more cells than the other. It's also worth noting that inferences with the separated model with two NE categories took _much_ longer to run compared to inferences with the combined model with a single NE category, presumably due to longer runtime during the fine-tuning step which does suggest 1 category would be preferred.

From the reports:

- Comparing inferences between combined & separated models, results are virtually the same (essentially diagonal Jaccard heatmaps)
  - For the vast majority of cells, NE + NE-tumor cells in the separated model match with NE cells in the combined model 
- Within the separated model, score heatmaps show that NE cells and NE-tumor cells are clustered together except for libraries from `SCPCS000109` which have the NE-tumor cells as an outlier category; these are cell libraries which is bearing in mind! 
	- RBCs are also sometimes also a highly similar category, which generally speaking is a curious result to see RBCs coming up like this in these cells, so we'll investigate inferred RBCs ongoingly throughout the module
- When we look at differences in marker gene expression for unique markers driving the NE/NE-tumor split in the separated model, we see...
	- We can't really say anything robust for comparisons from `SCPCL000122` and `SCPCL000125` (and probably `SCPCL000121`) since there are too few NE cells to get great signal
	- For remaining libraries,
		- NE marker genes show roughly the same expression for NE and NE-tumor
		- NE-tumor marker genes either show roughly the same expression _or_ indeed show high expression for NE-tumor, but across these libraries there appear to be a lot more NE cells with near-0 expression which may obscure interpretation

---> Let's use the model with the single category.

### DIFF
--- a/analyses/cell-type-neuroblastoma-04/.gitignore
+++ b/analyses/cell-type-neuroblastoma-04/.gitignore
@@ -5,3 +5,7 @@
 # Ignore the scratch directory (but keep it present)
 /scratch/*
 !/scratch/.gitkeep
+
+# Ignore certain exploratory notebook htmls
+exploratory-notebooks/reference-aggregation.nb.html
+exploratory-notebooks/separate-tumor.nb.html


analyses/cell-type-neuroblastoma-04/scripts/01_train-singler-model.R
```diff
@@ -93,7 +96,7 @@ if (opts$separate_tumor) {
 # Create and export an aggregated version of the reference
 nbatlas_trained <- SingleR::trainSingleR(
   ref = nbatlas_sce,
-  labels = nbatlas_sce$Cell_type,
+  labels = celltype_label, # specify vector we created above
```
sjspielman, line 99: noting this a bug fix!!

analyses/cell-type-neuroblastoma-04/exploratory-notebooks/separate-tumor.Rmd
```diff
@@ -0,0 +1,543 @@
...
+We expect that these distributions will show some difference (since `SingleR` will find genes no matter what!) between annotated cell types `NE` and `NE-tumor`.
+If the magnitude of the difference is large, it suggests that separating NE into two groups may not be unreasonable.
+If the magnitude of the difference is small, it suggests that separating NE into two groups is oversplitting.
+
+In the plot, each point represents gene expression in a given cell, and the black diamonds are the mean expression for that gene.
+
+That said, note that there are likely very different numbers of cells labeled as `NE` and `NE-tumor`, which may influence how these plots are interpreted.
+```{r}
+celltype_df |>
+  dplyr::filter(stringr::str_detect(singler_label_separated, "Neuro")) |>
```
allyhawkins, line 245:
```suggestion
  dplyr::filter(stringr::str_detect(singler_label_separated, "Neuroendocrine")) |>
```
same here with being specific? I think generally I would just use the full name even though it's a lot of typing.

allyhawkins (commented):
Thanks for making these changes! I agree that the UMAPs have vibes of overlapping NE and NE-tumor which doesn't surprise me. I don't think we need to spend more time on this and can move forward with just using the NE combined.

analyses/cell-type-neuroblastoma-04/exploratory-notebooks/separate-tumor.Rmd
```diff
@@ -0,0 +1,540 @@
...
+```
+
+
+```{r}
+# Define the marker genes
+ne_marker_genes <- ne_markers |>
+  purrr::reduce(c) |>
+  unique()
+
+ne.tumor_marker_genes <- ne.tumor_markers |>
```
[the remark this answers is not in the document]
sjspielman: Noting I opened #1207 to investigate further.

PR #1210, opened 2025-07-10
Explore gene filtering NBAtlas before SingleR

### LINKED ISSUE #1207: Explore whether to remove mito/ribo genes from NBAtlas before running SingleR

Part of #1151

The context for this issue is from this comment: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1206#discussion_r2195796131

In the NBAtlas paper, the authors removed mitochondrial and ribosomal genes from the reference before running SingleR:

> SingleR122 (v1.10.0) cell type prediction of in-house data with the NBAtlas as a reference was run as described above (with default parameters). Ribosomal and mitochondrial genes were removed from the reference to avoid these being selected as genes for mapping.

While we have seen actually pretty good preliminary results without performing this gene filtering, it would be worth comparing to a strategy where the reference is first filtered. This is because we have observed that ribosomal genes were indeed coming up as marker genes in #1206, so it's worth seeing whether this is having a dramatic effect on the cell type results. All of this will be done on an aggregated reference with 

This PR (which may be 2 PRs for size, but it might be fine in 1 PR since my reviewer and I are coordinating!) will contain:

- Updates to the run analysis script with a flag for this filtering behavior
- Updates to the script that trains the singler model to filter these genes if that flag is on
- Exploratory notebook to compare results with and without filtering strategies
- Bash script to reproduce those notebooks

No

No

the plan is before end of sprint!

sjspielman:

Closes #1207 

This PR performs the exploration for whether we should filter out mito/ribo genes. 

This PR contains the following changes:

* Generate script to run the exploratory analysis
* The exploration notebook which has all the good stuff...
  * UMAPs, confusion matrix, heatmaps including comparisons to consensus, and a direct look at the marker genes in the two models  
* Output HTMLs
* README updates
* Bug fix to the model training script - the regex was a bit too strict for ribo genes so I fixed this along the way

### Results

Results between filtered vs unfiltered models are _super similar_ - we have a lot of virtually diagonal heatmaps here! 
We do see that the unfiltered model contains ribosomal marker genes for T cells (which doesn't concern me _as much_ as other cell types), B cells, fibroblasts, pDCs, and NKs. There aren't mitochondrial marker genes though which is good. Aside from these genes, nearly all the other marker genes between models are identical. 

Given that results are virtually identical, I sort of think we should be more conservative and filter since we don't necessarily have great biological intuition for ribo genes being markers? But, I also don't think it will matter much! If we do make filtering the default, I'd also update the default behavior of `run-analysis.sh` to make gene filtering the default.

Here are the notebooks: [filter-genes_htmls.zip](https://github.com/user-attachments/files/21169188/filter-genes_htmls.zip)

renv

N/A

N/A


allyhawkins (approved):
It looks like this didn't really make a difference other than changing a few immune cell classifications. I don't think we need to spend too much time on this so going to go ahead and approve and note that because there is no difference than maybe we just actually use the full one? I don't know that we have enough evidence to confidently say the cells that are getting classified differently are incorrect or not, so I would err on the side of being less stringent. That's just me being cautious because of the controversy over these genes but I'm happy to be convinced otherwise.

PR #1231, opened 2025-07-22
Script to prepare anndata query object

### LINKED ISSUE #1228: Reduce AnnData object size in NBAtlas conversion script
This is a sub-issue of #1214, which is part of the #1151 epic.

The goal of #1214 is to add a script to run scANVI. Part of this will involve reading in the merged SCPCP000004 object. We will not be using the AnnData file we disseminate since we need to convert IDs from ensembl to gene symbol for NBAtlas object compatibility using `rOpenScPCA`.

Therefore, we will add an R script to prepare the input SCPCP000004 data that converts IDs and exports an AnnData object with `zellkonverter`. To save on space/memory, we can also modify the object further:

- Subset to the NBAtlas HVGs
  - SCVI/scANVI operates on the top n (we'll use 2000) HVGs in the reference _only_; data objects are fully subset to those top genes before model training.  
  - The query object is further subset to only these genes with [`SCANVI.prepare_query_anndata()`](https://docs.scvi-tools.org/en/latest/api/reference/scvi.model.SCANVI.html#scvi.model.SCANVI.prepare_query_anndata) ([source](https://github.com/scverse/scvi-tools/blob/70564c397b789943230b900500c557f31905d91b/src/scvi/model/base/_archesmixin.py#L463))
  - Therefore, we can subset to those genes right off the bat. For this, we'll need to input a text file of what those genes are
- Keep only `counts` since `logcounts` aren't used in `scANVI`  

This script will not require any changes to the environment, and will be able to be run on a laptop.

sjspielman:

Closes #1228 
_Not expected to pass CI until #1230 goes in; opening as draft_ Edit: ready for review now!

The goal of this PR is to add a script that prepares our merged object for input to `scANVI`. 

I added a script `scripts/03a_prepare-scanvi-query.R` that converted our merged SCE file into an AnnData file suitable for use straight away by `scANVI`:

- Remove extra object innards we no longer need; a smaller object is better for compute here
- Convert to gene symbols
- Subset to only the NBAtlas HVGs
- Add `colData` columns for consistency with the NBAtlas object
  - The batch and any covariate columns need to have the same names between reference objects
  - We need to have a `Sample` column to match what's in the NBAtlas object, so I make this from `library_id`s.
  - There are also two other covariate columns (tech version and seq unit) that have analogous variation in NBAtlas, so I'd like to have these columns coded up with matching values to test out if they are helpful to incorporate as `scANVI` model covariates (this is separate from the batch variable which we need regardless!). So, I get set up with these columns here

I added this step to `run-analysis.sh`, and I also updated the GHA to download the merged object for input to this script.

renv; runs on laptop

N/A

N/A

### DIFF
--- a/analyses/cell-type-ewings/conda-lock.yml
+++ b/analyses/cell-type-ewings/conda-lock.yml
@@ -13,9 +13,9 @@
 version: 1
 metadata:
   content_hash:
-    linux-64: 2ba52d3db3351616b048111f431b876536ec7441f4503c0a13ad3fe34f1717e8
-    osx-64: e70d56c542bc601cb830deedbf86475f01e6da16279868f89ce9d50218d5597f
-    osx-arm64: 29c444d38241b6b819a81dd17d1b468372592bb4391535c8ca4d3a9463295ffd
+    linux-64: 7bb8c898be4a42c23b8c861b06ac70295ea6a6d976e2aad8b82412da6f13a4d4
+    osx-64: 3345d5852b5ab55cdb02e24c41a6ea6a1e221b4895b19fdf1e9e78ee5bf5b68f
+    osx-arm64: 021cc4f61d15b233bc0d65e4df7830c74df6f5af3258f34749124c9820f36b50
   channels:
   - url: conda-forge
     used_env_vars: []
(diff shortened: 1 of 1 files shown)


analyses/cell-type-neuroblastoma-04/scripts/03a_prepare-scanvi-query.R
```diff
@@ -0,0 +1,95 @@
...
+merged_sce <- merged_sce[intersecting_genes, ]
+
+
+# add colData columns that match NBAtlas naming
+colData(merged_sce) <- colData(merged_sce) |>
+  as.data.frame() |>
+  dplyr::mutate(
+    Sample = library_id,
+    Assay = ifelse(suspension_type == "cell", "single-cell", "single-nucleus"),
+    Platform = ifelse(stringr::str_detect(tech_version, "10Xv3"), "10X_v3", "10X_v2")
```
allyhawkins, line 79: Does NBAtlas only have v3 and not v3.1? Since we have both here and technically 3.1 is different, even if it's minor, so I don't think we want to lump them together?

sjspielman: > Does NBAtlas only have v3 and not v3.1? 

Their metadata has only v3, not v3.1, but it's worth me taking a couple minutes to check back with the source publications to see if there is more info. But either way, that doesn't mean we need to lump _ours_ together - we can still keep v3 and v3.1 separate in our metadata and it should be fine to test out as a covariate that way too. But, if the tech is ambiguous for the NBAtlas, then we shouldn't use it. Stay tuned!

sjspielman: There are 3 datasets with cells with a `10x_v3` designation:

- Dong2020
  - The original paper indicates v3, not v3.1
- Costa 2022
  - The original paper indicates v3.1, which is different from the NBAtlas metadata
- Verhoeven 2022
  - The original paper indicates v2 and v3

This means we actually want to do two things:
- update the NBAtlas Costa2022 metadata to correctly match the actual tech version used, and document that!
- update this ifelse to a case_when to get a 3.1 option in there

sjspielman:
@allyhawkins want to get your opinion on this too - we have several samples that are pretty low-quality here (I'm basing this on # cells, but still want to check QC reports). Here's the first bit of a table sorted on # cells:

```
> colData(merged_sce) |> as.data.frame() |> dplyr::count(library_id) |> dplyr::arrange(n) |> print(n=10)
    library_id     n
1  SCPCL001058    32
2  SCPCL000124    73
3  SCPCL001052   313
4  SCPCL000120   330
5  SCPCL001065   435
6  SCPCL000118   987
7  SCPCL000134  1984
8  SCPCL000137  2318
9  SCPCL000138  2502
10 SCPCL001048  2534
```

I am thinking we definitely want to be excluding the _really_ low quality samples (fewer than 100 cells, looking at you) from scANVI, but I also wonder if it's better to do some (all?) of this filtering in the python script in case we want to fiddle with some threshold that determines samples to remove on the fly. Do you have an opinion on where/when that filtering might happen?

allyhawkins (approved):
Changes here look good. Even if we do filter, I don't think we want to do it here and instead can do it later.

sjspielman:
> Can you remind me if we do any removal of low quality samples prior to annotating with SingleR? I feel like we should use the same approach here in regards to which cells get included. That being said, I do agree that cells from samples with fewer than 100 cells are probably going to be bad and will affect the results so if we filter here we should probably remove those cells in both places.

Yes, we do process all samples with SingleR right now, but since that's one-at-a-time we can just not include the low-quality samples in our final analysis. The actual runtime for each inference is seconds too. I realize now it would have been good to look at those samples' annotations in the exploration notebook, might be something for me to circle back to until this is resolved: https://github.com/AlexsLemonade/OpenScPCA-admin/issues/380. 

It matters more to decide ahead of time for scANVI since it's in an integration framework with all samples at once.

allyhawkins:
> It matters more to decide ahead of time for scANVI since it's in an integration framework with all samples at once.

This makes me more inclined to add a filter for removing low quality libraries to the script that actually runs scANVI so it's easy to adjust. We probably want to evaluate with and without removal of those libraries.

sjspielman:
> This makes me more inclined to add a filter for removing low quality libraries to the script that actually runs scANVI so it's easy to adjust. We probably want to evaluate with and without removal of those libraries.

Yes, that's what I was thinking to do!

PR #1233, opened 2025-07-23
Update conda lock for Ewings

allyhawkins:
In doing some conda testing for Stephanie, I found out that the conda lock file for the Ewing module is out of date. We don't actually use this for any of the workflows still, but conda is included in the Docker container, so we should probably make sure it's up to date.


allyhawkins:
Cancelled running the workflow here since it won't actually use the new docker image and will do a manual trigger after merging and the new image is pushed.

Issue #1221, opened 2025-07-15
Update the non-ETP-ALL and ETP-ALL modules


sjspielman:
Currently, these two cell type modules are not running on real data. These modules were developed in parallel, so they should need the same fix. 

The goal of this issue is diagnose the bug with these modules to be able to get them up and running. Additional issues may be broken out depending on the scope of the fix(es).

sjspielman:
Noting that I had a perfectly fine time setting up the environment and at least starting to run to the non-ETP module on Linux. Previously when I opened this issue, I had only started locally testing on Mac and I encountered issues building the environment, but those issues didn't come back up on Linux. The READMEs do say the module was run on Linux, but it might be good to nail down the environment issues that are hindering running this cross-platform.

PR #1235, opened 2025-07-24
DRAFT: Testing the ETP module

sjspielman:
This is a draft PR to test things out for #1221. It is not intended to be reviewed.

### DIFF
--- a/analyses/cell-type-ETP-ALL-03/scripts/markerGenes_submission.R
+++ b/analyses/cell-type-ETP-ALL-03/scripts/markerGenes_submission.R
@@ -4,7 +4,7 @@
 
 projectID <- "SCPCP000003"
 out_loc <- rprojroot::find_root(rprojroot::is_renv_project)
-project_root <- file.path(out_loc, "..", ".."))
+project_root <- file.path(out_loc, "..", "..")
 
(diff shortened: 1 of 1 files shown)


analyses/cell-type-ETP-ALL-03/scripts/07_run_copykat.R
```diff
@@ -32,4 +32,12 @@ metadata <- metadata[which(metadata$scpca_project_id == projectID &
...
-purrr::walk(libraryID, run_copykat)
+# use a for loop instead of purrr so we can capture errors
+for (library_id in libraryID) {
+  tryCatch(
+    run_copykat(library_id),
+      error = function(e){
+        print(glue::glue("Error running {library_id} through copyKAT."))
+    }
+  )
+}
```
sjspielman, line 43:
```suggestion
}

```

sjspielman:
I want to note that I think I have actually gotten around the CI issues by turning `purrr:walk()` into a for loop and using `tryCatch`. This would have also worked presumably with one of the special `purrr` functions for this (`safely()` or `possibly()`) but this seems just as good. If this passes yet another round of CI, I'm going to move forward with those changes but in a different PR without so much commit litter from testing.

PR #1236, opened 2025-07-24
SingleR report: library size and PDX

### LINKED ISSUE #1234: Explore if library quality affected SingleR performance

Part of #1151 

We have run `SingleR` across all samples in `SCPCP000004` with the NBAtlas reference and explore results here: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/exploratory-notebooks/singler-results.Rmd

Something we didn't consider in this analysis is the relative performance of cells from lower- and higher-quality libraries. Several of the libraries in this project have very few cells (<100), and a couple more that are only a few hundred cells. We'll want to look at the QC reports to get a clearer sense of library quality rather than _just_ using number of cells.

A new section in the `singler-results.Rmd` notebook that looks at annotations across library quality, for example:

- What is the proportion of unlabeled cells for low vs high quality libraries?
- How do annotations for low vs. high compare to consensus cell types?
- Are there differences in marker gene expression?

We're probably with the current renv environment

No, will run on laptop

_No response_

sjspielman:

Closes #1234

The goal of this PR is to see if library quality/size strongly influenced cell type annotations with SingleR for the neuroblastoma samples. I ended up also adding a section of similar scope to look at PDX vs tissue samples to see if there were any differences there.

- First, I removed two _super bad_ libraries from the notebook altogether (see issue comment). These are so rough in the QCs that I don't think it's worth even looking at their annotations. We should definitely remove these from scANVI.
- Then, I added 2 sections at the bottom to look into each of library size and PDX vs tissue, with things like:
  - barplots of proportion of cell types in each library. Probably should have had this from the beginning, and now we have it!
  - heatmaps comparing to consensus
  - a check at the relationship between library size and % unclassified cells

Overall, the smaller libraries (>100, <1000) seem to have reasonable inferences, so I'm not worried about keeping them in the mix for analysis moving forward. The PDXs as well have limited cell type diversity which is not expected, but their annotations seem reasonable too! I ended up not adding in any additional dot plots since it didn't seem like they were needed given the results of the comparisons I did, but let me know if you think we'd benefit from it! I could also, as always, add more UMAPs....

Here is the updated report:
[singler-results.nb.html.zip](https://github.com/user-attachments/files/21416943/singler-results.nb.html.zip)

renv; no changes

N/A

N/A

### DIFF
--- a/analyses/cell-type-neuroblastoma-04/exploratory-notebooks/singler-results.Rmd
+++ b/analyses/cell-type-neuroblastoma-04/exploratory-notebooks/singler-results.Rmd
@@ -632,7 +632,7 @@ The left-most gene set on the x-axis is `Neuroendocrine`, and normal cell types 
 This plot suggests that the normal cells `SingleR` is picking up are likely to be normal cells: normal gene expression for non-NE cells is higher compared to NE cells, and we see the opposite trend for the `Neuroendocrine` marker genes.
 
 
-## Inferences from libraries with lower cell counts
+## Inferences from smaller vs. larger libraries 
 
 In this section, we briefly look at annotation results for libraries with fewer than 1000 cells which may be lower quality than other libraries.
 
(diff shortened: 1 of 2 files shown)


analyses/cell-type-neuroblastoma-04/exploratory-notebooks/singler-results.Rmd
```diff
@@ -620,6 +631,153 @@ ggplot(compare_annotations_expr_df) +
...
+```
+Although there are only four libraries with <1000 cells so it is challenging to make definitive conclusions, it does not seem like the overall cell type distributions are very different between smaller and larger libraries.
+
+### Heatmap
+
+This section shows a heatmap comparing annotations from these libraries only to the consensus annotations.
+
+```{r, fig.width = 8, fig.height = 8}
+make_jaccard_heatmap(
+  celltype_recoded_wide_df |> dplyr::filter(library_id %in% libraries_low_cells$library_id),
```
allyhawkins, line 677:
```suggestion
low_cells_df <- celltype_recoded_wide_df |> 
  dplyr::filter(library_id %in% libraries_low_cells$library_id)

make_jaccard_heatmap(
  low_cells_df,
```
This is so minor but I don't love filtering on one line as input to the function. We should do the filtering first and then provide it as input.

allyhawkins (approved):
I think this looks reasonable. I agree about removing those two libraries after looking at the QC reports. Looking at this there aren't any obvious reasons to exclude the rest of the libraries with low cell counts, so I think it's probably fine to keep them in for now. 

As for PDX, the cell types look as expected to me. I don't think we need to look further at those.

PR #1238, opened 2025-07-28
Patch the (non)-ETP-ALL modules

### LINKED ISSUE #929: Analysis module failing in CI

### LINKED ISSUE #967: Analysis module failing in CI

### LINKED ISSUE #1195: Analysis module failing in CI

### LINKED ISSUE #1221: Update the non-ETP-ALL and ETP-ALL modules

sjspielman:
Closes #1221 
Closes #1195
Closes #929 
Closes #967

This PR gets the `non-ETP-ALL` and `ETP-ALL` running in CI without errors! It also makes a few updates which will help for these modules in the long run. I performed lots of testing in #1235, and the modules have been running through CI without error consistently for a while now so I'm feeling confident about this fix.

Generally speaking, these modules have ~identical code, so I made the same code changes for each module as follows:

1. First, the main bug is in `07_run_copykat.R`. At one point, Josh had added a `tryCatch()` into one of the scripts (#966) but it didn't work. That was because `tryCatch` indeed doesn't work inside `purrr::walk()`. Instead, we'd have to use a function like `purrr::safely()` to capture the error. I did try to get that going in #1235, but it ended up causing the iteration to end entirely when a sample failed `copyKat`. In the end then, I decided to use a for loop with `tryCatch()` so that even if a sample fails, the others still get run through.
2. I updated the modules to include a `run-module.sh` script, and I now call this script in their GHAs. Previously, the modules were calling each individual script at a time from the GHA so this is more modular and maintainable.
3. I further updated the GHAs to use the Docker image, which saves a ton of compute time during environment building. 
  - For this, I had to update paths across all the R scripts... They use `rprojroot` to grab the path but since we're in Docker now, the way to formulate the path had to be updated. Same code change for this is in basically all the R scripts
4. README updates to indicate: the new run script, link to the Docker image, and a caveat that the module has not been tested on other operating systems.

Opening as a draft to make sure I ported all the changes from #1235 over correctly first!


allyhawkins (approved):
Seems good to me! Hopefully this does the trick.

PR #1240, opened 2025-07-28
Add git into (non)-ETP docker

sjspielman:
In #1238, I implement CI fixes for the (non-)ETP modules!

One aspect of that was updating paths because can't use `git` in Docker, but Josh reasonably pointed out: leave the paths alone and just...install `git` into the container: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1238#discussion_r2237030733

This PR just updates these modules' Docker images to get git in there. Note that the modules still have potential to stochastically fail CI at this point, so as long as Docker still builds, it's fine for this PR.


allyhawkins (approved):
LGTM

PR #1248, opened 2025-08-01
Add scANVI query script

### LINKED ISSUE #1214: Add script to run scANVI

Part of #1151

The next step in the `cell-type-neuroblastoma-04` module will be run to scANVI with scArches label transfer using the NBAtlas reference. This issue tracks adding a script for this. Ideally this script will consider the full NBAtlas reference (aka, not the subset).

References:
- Tutorial from `scvi-tools`: https://docs.scvi-tools.org/en/1.3.2/tutorials/notebooks/multimodal/scarches_scvi_tools.html#id1
- scANVI reference: https://docs.scvi-tools.org/en/1.3.2/api/reference/scvi.model.SCANVI.html
- NBAtlas paper code
  - `scVI`: https://github.com/VIBTOBIlab/NBAtlas_manuscript/blob/main/scVI_NBAtlas_manuscript.ipynb
    - Note that NBAtlas appears to only have considered the `Study` batch effect without covariates 
  - `scANVI`: https://github.com/VIBTOBIlab/NBAtlas_manuscript/blob/main/scArches_SCANVI_NBAtlas_v2_manuscript.ipynb 

Python script

Sure will! Incoming separately

Sure will! Stay tuned pending #1212

Next 1-2 sprints

### LINKED ISSUE #1246: Consider library-based approach for scANVI label transfer
> > > Does it have to run on the merged query object is my question. If you train on NBatlas in one step, Can you then run the query on individual samples? Yes, scripts would run serially for now, but in a workflow we could split them up to run in parallel.
> > 
> > I have never seen any examples scANVI used in this way, so I really don't have a sense of how performance would be affected. It's something that can be tested out, but I will say that the cell types that are coming out of this script version seem pretty reasonable so I feel good about this approach. If we were to try integrating one object at a time, I'd want to take some time directly compare those results to the inferences I'm getting here since this the recommended approach.
>
> I'd be somewhat surprised if it gave different results, based on what I know of the method, and it would actually be quite bad for the method. But even so, if we ever expect to run this on a different data sets, or multiple references, it would still make sense to separate the training and querying into different scripts for better modularity.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1245#issuecomment-3140824178_
            
----

This comments linked above are part of a discussion for how to run scANVI/scArches. The current script that has been PR'd in #1245 operates on the merged object, but we'd like to also test the performance on individual libraries to be able to run the label transfer in parallel on one ScPCA library at a time. 

For this, we'd have a script to build and export the scANVI reference model and then a separate script to integrate in a single library. Currently, we are preparing the merged object for input to the scANVI script, so we'd have to also create a new version of or modify of that prep script to prepare a single library, including anndata export, at a time. 

One additional item I also want to note is that we can expect differences in number of epochs when integrating one query dataset at a time, and it's unclear how this will affect runtime.

sjspielman:

Closes #1214
Closes #1246 

This PR adds the 2nd scANVI script which runs labels transfer on the query using the previously built scANVI model on the NBAtlas reference.

- 1 new python script to train the model in `scripts/03c_run-scanvi-label-transfer.py`. It's shorter than the last one!
  - I thought about if we wanted to pull out any code into a utils file to be used by both python scripts, but the main thing I saw was setting up the training dictionary and that didn't really seem worth the effort to me? Let me know if you disagree.
  - For this one, I feel we definitely want the TSV so we can input into an R script that will be part of the pipeline to actually assign the cell types. So all outputs are required for this one.
  - documented in `scripts/README.md`
- Added step to run script in `run-analysis.sh`

laptop, conda

N/A

N/A


analyses/cell-type-neuroblastoma-04/scripts/03c_run-scanvi-label-transfer.py
```diff
@@ -0,0 +1,160 @@
...
+from pathlib import Path
+
+import anndata
+import pandas as pd
+from scipy.sparse import csr_matrix
+import scvi
+import torch
+
+# Define constants
+BATCH_KEY = "Sample"
```
jashapiro, line 17: Is the correct batch the sample or the library?

analyses/cell-type-neuroblastoma-04/scripts/03c_run-scanvi-label-transfer.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    query.obsm[SCANVI_LATENT_KEY] = scanvi_query.get_latent_representation()
+    query.obs[SCANVI_PREDICTIONS_KEY] = scanvi_query.predict()
+
+    ################################################
+    ################ Export objects ################
+    ################################################
+
+    # Export the query-trained scANVI model
+    scanvi_query.save(arg.query_scanvi_model_dir, save_anndata=True, overwrite=True)
```
jashapiro, line 148: How big are these files? Are we ever going to use the query-trained model?

analyses/cell-type-neuroblastoma-04/scripts/03c_run-scanvi-label-transfer.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # Extract data frame of latent representation and predictions
+    latent_df = pd.DataFrame(query.obsm[SCANVI_LATENT_KEY])
+    latent_df = latent_df.rename(columns=lambda x: SCANVI_LATENT_KEY + "_" + str(x))
```
jashapiro, line 152: With your love of `glue`, I would have expected you to use an fstring here!
```suggestion
    latent_df = latent_df.rename(columns=lambda x: f"{SCANVI_LATENT_KEY}_{x}")
```
(I don't think the `str()` is needed with an fstring?)

analyses/cell-type-neuroblastoma-04/scripts/03c_run-scanvi-label-transfer.py
```diff
@@ -0,0 +1,160 @@
...
+    latent_df.index = query.obs.index # set index for joining
+    combined_df = latent_df.join(
+        query.obs[ expected_columns + [SCANVI_PREDICTIONS_KEY] ]
+    )
+    combined_df.to_csv(arg.predictions_tsv, sep="\t", index=False)
```
jashapiro, line 157: Here again I'm wondering if we need the full output? I understand the desire to have both for now, but I feel like we probably want a simpler output option with just the cell and prediction?

jashapiro (commented):
This looks good overall. I had a few questions that I hope are mostly just for information, and then I wondered about whether you need all of the output you are generating or if we can simplify that for most use cases.

analyses/cell-type-neuroblastoma-04/scripts/03c_run-scanvi-label-transfer.py
sjspielman: It's the library, but the name of the column is `Sample` to match the NBAtlas object. I will add a comment to clarify this.

analyses/cell-type-neuroblastoma-04/scripts/03c_run-scanvi-label-transfer.py
sjspielman: >  + whatever scoring metric might be helpful. 

A lot of the calculations to assess model performance need the model object itself, so exporting it to a TSV wouldn't necessarily be helpful. Maybe the "just the results" export option should export the model too, but not also the anndata that fed into it?

jashapiro: I would expect that you would want to include checks for convergence within this script if they are important to decide whether the cell type predictions are valid and should be used downstream. If you are going to include other metrics, we should decide which ones are necessary for every run vs. doing once as a general method check. 

Put a different way: will we get substantially different results from evaluating the reference model vs, the query-tuned models? Do we need to evaluate every query-tuned model? And if that is slow, I would not expect that we want to export the models if we are doing a "just the predictions" export. The thought was that if you are just running to get some predictions, we don't need the intermediates. If you are trying to do a careful evaluation of the results, then you probably want everything.

sjspielman: > I would expect that you would want to include checks for convergence within this script if they are important to decide whether the cell type predictions are valid and should be used downstream

Aside from the `criticism` module (whose output remains to be seen since I cannot generate it, and frankly seems like overkill here...), the main way to assess convergence that I can tell is visualizing loss curves so or based on metrics in `model.history`, which would have to be done interactively in a separate notebook. We can look at some of this directly (currently the code is set to check/record metrics every 5 epochs but this can be bumped!), for example two different loss metrics from a scANVI fit, but aside from something like checking that the final diffs are small I'm not sure what one could do _in the script_?

```
>>> model.history["elbo_validation"]
      elbo_validation
epoch                
4         1195.286865
9         1193.176758
14        1192.034912
19        1191.378174
24         1191.00647
29        1190.807617
34        1190.570435
39        1190.424072
>>> model.history["reconstruction_loss_validation"]
      reconstruction_loss_validation
epoch                               
4                        1153.265015
9                        1151.178467
14                       1150.120239
19                       1149.474976
24                        1149.17981
29                       1149.003662
34                       1148.859985
39                       1148.708374
```

> To clarify my initial: when I was talking about scoring metrics, I was specifically referring to whether there might be a score associated with each prediction, not model-wide scores.

Ah, gotcha! Yes, we can pull out a posterior probability for labels I believe. I'll update to include this in the TSV.

jashapiro:
> * Otherwise, the model as well as an additional TSV of model metrics from `model.history` is exported. I'm attaching a (test data) example here of what that TSV looks like so you can a sense. Let me know what you think of this behavior and how we might tweak it

It is hard to evaluate the export without knowing what the use case you are envisioning is. Is there a reason not to limit the export to the model file and assume that any model evaluation would happen in Python? My general preference would be for as much of that evaluation as is reasonable to happen within this script, at the time of building the model, but I don't necessarily think that needs to be implemented now. I think it may be simplest to just have the option where if the model output is not given, the model is not exported?

PR #1252, opened 2025-08-05
Fix cell-type-wilms-tumor-06: Use the ubuntu 22.04 runner

sjspielman:
The `cell-type-wilms-tumor-06` module is failing in CI with a new `inferCNV` error we have not seen before with this project (#1249). Here I'm seeing if it passes CI on the `ubuntu 22.04` runner instead of the recently-updated latest runner on which it's failing for sample `SCPCS000199`.


sjspielman:
Given https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1249#issuecomment-3150773434 and the fact that this is now passing again with the sample that newly failed, I think this is indeed the move to "downgrade" this runner to the previous ubuntu version for this module's GHA. I'm therefore going to request review here!

jashapiro:
I don't really like this solution unless it is very clearly marked as temporary and there is an issue filed to fix it more generally. We would like to be robust to OS if we can be.

PR #1258, opened 2025-08-08
Patch for CI for wilms-06 module

### LINKED ISSUE #1249: Analysis module failing in CI

sjspielman:
Closes #1249 

This PR hopefully addresses the CI failure in the `cell-type-wilms-tumor-06` module. I identified the section of the code throwing the error, and it's for a step which can be tweaked from the initial `run_infercnv()` call. So, I updated some of the parameters passed into inferCNV if testing is on. Locally on the RStudio Server I was able to reproduce the error before the bug fix, and avoid the error with the bug fix. Let's see if CI agrees! Note that I added in the sample (`SCPCS000199`) which had the error in the first place to run through CI on PRs.


allyhawkins (approved):
🎉

Issue #1266, opened 2025-08-12
Add conda environment for SCimilarity


allyhawkins:
### Describe the goals of the changes to the analysis module.

Before we can actually run SCimilarity (including loading the full model), we need to set up a conda environment with `SCimilarity` installed. 
We should use a similar environment to the one described here: https://genentech.github.io/scimilarity/install.html#conda-environment-setup

### What will your pull request contain?

Here we'll update the `environment.yml` file to include installation of SCimilarity via `pip` and add a `conda-lock.yml` file. We should also add installation of the conda environment to the GHA. 

### Will you require additional software beyond what is already in the analysis module?

Yes, we need to add `SCimilarity`. We also might want to include `anndata` since we will need that to read in the H5AD files.

PR #1268, opened 2025-08-12
Add reference TSV file with cell type annotations and ontology IDs for SCimilarity labels

### LINKED ISSUE #1254: Create a reference TSV with SCimilarity cell types and cell ontology IDs

_No response_

We would like to incorporate SCimilarity annotations into our consensus cell types. Before we can do that, we need to make a table with all possible labels in the SCimilarity model and their associated ontology IDs. The paper and docs say that the cell ontology IDs are available, but we need to actually download the model file and open it to confirm that. 

The model can be found on zenodo here: https://zenodo.org/records/8240464

One thing to note is that there are multiple versions of the model. The link from the cell annotation vignette links you to v0 which has two separate models - one for annotation and one for cell search. We want the one for annotation. However, the most recent version for zenodo just has one combined version of the module for cell search. 

It looks like the list of possible cell types is also available in [Supplementary table 1](https://www.nature.com/articles/s41586-024-08411-y#Sec50), but we should probably use what's in the actual model to make sure we don't miss anything. 

This will probably include a setup script to grab the annotations from the model and save them to a TSV file. 
The PR will also include the actual TSV. 

I'm still torn on if this should be part of the new `cell-type-scimilarity` module or if it should be part of `cell-type-consensus`. I think we want to make the TSV file as part of `cell-type-scimilarity` and then just use that file when we make updates to consensus cell types in that module. 

_No response_

I think if we need to read in the full model that is 30 GB compressed then we probably need to use a resource with enough memory to load that model... The smaller annotation only model is only 7 GB though. 

_No response_

allyhawkins:

Closes #1254 

The goal of this PR is to create a reference TSV file with the cell type annotations and ontology IDs found in the `SCimilarity` foundational model. We can then use this to create an udpated consensus cell type reference. 

- The first thing I did was download the zip file with the `SCimilarity` model from [`zenodo`](https://zenodo.org/records/10685499). I downloaded the most recent version since the file names match what's outlined in the [tutorial for cell type annotation](https://genentech.github.io/scimilarity/notebooks/cell_annotation_tutorial.html). As a side note, I did download the older smaller file for annotation and the labels ended up being equivalent, but I think we should still use the recent version. 
- I was originally going to follow the tutorial to read in the model with `SCimilarity` and then grab the annotations to export, but when I was digging around the files, I found the `annotation/reference_labels.tsv` that matched the annotations listed in the tutorial. So I decided to just write an R script that reads in that file, grabs the associated cell ontology ID for each annotation and then exports it as a TSV file. This TSV file won't actually be used to run the analysis, but I plan to use it to assign ontology IDs to the final annotations and for assigning consensus cell types. 
- I initiated the `run-analysis.sh` script here to include downloading and uncompressing the model (if it doesn't already exist) and then creating the reference TSV file. 

A few caveats: 

- There are 9 labels that did not have a direct match in cell ontology. I created a separate TSV file for those 9 labels and found the most appropriate human readable value that matches. For all but `native cell` and `animal cell` there was a direct synonym. For `native cell`, I just used `cell` and then `animal cell` I used `eukaryotic cell`. 
- There is a suggestion in the annotation tutorial about subsetting the reference to just cell types we want. Jackie and I had talked about not doing that, but I don't love having `cell` or `eukaryotic cell` in the reference? I think we should still include it here since its just a reference table but when we actually run this we might think about subsetting. Here's the text from the tutorial about this: 
> Alternatively, we can subset the reference to just the cell types we want to classify to. This also reduces noise in cell type annotation.

> Note that subsetting can slow classification speeds as the kNN is optimized for the full reference.

Yes

I updated renv here to include the packages needed for mapping ontology IDs. 
Also, the model file is ~ 30 GB so you need to have enough space if you want to download it. 

### Author checklists

_Check all those that apply._

### DIFF
--- a/analyses/cell-type-scimilarity/.Rprofile
+++ b/analyses/cell-type-scimilarity/.Rprofile
@@ -1,4 +1,3 @@
-source("renv/activate.R")
 # Don't activate renv in an OpenScPCA docker image
 if (Sys.getenv('OPENSCPCA_DOCKER') != 'TRUE') {
   source('renv/activate.R')
--- a/analyses/cell-type-scimilarity/.Rprofile
+++ b/analyses/cell-type-scimilarity/.Rprofile
@@ -1,3 +1,4 @@
+source("renv/activate.R")
 # Don't activate renv in an OpenScPCA docker image
 if (Sys.getenv('OPENSCPCA_DOCKER') != 'TRUE') {
   source('renv/activate.R')
--- a/analyses/cell-type-scimilarity/README.md
+++ b/analyses/cell-type-scimilarity/README.md
@@ -8,15 +8,17 @@ For cell type annotation, we use the model available on [Zenodo](https://zenodo.
 
 ## Usage
 
-You can run this module with the `run-analysis.sh` script present in this directory:
+Prior to running `SCimilarity` on all samples, the `setup-analysis.sh` script will need to be used to download the model and create a reference file with ontology identifiers for all cell type annotations. 
 
+This can be done by running the following command: 
 ```sh
-./run-analysis.sh
+./setup-analysis.sh
 ```
+
 
 ## Input files
 
-This module requires the processed `SCE` objects for all projects as H5AD files. 
+This module requires the processed objects for all projects as H5AD files. 
 
 These files can be obtained with the following code run from the root of the `OpenScPCA-analysis` repository.
 You must be logged into your AWS account to download these files.
--- a/analyses/cell-type-scimilarity/README.md
+++ b/analyses/cell-type-scimilarity/README.md
@@ -4,9 +4,15 @@
 
 This module will be used to annotate cell types in all ScPCA libraries using [`SCimilarity`](https://genentech.github.io/scimilarity/index.html). 
 
+For cell type annotation, we use the model available on [Zenodo](https://zenodo.org/records/10685499). 
+
 ## Usage
 
-TBD
+You can run this module with the `run-analysis.sh` script present in this directory:
+
+```sh
+./run-analysis.sh
+```
 
 ## Input files
 
--- a/analyses/cell-type-scimilarity/references/README.md
+++ b/analyses/cell-type-scimilarity/references/README.md
@@ -1,0 +1,15 @@
+# References
+
+This folder contains any reference files used for cell type annotation. 
+
+* `scimilarity-missing-ontologies.tsv`: This file contains a table of all annotations that are part of the [`model_v1.1`](https://zenodo.org/records/10685499) that do not have a direct cell ontology match. 
+For each cell type annotation present in the `scimilarity_celltype_annotation` column we identified the most closely matching name in cell ontology, indicated in the `human_readable_value` column. 
+
+* `scimilarity-mapped-ontologies.tsv`: This file contains a table of all annotations that are part of [`model_v1.1`](https://zenodo.org/records/10685499) and their associated [cell ontology identifiers](https://www.ebi.ac.uk/ols4/ontologies/cl).
+This file was produced by running `scripts/01-assign-ontology-ids.tsv` and contains the following columns: 
+
+| | | 
+| -| - | 
+| `scimilarity_celltype_annotation` | The cell type annotation present in the `SCimilarity` model | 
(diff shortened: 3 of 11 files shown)


analyses/cell-type-scimilarity/.Rprofile
```diff
@@ -1,3 +1,4 @@
+source("renv/activate.R")
```
sjspielman, line 1: renv got aggressive here i think?

analyses/cell-type-scimilarity/README.md
```diff
@@ -4,9 +4,15 @@
 
 This module will be used to annotate cell types in all ScPCA libraries using [`SCimilarity`](https://genentech.github.io/scimilarity/index.html). 
 
+For cell type annotation, we use the model available on [Zenodo](https://zenodo.org/records/10685499).
```
sjspielman, line 7: Just want to note for #1266, from Zenodo

> This release of the model is intended for use with v0.2.0 of the SCimilarity API.

sjspielman (commented):
Just was skimming through, left tiny comments for fun!

analyses/cell-type-scimilarity/scripts/01-assign-ontology-ids.R
```diff
@@ -0,0 +1,86 @@
...
+                                                      scimilarity_celltype_annotation,
+                                                      human_readable_value)) |> 
+  dplyr::left_join(ontology_labels_df, by = "human_readable_value") |> 
+  dplyr::rename(scimilarity_celltype_ontology = ontology_id) 
+
+# check that all labels have an ontology id 
+stopifnot(
+  "Ontology IDs cannot be found for all cell type annotations" = 
+    sum(is.na(labels_df$scimilarity_celltype_ontology)) == 0 
+)
```
jashapiro, line 82: Instead of just quitting here, I would do one of two things (or both).
1. Write out a list of the missing terms.
2. Still create the output file with the NA values.

allyhawkins:
@jashapiro this should be ready for another look! 

- I updated the shell script to be `setup-analysis.sh` and updated documentation based on that change. I'll file a separate issue to track adding this to CI and note the current limitations there. 
- I modified the script for generating the TSV file to read in the annotations file directly and not just the model directory. 
- The script also now has a check for any missing annotations and prints them out rather than quitting. Either way I do write the file out.

jashapiro (approved):
LGTM!

analyses/cell-type-scimilarity/scripts/01-assign-ontology-ids.R
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env Rscript
+
+# This script is used to match cell types present in the SCimilarity model to cell ontology IDs
+# The output is a TSV file that contains `scimilarity_celltype_ontology`, `scimilarity_celltype_annotation`, and `human_readable_value`
```
sjspielman, line 4: Might want to indicate that this exports 2 tsv files

analyses/cell-type-scimilarity/references/README.md
```diff
@@ -0,0 +1,15 @@
+# References
+
+This folder contains any reference files used for cell type annotation. 
+
+* `scimilarity-missing-ontologies.tsv`: This file contains a table of all annotations that are part of the [`model_v1.1`](https://zenodo.org/records/10685499) that do not have a direct cell ontology match.
```
sjspielman, line 5: can we also say that the same script exported this tsv too? (as in the next bullet?)

analyses/cell-type-scimilarity/scripts/01-assign-ontology-ids.R
allyhawkins: It doesn't export 2 TSV files, it only exports the one TSV file with the assigned ontology IDs. The other file I made manually and saved to the repo.

sjspielman: There are definitely two tsvs in the `make_options` which is why I commented (full disclosure I didn't scroll past there 😬), so I guess the `missing` one is stale code and should be removed from the script

PR #1269, opened 2025-08-13
Add scANVI exploratory notebook

### LINKED ISSUE #1261: Add exploratory notebook for scANVI results

Part of #1151 

The goal of this issue is to add an exploratory notebook to assess the scANVI results. This should include:
- Compare of scANVI to SingleR and consensus cell types
- Assess confidence of annotations using posterior probabilities
- Explore of differences between scANVI and SingleR annotations

Exploratory notebook

No

No

This sprint

sjspielman:

Closes #1261

The goal of this notebook is to have an initial look at the scANVI results and compare to SingleR to see if these annotations are similar enough to use in conjunction moving forward for final labels.

I added an exploratory notebook with these sections:

- First, I look at the scANVI posterior probabilities and choose a cutoff of 0.9 which seemed like a good standard to use for these values which are pretty high on the whole!
- Plots: UMAP and heatmaps, as well as some quick stats
- Finally, I look at the fraction of cells per library where SingleR/scANVI exactly agree

I conclude we have similar annotations!

Here's the rendered notebook:
[scanvi-singler-comparison.nb.html.zip](https://github.com/user-attachments/files/21759739/scanvi-singler-comparison.nb.html.zip)

renv

N/A

N/A


analyses/cell-type-neuroblastoma-04/exploratory-notebooks/scanvi-singler-comparison.Rmd
```diff
@@ -0,0 +1,391 @@
...
+    axis.text = element_blank(), 
+    axis.ticks = element_blank()
+  ) +
+  guides(color = guide_legend(override.aes = list(size = 2, alpha = 1)))
+```
+
+
+## Heatmap
+
+This section shows a stacked heatmap comparing the `SingleR` and `scANVI` annotations to the consensus validation groups.
```
allyhawkins, line 297: Didn't we annotate these with the immune zoom version of the atlas? I thought in other plots you've shown you get different types of NK cells and different types of T cells but I don't see that here? I think if this plot is summarizing the comparison to consensus that's fine to keep in, but we also want a plot that looks at all of the originally assigned `NBAtlas` labels comparing `scANVI` to `SingleR`.

analyses/cell-type-neuroblastoma-04/exploratory-notebooks/scanvi-singler-comparison.Rmd
```diff
@@ -0,0 +1,391 @@
...
+  # concatenate vertically into HeatmapList object
+  purrr::reduce(ComplexHeatmap::`%v%`) |>
+  ComplexHeatmap::draw(heatmap_legend_side = "right")
+```
+
+
+We'll also show some statistics from a confusion matrix comparison between `SingleR` and `scANVI`:
+
+```{r}
+# ensure factors are compatible for caret
```
allyhawkins, line 329: I want to confirm that this is looking at agreement between all labels in the immune zoom reference?

analyses/cell-type-neuroblastoma-04/exploratory-notebooks/scanvi-singler-comparison.Rmd
```diff
@@ -0,0 +1,391 @@
...
+
+## Cross-library comparison
+
+Across libraries, what fraction of `SingleR` and `scANVI` labels exactly agree?
+Note that this analysis does not consider similar labels (e.g. two types of T cells) but only does a quick look at identical labels.
+We'll also highlight the PDX samples here to see if there was any associated bias.
+
+
+```{r, fig.width = 10, fig.height = 4}
+frac_agree_df <- celltype_wide_df |>
```
allyhawkins, line 350: There are four libraries that have less than 50% agreement that are both PDX and non-PDX. Are these the ones with low cell numbers? We should look at these and see if there's an obvious reason as to why these disagree so much.

analyses/cell-type-neuroblastoma-04/exploratory-notebooks/scanvi-singler-comparison.Rmd
```diff
@@ -0,0 +1,391 @@
...
+    x = "Library size (n cells)",
+    y = "Fraction of SingleR and scANVI\ncells that agree"
+  ) +
+  theme(legend.position = "bottom")
+
+p1 + p2
+
+```
+
+## Conclusions
```
allyhawkins, line 382: I think it would be helpful to get a table in here that shows you % of each library that has exact agreement between `SingleR` and `scANVI`, % of each library where cells are in the same family (e.g., the same cell is labeled as two different T cell types), and % of each library that disagrees and is not in the same family.

allyhawkins (commented):
Thanks for putting this together! I think it's really helpful to have this as a guide to setting up the final annotations. My biggest comments are about changing the faceted histograms to a different type of plot and then adding in a table summarizing agreement. We should also have a heatmap that looks directly at agreement between `SingleR` and `scANVI` using the full immune zoom NBAtlas labels.

analyses/cell-type-neuroblastoma-04/exploratory-notebooks/scanvi-singler-comparison.Rmd
```diff
@@ -0,0 +1,391 @@
...
+## scANVI posterior probabilities
+
+Before comparing to `SingleR`, we should get a sense of reliability for the `scANVI` results.
+`scANVI` returns a cell-specific posterior probability for each label, so we'll begin by looking at these distributions.
+
+First, what is the distribution of posterior probabilities for assigned labels?
+We'll look at this on a per-library basis to see if any libraries may have less confident annotations in general.
+
+```{r fig.height=12, fig.width=15}
+scanvi_df |>
```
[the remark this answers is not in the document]
allyhawkins: Okay let's keep the histogram for the first plot then.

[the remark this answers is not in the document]
sjspielman: Btw, looking in a ridge plot view (with medians for each distribution also shown) I feel like we can get away with being a lot conservative and drop to 0.75 for a threshold? Either way I'm happy to explore this more if you want, let me know!

Screenshot 2025-08-14 at 12 29 17 PM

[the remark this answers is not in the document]
sjspielman: Yes, that is correct: The total across all labels will be 1. For a value of 0.75, the _most_ the next could be is 0.25, although due to how this is calculated you'll almost certainly never see a true 0 for any cell types (but lots of like `1e-10`).

> I also think we should be less stringent here without losing quality since ultimately we are also looking at agreement with SingleR to get the final label.

Yeah, this is fair, because of the comparison. I really wouldn't want to do _no_ threshold, but I don't need to be so conservative since this isn't the final.final annotation anyways.

analyses/cell-type-neuroblastoma-04/exploratory-notebooks/scanvi-singler-comparison.Rmd
```diff
@@ -0,0 +1,682 @@
...
+  ############ into ggplot ################
+  ggplot() + 
+  aes(x = scanvi_posterior) + 
+  # ugly but _very visible_ color
+  geom_histogram(bins = 100, fill = "magenta", color = "magenta") + 
+  facet_wrap(vars(library_id), scales = "free_y", ncol = 7)
+```
+
+
+Across libraries, the median posterior probability is always exceptionally high at which suggests `scANVI` confidence is roughly the same across libraries.
```
allyhawkins, line 224:
```suggestion
Across libraries, the median posterior probability is always exceptionally high which suggests `scANVI` confidence is roughly the same across libraries.
```

analyses/cell-type-neuroblastoma-04/exploratory-notebooks/scanvi-singler-comparison.Rmd
```diff
@@ -0,0 +1,682 @@
...
+```
+
+While annotation reliability is about the same on a per-library basis, it seems that certain cell types may be much easier to classify than others which may be a factor of how well represented they are in the `NBAtlas` reference.
+There are very few cells in general associated with the cell types that have very low median posterior probabilities (`NKT cell`, `Migratory cDC`, `Circulating NK cell`).
+
+A `0.75` threshold seems like it may be a good middle-ground for preserving labels while removing low-confidence annotations:
+
+```{r fig.height=8, fig.width=8}
+scanvi_ridgeplot +
+  geom_vline(xintercept = 0.75, color = "firebrick")
```
allyhawkins, line 248: I think you can just do this in the first plot, I don't think we need two versions.

allyhawkins (approved):
This looks good to me! Thanks for adding in the table, I think it's really helpful. I also don't think this changes my mind about the strategy that we talked about so I think we're good to proceed with a notebook that does the final assignments and has some final validation plots.

Issue #1255, opened 2025-08-07
Create an exploratory notebook to evaluate incorporating SCimilarity annotations into consensus cell types


allyhawkins:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

This should be added as part of the `cell-type-consensus` module. 

### Describe the goals of the changes to the analysis module.

We want to update our consensus cell type reference to account for annotations from `SCimilarity`. Before we do that, we should explore how adding in these labels may affect the rules that we have currently in place. 

The general idea we have is to assign a consensus cell type label when there is agreement in 2 of the 3 methods. To do this we should calculate the LCA for all possible annotation pairs - SingleR + CellAssign, SingleR + SCimilarity, SCimilarity + CellAssign and then assign an LCA if our existing criteria is met. If we have multiple LCAs that meet our criteria we should choose the most specific LCA as the consensus cell type.

We should create an exploratory notebook that looks at all possible matches and LCAs. We want to see if we keep our current criteria (< 170 descendants for the LCA with exceptions), do we include LCAs that are too broad, and do we lose LCAs that are specific? If so, we might need to update our exceptions. 

### What will your pull request contain?

An exploratory notebook that calculates the LCAs between all possible pairs and investigates cell types retained and lost with our current LCA criteria. We should identify if there are any other exceptions we should make. 

### Will you require additional software beyond what is already in the analysis module?

No 

### Will you require different computational resources beyond what the analysis module already uses?

No

PR #1278, opened 2025-08-18
Exploration of adding SCimilarity annotations to the consensus reference - part 1

allyhawkins:

Towards #1255

Here I am looking at all the possible combinations of cell type annotations between `SingleR`, `CellAssign` and now `SCimilarity` and seeing how we will need to update our approach to assigning consensus cell types. We have discussed that we want to assign a consensus cell type label if 2 out of 3 methods agree on a label. To do that, we need to first get all possible combinations of the three annotations and then get each pair from those combinations. For each pair, we look at the latest common ancestor or LCA and then determine if that LCA meets the criteria we previously outlined and identify if we need to update our criteria. As a reminder, you can find the original guidelines we used for assigning a consensus cell type [here](https://github.com/AlexsLemonade/OpenScPCA-analysis/tree/main/analyses/cell-type-consensus#creating-a-reference-for-consensus-cell-types). 

I am focusing this PR on looking at all possible pairs and updating our rules to assign an LCA to that pair. 

- First the LCA for all pairs is determined. I removed any pairs that did not meet our original criteria (> 170 descendants or part of the list of "broad" cell type terms). Then I looked at the remaining cell types to see if there were any we wanted to throw out that were too specific. I identified a few terms that we should add to our list of celltypes to exclude which I noted in the notebook. 
  - A warning that this is close to 200 cell types compared to the prevous ~50 cell types. I tried to catch all of the cases where there may be "broad" assignments or where the pairs looked funky, but let me know if you think I missed any key ones! 
- Then I looked at all the cases where we have multiple LCAs returned that meet the criteria. In that case I ended up getting a new LCA and if the new LCA (using the original returned LCAs) also meets our criteria, then we use that assignment. If multiple LCAs are again returned I threw them out with the exception of a kidney cell type that only had ~30 descendants. I'm most curious what we think of this approach? Alternatively we could just disregard everything except HSCs that have multiple LCA matches like we were doing previously. 
- Finally, I looked at cell types that would get thrown out using our criteria. Previously we made an exception for neuron and epithelial cell, which I think we still want to do. I found a few more cell types that we probably want to include and added those. 
- For each combination of cell types we then have up to 3 potential LCAs that could be used as the consensus cell type. In the majority of cases there is just one unique LCA so we can use that, but there is a handful (~ 5%) of cases that have 2 or more unique LCAs. This notebook was getting pretty beefy, so I decided to hold off on exploring that for the next PR. 

Yes! The next PR will be to look at deciding on the final consensus cell type label when there are multiple possible LCAs. 

These aren't really "results", but the notebook contains a summary of my thoughts and findings looking into the possible cell type assignments. 

There's a lot of possible cell types here and it's much harder to find exceptions like we did with the labels the first time around. For some context.... there's 1.5 million possible combinations between the three annotations and from that I have ~ 300k combinations that have at least 1 pair with 1 LCA. 

Are there any cell types that are "too broad" that we are including that we should exclude? 

What do you think about calculating the secondary LCA if multiple LCA terms are found? Should we knix this and only consider those with 1 LCA to begin with? 

### Author checklists

_Check all those that apply._


jaclyn-taroni:
I am looking at this and this bit:

> Then I looked at all the cases where we have multiple LCAs returned that meet the criteria. In that case I ended up getting a new LCA and if the new LCA (using the original returned LCAs) also meets our criteria, then we use that assignment. If multiple LCAs are again returned I threw them out with the exception of a kidney cell type that only had ~30 descendants. I'm most curious what we think of this approach? Alternatively we could just disregard everything except HSCs that have multiple LCA matches like we were doing previously.

Seems really hard to express clearly in writing. Is there another feature we can use to choose one from multiple LCAs that reflects a term's specificity?

allyhawkins:
> I am returning some typos I found. I think this looks good. I am concerned that the handling of multiple LCAs will be difficult to explain to users unless we draw a picture, so I'm enthusiastic about finding a different approach. That being said, that could perhaps wait until the next PR.

I definitely agree that the handling of multiple LCAs is not my favorite. In thinking about some possible measures, we could pull the term with the fewest number of descendants or the term that's furthest from the root. I can file an issue to look into other options as a separate PR but include it as a separate section in this notebook if that works for you?

Issue #1272, opened 2025-08-14
Create TSV of normal cell types associated with diagnoses


sjspielman:
We plan to add `inferCNV` to `scpca-nf`. To help us prepare for this, we will need to determine the normal cell types to use as a reference with ScPCA samples. We'll do this on a _diagnosis-specific_ (not project-specific, as projects can have multiple diagnoses!) basis.

The goal of this issue is to establish a TSV file with columns as described:
- Broad diagnosis
  - We'll use _this_ information to determine normal cell types
  - Groups will be the same as we used for manuscript: B leukemia, T leukemia, brain and cns (_sans NB_!), sarcomas, and then the rest on a case-by-case basis (aka these aren't actually broad groups anymore)
- Sample diagnosis
  - This column will have all the diagnoses in ScPCA, matching what values are in the sample metadata. In the end, we'll use this column to grab relevant normal cells for the given sample being processed
- Normal cell types
  - Comma-separated list of consensus cell type labels, similar to how we format the `organs` column in the [cell type reference tsv](https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/references/celltype-reference-metadata.tsv)

This file should live in the `infercnv-consensus-cell-type` module, which will also need a smidge of README reframing since it's not just for pooled normal references anymore.

allyhawkins:
> > Comma-separated list of consensus cell type labels, similar to how we format the organs column in the [cell type reference tsv](https://github.com/AlexsLemonade/scpca-nf/blob/8a85702f7c3e616d1d73335d4139d3c72fabfd4e/references/celltype-reference-metadata.tsv)
> 
> I'm realizing this is going to get very long very fast - and therefore also harder to maintain - with all the immune cell types, so I wonder if we can have broader consensus groups here somehow. We have the validation groups which might be a starting point, though again once we have some results from [#1255](https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1255) we'll have a better sense of where we're headed with consensus in general.

Currently we have around 50 possible cell types but with the addition of SCimilarity looks like that number will be over 200 with so many of them being immune cell types. I agree that we should probably use our broad validation groups to help us here.

PR #1284, opened 2025-08-19
Add draft TSV mapping diagnosis/cell type groups

sjspielman:

Towards #1272 

The goal of this PR is to get moving on exploring consensus cell type distributions. I have made two broad types of changes here:

- Generally speaking, I reframed `README.md` files so the module doesn't present as only about pooled references, and I updated some script/directory names too.
- I added a TSV mapping broad diagnosis groups to validation groups. This is only a draft since it will change with consensus, so I added a TODO in the README. But, having this will at least allow me to start developing a notebook and we can update later once we're ready with `SCimilarity`. Specifically, we can update the TSV once the groups are decided (soonish!), but the notebook itself probably can't be finalized until we actually _run_ `SCimilarity` (this'll be a little while).   

I considered these validation groups to list out cell types: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/consensus-validation-groups.tsv.

I created this "baseline list" of cells we generally expect to be normal, unless we have good reason not to for a given diagnosis:
```
B cell, dendritic cell, hematopoietic precursor cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, T cell, endothelial cell, pericyte, fibroblast
```
Note that I didn't include `adipocyte` here, since we've seen it shouldn't be in Ewings and after reading more about these cells overall, it seems like it could be more generally dodgy. Worth noting, I didn't try to play the game of "catch every single type of normal cell that could be present," and instead opted to be conservative with just this consistent baseline.

In the TSV...

- All the "Other solid tumors" and "Brain & CNS tumors" use this list. I couldn't really find any strong exceptions for any of these diagnoses, so in they all went! Please let me know if you think I may have missed something for one of those.
- For "Sarcoma" I removed `endothelial cell, pericyte, fibroblast`
- And then we have a bunch of leukemias! I included four diagnosis categories based on what I see we currently have in ScPCA, and removed cell types for each that could potentially be tumor. 

Let me know what you think or if you catch something I missed!

Yep


sjspielman:
We're back to this PR following the brief detour in #1286! I made a couple changes since:

- I removed `hematopoietic precursor cell` all around
- I added `endothelial cell, pericyte` to sarcomas and leukemias
- I decided to remove `endothelial cell, pericyte, fibroblast` from Gastrointestinal stromal tumor, since I think this one might be tricky with those cell types
- Based on this discussion back in the day, I went ahead and removed `fibroblast` from WT
  - discussion: https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/808#discussioncomment-10940556
  - This is where we discussed excluding "stromal" cells from the Wilms normal reference. _But_, in this case, the annotation reference we were using didn't contain a distinct `fibroblast` category which _might_ have been ok, but I'm betting it may be hard for cell typing to distinguish between normal fibroblast and tumor so this seems safer to me

analyses/infercnv-consensus-cell-type/references/diagnosis-celltype-groups.tsv
```diff
@@ -0,0 +1,15 @@
+diagnosis_group	celltype_groups
+B-cell leukemia	dendritic cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, T cell, endothelial cell, pericyte
+T-cell leukemia	B cell, dendritic cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, endothelial cell, pericyte
+Myeloid leukemia	B cell, dendritic cell, innate lymphoid cell, macrophage, natural killer cell, plasma cell, T cell
+Mixed-type leukemia	dendritic cell, innate lymphoid cell, macrophage, natural killer cell, plasma cell, endothelial cell, pericyte
```
allyhawkins, line 5: A few comments here: 
- I think we should add fibroblast here. I don't know how often they'll show up in leukemias, but I also don't think endothelial cells and pericytes will be there. They are for sure not leukemia though and if they are there because they are circulating in the blood then I think we want to capture that. 
- For myeloid we should probably remove any myeloid cell types. As we know, different myleoid cells can resemble each other so I would remove macrophage and dendritic cell from that list. 
- The mixed-type is really hard... I don't feel confident in really including any myeloid or lymphoid lineage cells there which may mean we won't have enough to actually run inferCNV, but I think for those it might be okay. There's only 5 samples that fall into this group.

analyses/infercnv-consensus-cell-type/references/diagnosis-celltype-groups.tsv
```diff
@@ -0,0 +1,15 @@
+diagnosis_group	celltype_groups
+B-cell leukemia	dendritic cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, T cell, endothelial cell, pericyte
+T-cell leukemia	B cell, dendritic cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, endothelial cell, pericyte
+Myeloid leukemia	B cell, dendritic cell, innate lymphoid cell, macrophage, natural killer cell, plasma cell, T cell
+Mixed-type leukemia	dendritic cell, innate lymphoid cell, macrophage, natural killer cell, plasma cell, endothelial cell, pericyte
+Sarcoma	B cell, dendritic cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, T cell, endothelial cell, pericyte
```
allyhawkins, line 6: Bone sarcomas have similar phenotypes to monocytes, so I would remove those from the list. I think keeping myleoid is okay, but monocytes in particular I would remove.

analyses/infercnv-consensus-cell-type/references/diagnosis-celltype-groups.tsv
```diff
@@ -0,0 +1,15 @@
+diagnosis_group	celltype_groups
+B-cell leukemia	dendritic cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, T cell, endothelial cell, pericyte
+T-cell leukemia	B cell, dendritic cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, endothelial cell, pericyte
+Myeloid leukemia	B cell, dendritic cell, innate lymphoid cell, macrophage, natural killer cell, plasma cell, T cell
+Mixed-type leukemia	dendritic cell, innate lymphoid cell, macrophage, natural killer cell, plasma cell, endothelial cell, pericyte
+Sarcoma	B cell, dendritic cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, T cell, endothelial cell, pericyte
+Brain & CNS	B cell, dendritic cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, T cell, endothelial cell, pericyte, fibroblast
+Adrenocortical carcinoma	B cell, dendritic cell, innate lymphoid cell, macrophage, monocyte, myeloid, natural killer cell, plasma cell, T cell, endothelial cell, pericyte, fibroblast
```
allyhawkins, line 8: I would remove fibroblasts from any carcinomas or solid tumors that have cells that might resemble fibroblasts. That means at minimum we should remove them from ACC. I'm on the fence on if they should be included in NB, but I think I've convinced myself it's okay to keep there.

allyhawkins (commented):
I left some comments about updating the leukemia group mostly that I think we should address before merging this. 

I also want to note that when I was looking at the map you added for diagnosis, I think we should update some of the names slightly. Sorry I didn't notice this earlier, but they should be minor: 

- For the "myeloid leukemia" group there is only one type, AML, so let's just use "acute myeloid leukemia" as the name rather than shortening it. 
- For the "mixed-type leukemia" they both have the same human readable value from ontology of "Mixed phenotype actue leukemia" so let's used that.

analyses/infercnv-consensus-cell-type/references/diagnosis-celltype-groups.tsv
sjspielman: > The mixed-type is really hard... I don't feel confident in really including any myeloid or lymphoid lineage cells there which may mean we won't have enough to actually run inferCNV, but I think for those it might be okay. There's only 5 samples that fall into this group.

I think this would be a fine outcome if these samples don't get run through. It's a _hard_ one...

sjspielman:
Changes since last time:

- Updated leukemia group names as suggested
- `Mixed phenotype acute leukemia` has been limited to just `endothelial cell, pericyte, fibroblast`. As we've noted this may not get us enough cells, but it doesn't affect many samples, and this one is hard!
- Removed `monocyte` from `Sarcoma`
- Removed `fibroblast` from...
  - `Adrenocortical carcinoma` as suggested
  - `Gastrointestinal stromal tumor`: The stromal compartment here could be confused with fibroblast
  - `Germ cell tumor`: These can potentially, in part, arise from stromal cells
  - `Hepatoblastoma`: These may have mesenchymal components could be confused with fibroblast
- Noting I did _not_ remove `fibroblast` from...
  - Neuroblastoma (as discussed)
  - Melanoma: really different cell type of origin here so seemed less likely these would get confused 
  - Rhabdoid tumor: again, really different cell type that felt sufficiently different to me

analyses/infercnv-consensus-cell-type/references/broad-diagnosis-map.tsv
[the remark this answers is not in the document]
sjspielman: ah! yes. boo manual changes, thank youuu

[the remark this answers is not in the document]
sjspielman: done: https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1284/commits/13a394a8731f1fb42b50275c84b22d56647843f9

Issue #1271, opened 2025-08-13
Setup testing for SCimilarity in CI


allyhawkins:
### Describe the goals of the changes to the analysis module.

In order to run SCimilarity, we need to be able to download a 30 GB model and run a script that will use > 64 GB of RAM. This is going to be a problem for CI since the smaller runners we use do not support even downloading the model, and it takes a very long time! 

We should probably create a testing version of the model to use in CI so that we can decrease the compute requirements and easily run it in CI. 

### What will your pull request contain?

I'm not entirely sure how this will work but one idea I have is to write a script to subset the model that we do not include as part of CI and save that model in a public bucket that we can pull in a GHA.

PR #1285, opened 2025-08-19
Scripts for running SCimilarity 

### LINKED ISSUE #1256: Script to run SCimilarity on a single sample

_No response_

We need to get things set up to actually run SCimilarity on a sample. This PR should include a script that runs SCimilarity on a processed H5 file and exports the annotations. 
We should follow the [annotation tutorial](https://genentech.github.io/scimilarity/notebooks/cell_annotation_tutorial.html) when doing this. 

A script to run `SCimilarity` on a single `_processed_rna.h5ad` file. 
Probably also some environment updates needed to run the script too. 

Yes, we will need to add `SCimilarity` here. 

At least 64 GB of RAM 😢 

_No response_

allyhawkins:

Closes #1256

Here I am getting things setup to be able to run SCimilarity on all ScPCA samples by adding a python script to annotate cell types for a single object and a shell script to run the python script across all samples in a project. The script takes about a minute to run on a single sample, with loading and formatting the object taking longer than actually running SCimilarity, so I think we should be good to go ahead and run this for all samples in ScPCA. 

In the script to actually perform the annotation, I mostly followed what was outlined in the [SCimilarity annotation tutorial](https://genentech.github.io/scimilarity/notebooks/cell_annotation_tutorial.html) with a few tweaks: 

- To nobody's suprise the model uses gene symbols and not Ensembl IDs. In order to run SCimilarity the genes must have the same names and the object must have the genes in the same order. Since we already store gene symbols in the `.var` slot, I ended up first changing the row names to those gene symbols and then using the `anndata.var_names_make_unique()` function to make them unique and avoid index errors related to duplications later on. This behaves similarly to the `unique` option we have implemented in the `rOpenScPCA::sce_to_symbols()` function where they just append a `-1` to the symbol for the duplicates. 
- I also had to create a specific `.layers["counts"]` layer that had the raw counts from `.raw.X` since that's what's expected by `lognorm_counts()`. 
- I chose to use the `get_predictions_knn` rather than `annotate_dataset` function since that just returns the predictions and stats while `annotate_dataset` adds columns to the AnnData cell metadata. 
- For the output, I created a TSV file with the barcodes, cell type annotation, and then the `min_dist` which is part of the `nn_stats` data frame that gets returned. This was noted as being a good measure of confidence in the tutorial. I can add more of the metrics that are output (see [the docs](https://genentech.github.io/scimilarity/modules/cell_annotation.html#scimilarity.cell_annotation.CellAnnotation.get_predictions_knn) for information on the stats that are output), but I thought having min_dist would be sufficient? 

I also created a shell script that then runs this for all samples in a project. When we assigned consensus cell types in this repo, we did it on a projecty by project basis so that we didn't have to have everything downloaded at once, but I can update it to run on all projects that are currently downloaded instead of on a single project? 

Also note that this is not in CI yet because we need to have a subset version of the model to use without downloading the full one. I will work on that next as part of #1271. 

One other small thing is that I was having environment issues with the server and the existing lock file. I had to downgrade to python3.10 to fix it. Building the environment is part of CI so we will be able to make sure that change doesn't affect installation. 

Yes! Next we need to setup CI to test this and then I plan to run this on everything and start looking at comparing results to existing consensus cell types. 

### Results

There are no results yet, but I did run this on a few samples just to make sure things worked and the output looked as expected. Once I actually run this on all the samples I'll make sure the results are present in S3. 

At least 64 GB of ram so use the server to run this

- Are we okay with how I'm handling the gene id conversion? 
- Are there other metrics we should include in the outputs besides the `min_dist` and the annotation? 

### Author checklists

_Check all those that apply._


jashapiro:
> * To nobody's suprise the model uses gene symbols and not Ensembl IDs. In order to run SCimilarity the genes must have the same names and the object must have the genes in the same order. Since we already store gene symbols in the `.var` slot, I ended up first changing the row names to those gene symbols and then using the `anndata.var_names_make_unique()` function to make them unique and avoid index errors related to duplications later on. This behaves similarly to the `unique` option we have implemented in the `rOpenScPCA::sce_to_symbols()` function where they just append a `-1` to the symbol for the duplicates.

As I start to review this, a first question: do you know if the model includes `-1` gene names? If not, do we want to consider summing counts across identical gene names rather than making them unique?

PR #1286, opened 2025-08-20
Add diagnosis map TSV

sjspielman:
As discussed in review for #1284, this PR takes a quick detour to add a TSV mapping broad diagnosis groups to individual diagnoses in ScPCA. 

I used the map we have currently in the `scpca-paper-figures` repo, and make a new TSV from it where:
- I removed non-cancerous
- leukemias have been broken down by cell type
- the "other solid tumor" group is gone and replaced with the ontology diagnosis
- noting i didn't remove any columns, they all seem maybe useful and it's a small file!

For all of this, I used the `human_readable_value` to recode categories. Please let me know if you prefer I use actual ontology ids to parse the leukemias instead of using string matching based on `human_readable_value`. For that, I'd need to make some separate vectors with the ontologies grouped already and use those in the case_when.

I manually ran this script to add the TSV. I did add the script docs, but I will wait to add `references/README.md` docs for this file until #1284 since conflicts.


allyhawkins (approved):
LGTM, thank you for doing this!

PR #1288, opened 2025-08-20
Function for formatting AnnData prior to running SCimilarity

### LINKED ISSUE #1287: Sum duplicate gene IDs prior to running SCimilarity
> This looks good to me, with only the question about whether we want to sum gene ids. My general feeling is that if the model does not have any genes labeled as `-1` or otherwise modified gene names, then we should probably collapse them. I have not looked into whether there is a simple method to do this with the underlying matrices in anndata/numpy. If it is not simple, then presumably we will want to write a function to do the operation, taking our processed anndata object and creating a minimal anndata object for input with just the counts array and appropriate labels. But I think that can be put off to a separate PR, as everything else here looks good!

I looked into this and there are no `-1` genes in the gene list, so those genes will not get considered. I'm good to sum the genes instead as you suggested. I think when we do that we will want to take the counts from the input AnnData object, group by the gene ID and sum them and then create a new object with just those counts like you suggest. I'm going to merge this as is and then file a separate issue/PR to add a function to do that.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1285#issuecomment-3206952302_

allyhawkins:

Closes #1287 

Here I'm adding a function to create a correctly formatted AnnData object from the processed AnnData for use with `SCimilarity`. As part of this function, I changed the way we handle duplicate gene symbols to now sum the counts rather than make the names unique. 

I added a `format_scimilarity` function which is now used to create a new AnnData object with just the raw counts and the gene symbols de-duplicated by summing the counts. 

I first add the gene symbols as the `var_names` and make sure the raw counts are in `X` before using the `anndata.to_df()` function to convert the `X` slot to a pandas dataframe. Then I was able to group by the gene symbol and take the sum and use that summed data frame to create the new Anndata object. 

One small note is that you need to have the raw counts in `X` to compute embeddings for scimilarity and in `layers["counts"]` for the log normalization step. I make sure this is accounted for when making the new object. 

While here, I also added a scripts READMe and updated some documentation since we did not do that in the last PR. 

Yes! 


analyses/cell-type-scimilarity/scripts/02-run-scimilarity.py
```diff
@@ -10,9 +10,57 @@
...
+        raise ValueError(
+            "The input AnnData object must have a 'gene_symbol' column in `adata.var`."
+        )
+    
+    # set gene symbols as var_names and make sure X has the raw counts
+    adata.var_names = adata.var["gene_symbol"].astype(str)
+    adata.X = adata.raw.X
+
+    # create a DataFrame with raw counts, dropping anything that doesn't have a gene symbol 
+    counts_df = adata.to_df().drop(columns=['nan'])
```
jashapiro, line 48: I think you can keep this sparse if youare willing to be a bit more verbose.
```suggestion
    counts_df = pandas.DataFrame.sparse.from_spmatrix(
        adata.raw.X, 
        columns=adata.var["gene_symbol"].astype(str)
    ).drop(columns=['nan'])
```

Note that this should allow you to remove the two `adata` modifications as well (though I didn't start there, so didn't include that in the suggestion).

analyses/cell-type-scimilarity/scripts/02-run-scimilarity.py
```diff
@@ -10,9 +10,57 @@
...
+    # Collapse duplicates by summing and make sparse
+    collapsed_df = counts_df.T.groupby(level=0).sum().T
+
+    # Build new AnnData with collapsed counts stored as layers and in X
+    # this is expected by SCimilarity
+    adata_collapsed = anndata.AnnData(
+        X = csr_matrix(collapsed_df),
```
jashapiro, line 55: Do you need values in both places, or would it work if only put into the "counts" layer (leaving `X` empty)?

Either way, I would do the conversion just once!

jashapiro (approved):
This looks good to me, but I had a sugquestion about keeping the data frame sparse. It might depend on the version of `pandas` you are using whether this works, but it should reduce memory usage and time. Not that that will be a problem given that this has to run with a lot of memory anyway!

PR #1293, opened 2025-08-21
Handle multiple LCAs by using total descendants

### LINKED ISSUE #1255: Create an exploratory notebook to evaluate incorporating SCimilarity annotations into consensus cell types

### LINKED ISSUE #1289: Explore approaches to handling multiple LCAs
> I am returning some typos I found. I think this looks good. I am concerned that the handling of multiple LCAs will be difficult to explain to users unless we draw a picture, so I'm enthusiastic about finding a different approach. That being said, that could perhaps wait until the next PR.

I definitely agree that the handling of multiple LCAs is not my favorite. In thinking about some possible measures, we could pull the term with the fewest number of descendants or the term that's furthest from the root. I can file an issue to look into other options as a separate PR but include it as a separate section in this notebook if that works for you?

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1278#issuecomment-3207886904_

### LINKED ISSUE #1291: Determine an approach for deciding on the consensus label when more than one pair of annotations has an LCA

_No response_

In #1278 we looked at all possible pairs of annotations that could be observed between SingleR, CellAssign, and SCimilarity. For each pair we get the latest common ancestor (LCA) and if the LCA fits a set of criteria that we define, we consider it as a potential consensus cell type label. 

In a small percentage of scenarios, more than one pair of annotations can return a potential LCA. This is likely to happen when all three annotations are in agreement or at least in the same lineage. We need to figure out a set of rules to use for picking which of the possible LCA's will ultimately be assigned the consensus cell type label. 

This PR should include some exploration of how to identify the best LCA among the possible pairs. In general, I think we should just figure out how to find the most specific term and use that. Another thought I had was to prioritize the CellAssign/SCimilarity agreement since those are likely to be more specific than any agreement with SingleR/Blueprint. 

I already started a section in the notebook added in #1278 to work from, but if that notebook gets too wieldy (see #1289) then we may just want to export the TSV of the combinations and associated LCAs to explore in a secondary notebook. 

No 

No 

After completion of #1289

allyhawkins:

Closes #1289 
Closes #1291 
Closes #1255

Here I'm wrapping up the notebook that looks at updating the consensus cell types with SCimilarity annotations by adding two small sections that explore using total number of descendants to choose a single LCA when multiple LCAs are an option. 

- I looked first at how to choose a single LCA when multiple LCAs are returned for a possible pair of annotations. I looked at the possible combinations of LCAs returned and generally there was a more specific annotation that corresponded to the one with fewer descendants. Looking at this I didn't really see a reason to not just use the lowest number of descendants? 
- I added to the section that looks at assigning the actual consensus cell type using the LCA for each of the pairs. 
  - For the majority of cases, there's 1 unique LCA, so we just use that. 
  - If there's more than 1 case it appears to be because they are from the same lineage (e.g., T cell and CD4 T cell). In that case, we again take the fewest number of descendants/ more specific term. 
  - There is one case where there's a tie between hematopoietic precursor cell and stem cell. I ended up choosing to use hematopoietic precursor cell because those all correspond to matches with the HSC lineage in blueprint. 

Here's a copy of the rendered report for easy review: 
[06-update-consensus-scimilarity.html.zip](https://github.com/user-attachments/files/21921899/06-update-consensus-scimilarity.html.zip)

Yes, next up we need to write a script to create this new reference file 

### Author checklists

_Check all those that apply._


jaclyn-taroni (approved):
LGTM

PR #1294, opened 2025-08-21
Allyhawkins/1271 testing scimilarity

### LINKED ISSUE #1271: Setup testing for SCimilarity in CI

allyhawkins:

Closes #1271 

Here I'm making a few modifications to hopefully allow for us to be able to test the scripts for running `SCimilarity` in CI.

After some digging, it looks like the model in zenodo has two difference pieces - one used for annotation and one used for cell search. We aren't using the cell search functionality and the files in that folder are > 20 GB while the annotation files are only around 8 GB. I also confirmed looking at the source code for the [`CellAnnotation` function](https://github.com/Genentech/scimilarity/blob/888e80323ec9e17da556a92754fb06bc7f1fadfd/src/scimilarity/cell_annotation.py#L37-L49) that it only looks for files in the `annotation` folder. I tested running the script using a minimal model with only the `annotation` directory and no `cellsearch` directory and things worked with no warning messages or anything from `SCimilarity`. So I think we can use a smaller version of the model in CI rather than grabbing the full model from zenodo and unzipping it which takes up too much space. 

We needed a central place to store this smaller version of the model that we can easily grab and download in CI since it's still ~ 8 GB so obviously cannot go on GitHub. We are planning to eventually run this as part of `scpca-nf` and we have a folder with publicly available cell type references in `s3://scpca-references`, so I temporarily added the full model there. You can find it in `s3://scpca-references/celltype/scimilarity_references`. 

Then, I udpated the `setup-analysis.sh` script to have an `annotation_only` option that downloads the smaller version of the model with just the annotation. I still left the default to grab it from zenodo, but now in CI we can run that script with `annotation_only=1`. 

Finally, I updated the GHA for this module to download some example projects, download the annotation only model, and then run the SCimilarity scripts. I just randomly picked some projects since I don't think we want to require all of them to run in CI? 

I also expect this will still be a little slow since 8 GB is still a reasonably sized download, but it is much more reasonable than > 30 GB. 

yes 


analyses/cell-type-scimilarity/setup-analysis.sh
```diff
@@ -7,9 +7,15 @@
 
 # Usage:
 #
-# ./run-analysis.sh
+# ./setup-analysis.sh
 #
 
+# By default the full model will be downloaded from zenodo
+# To save space, `annotation_only=1` can be used
```
jashapiro, line 14: I think I would reverse these options, but either way...

jashapiro (approved):
Looks good. I might just reverse the options, as I expect the annotation only version is the one we will want to default to? I don't think it is too important one way or the other though!

jashapiro (approved):
👍🏼

analyses/cell-type-scimilarity/setup-analysis.sh
```diff
@@ -20,28 +25,45 @@ cd ${module_dir}
...
 # compressed output file and final model directory
-compressed_scimilarity_model="${model_dir}/model_v1.1.tar.gz"
 scimilarity_model_dir="${model_dir}/model_v1.1"
 
-if [[ ! -d $scimilarity_model_dir ]]; then
+if [[ ! -d $scimilarity_model_dir && $zenodo -eq 0 ]]; then
+
+  echo "Downloading only the annotation SCimilarity model from s3"
+  s3_model='s3://scpca-references/celltype/scimilarity_references/model_v1.1'
+  aws s3 cp $s3_model $scimilarity_model_dir --exclude "cellsearch/*" --recursive --no-sign-request
```
jashapiro, line 47: Just to clean up the GHA output a bit.
```suggestion
  aws s3 cp $s3_model $scimilarity_model_dir --exclude "cellsearch/*" --recursive --no-sign-request --no-progress
```

PR #1306, opened 2025-08-25
Docker image for cell-type-scimilarity

allyhawkins:
#### Please link to the GitHub issue that this pull request addresses.

Cloess #1303

#### What is the goal of this pull request?

Here I'm just filling out the Dockerfile for the `cell-type-scimilarity` module to install both the R and conda environments. 

#### Briefly describe the general approach you took to achieve this goal.

I mostly just used the approach we've been using in the other modules to have both renv and conda. I also uncommented building the image in the GHA. I tested building this locally and then went into the image to make sure scimilarity was able to be loaded and all looked good. 

#### If known, do you anticipate filing additional pull requests to complete this analysis module?

Yes! In my next PR I'll update the GHA to actually use the image to run the workflow rather than rebuilding the environment. 

#### Analysis module and review

- [ ] This analysis module [uses the analysis template and has the expected directory structure](https://openscpca.readthedocs.io/en/latest/contributing-to-analyses/analysis-modules/).
- [ ] The analysis module `README.md` has been updated to reflect code changes in this pull request.
- [ ] The analytical code is documented and contains comments.
- [ ] Any results and/or plots this code produces have been added to your S3 bucket for review.

#### Reproducibility checklist

- [x] Code in this pull request has been added to the GitHub Action workflow that runs this module.
- [x] The dependencies required to run the code in this pull request have been added to the analysis module `Dockerfile`.
- [ ] If applicable, the dependencies required to run the code in this pull request have been added to the analysis module conda `environment.yml` file.
- [ ] If applicable, R package dependencies required to run the code in this pull request have been added to the analysis module `renv.lock` file.

PR #1307, opened 2025-08-25
Include ontology IDs in the SCimilarity output

### LINKED ISSUE #1302: Include SCimilarity ontology IDs in the output TSV

_No response_

Currenlty in `cell-type-scimilarity`, we output a TSV file with the assigned annotations as a column but no ontology ID. We have a reference file with the ontology ID for each of the annotations, so we should incorporate those into the final TSV file so that we have both the human readable annotation and the id. 

We could add the TSV file with the ontology IDs as an argument to the current script for running SCimilarity and then join it with the predictions data frame prior to exporting the file. I think this will be more efficient than having a completely separate R script that reads in the cell type annotations TSV and joins with the ontology IDs. 

No 

No 

Next sprint

allyhawkins:

Closes #1302 

This PR modifies the output from running SCimilarity to include two new columns - the Cell Ontology ID and the name associated with that ID. 

- I added an argument to the script that runs SCimilarity to read in the ontology ID map file that we have and then join that with the output annotations. This adds a column with the ID and a column with the name associated with that ID as specified in Cell Ontology. For the most part this is the same but not 100 % of the time which is why I kept that column there. 
- While working on this I decided I didn't like "human_readable_value" as a name for a column, so I changed to `cl_annotation`. That's the name we use in the consensus cell typing module and I also noticed it was being used in the NB module, so I like it. There's a number of small changes here to account for that change. 
- The GHA for this module is also now using the Docker container added in #1306. 
- I also filled in the output section of the README to be able to document the columns that are included in the output right  now. 

Yes

### DIFF
--- a/analyses/cell-type-neuroblastoma-04/README.md
+++ b/analyses/cell-type-neuroblastoma-04/README.md
@@ -21,6 +21,7 @@ We used the following approach to assign final labels:
 See [`references/nbatlas-label-map.tsv`](references/nbatlas-label-map.tsv) and [`references/README.md`](references/README.md) for how labels were mapped to families.
 * If `SingleR` and `scANVI/scArches` labels disagree, we then compare to the consensus cell type; if at least one inference agrees with the consensus cell type, we assign the inferred label.
 We perform this first at the label level and then at the broad family level.
+* If the consensus cell type is `Unknown` and one of `SingleR` or `scANVI/scArches` labels is `Neuroendocrine` and the other label is one of `Schwann`, `Stromal other` or `Fibroblast`, we assign the `Neuroendocrine` label
 * Finally, following the conclusions of Bonine et al. (2024), we designate any cell labeled as `Neuroendocrine` as a tumor cell.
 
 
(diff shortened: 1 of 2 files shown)


analyses/cell-type-scimilarity/scripts/01-assign-ontology-ids.R
```diff
@@ -37,7 +37,7 @@ opts <- parse_args(OptionParser(option_list = option_list))
 
 # check that annotations exist
 stopifnot(
-  "--model_dir is missing annotation/reference_labels.tsv" = file.exists(opts$model_annotations_file),
+  "--model_annotations_file is missing annotation/reference_labels.tsv" = file.exists(opts$model_annotations_file),
```
jashapiro, line 40:
```suggestion
  "--model_annotations_file does not exist" = file.exists(opts$model_annotations_file),
```

analyses/cell-type-scimilarity/scripts/02-run-scimilarity.py
```diff
@@ -155,6 +170,8 @@ def main() -> None:
       "scimilarity_celltype_annotation": predictions.values, 
       "min_dist": nn_stats["min_dist"]
     })
+    # add in ontology IDs
+    predictions_df = predictions_df.join(ontology_map.set_index('scimilarity_celltype_annotation'), on='scimilarity_celltype_annotation')
```
jashapiro, line 174: I'm assuming if the index was already set you don't need to reset it here. I also added validation, which I think is correct? 
```suggestion
    predictions_df = predictions_df.join(ontology_map, on="scimilarity_celltype_annotation", validate="many_to_one")
```

jashapiro (approved):
This looks good to me. I made a couple of small suggestions, based on how I _think_ pandas works, but they could be all wrong!

PR #1309, opened 2025-08-25
More tumor cells

### LINKED ISSUE #1283: Discussion: NB tumor cell labels
In the NB annotations being performed in #1282, we are assigning cells as `Neuroendocrine` (aka, the tumor category) only if both SingleR and scANVI are `Neuroendocrine`. This goal of this issue is to consider whether we might want to also have "low-confidence" tumor annotations in the object. Specifically, we generally assume that "Unknown" consensus cells are more-likely-than-not to correspond to tumor cells. Currently, if we have a cell that is labeled `Neuroendocrine` by _one of_ SingleR or scANVI and is `Unknown` in consensus, this cell is getting labeled `Unknown`. 

This ends up potentially missing a lot of tumor cells. In #1282, we are annotating 85% of cells, with 27952 cells remaining `Unknown`. Of those, 77% (21609 cells) would become low-confidence tumor - their SingleR _xor_ scANVI is `Neuroendocrine`, and consensus is `Unknown`. If we were to label these e.g. `Neuroendocrine (low confidence)`, then we will have annotated 96.6% of cells (a remaining 6343 cells unannotated).

sjspielman:
Closes #1283 

This PR updates the NB annotation notebook to include more tumor cells as discussed. We are now a bit above 94% of cells annotated, and all libraries now have >50% cells annotated.

While implementing this, I did not end up putting in any confidence columns though because a) it would really make the parsing code a mess (I'd have to repeat the case when conditions for a confidence column), and it's a really amorphous concept here anyways. I think the documentation we have in this module is pretty clear. I can still think more about a "confidence" concept though if we really want to include it.

It's so orange!
Screenshot 2025-08-25 at 3 51 55 PM

### DIFF
--- a/analyses/cell-type-neuroblastoma-04/README.md
+++ b/analyses/cell-type-neuroblastoma-04/README.md
@@ -21,6 +21,8 @@ We used the following approach to assign final labels:
 See [`references/nbatlas-label-map.tsv`](references/nbatlas-label-map.tsv) and [`references/README.md`](references/README.md) for how labels were mapped to families.
 * If `SingleR` and `scANVI/scArches` labels disagree, we then compare to the consensus cell type; if at least one inference agrees with the consensus cell type, we assign the inferred label.
 We perform this first at the label level and then at the broad family level.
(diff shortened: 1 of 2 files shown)


allyhawkins (approved):
LGTM! I am fine not including a confidence level. I think when you port this module to OpenScPCA-nf you should include the scanvi and singleR annotations as columns in addition to the final annotations. The results will then be available in the results bucket so you don't need to actually add them to the repo as you mentioned.

PR #1312, opened 2025-08-28
Update creation of consensus reference script to include SCimilarity

### LINKED ISSUE #1298: Update creation of consensus reference script to include SCimilarity

_No response_

Now that we have established the rules that we want to use for updating consensus cell types, we need to update the consensus cell type reference file. We currently have a script, [`03-prepare-consensus-reference.R`](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/scripts/03-prepare-consensus-reference.R) that we use to create the file that should be updated. 

I think we could just update the existing script and reference TSV file to now have additional columns for the SCImilarity annotations and additional rows to account for the new matches. 

Alternatively, we could create a new script that is specifically for creating the consensus reference incorporating all three annotations and create a new consensus reference file, but since we're moving towards having the SCimilarity annotations present, I think we should be okay to modify the existing file. 

I will note that this TSV will be much larger and have ~300K rows. 

No 

No 

Next sprint

allyhawkins:

Closes #1298 

Here I'm updating the script to create the consensus cell type reference to incorporate SCimilarity. This also updates the actual reference file. The new reference file now contains a row for all possible combinations of annotations from blueprint, panglao, and scimilarity and columns for the final consensus cell type annotation as well as a column for the LCA for each possible pair. This means if only 2 out of 3 methods are present a cell type can still be assigned if those 2 methods agree. If an LCA is not assigned for a pair then I left `NA` in the column. 

- The script that we were using to create the reference has now been updated to incorporate the `SCimilarity` annotation. Here I mostly copied from code used in the exploratory notebook with some modifications, mainly to make sure we are keeping the LCA for each pair and adding back in the original names for annotations in each reference. 
- I also updated documentation throughout to document the new columns in the reference file and describe this new strategy. 

I also expect this might cause issues with checks since we updated the reference file that gets used, so I might make some additional fixes depending on what happens there. 

Yes 

I was able to run this locally in the project. 

How do we feel about keeping the `NA` in there for cases where no LCA is found between a pair? We could change this to "unknown". 


allyhawkins:
> > How do we feel about keeping the NA in there for cases where no LCA is found between a pair? We could change this to "unknown".
> 
> @allyhawkins, can you comment on the downstream ramifications of this choice?

When we assign final consensus cell types in each object we leave the ontology id (`consensus_celltype_ontology`) column as `NA` if no consensus cell type can be assigned and then the annotation column (`consensus_celltype_annotation`) gets set as `unknown`. So, at minimum, I think we want to keep the `*_ontology` columns with `NA`, but we could set the `*_annotation` columns to have `unknown` in that scenario? 

One other thing to keep in mind is that setting the `consensus_celltype_annotation` column to `unknnown` happens after assigning consensus cell types because the reference only contains combinations that could produce a consensus cell type. If the combination isn't in the reference, then we still need to have code to change it to `unknown` regardless of what we set here.

PR #1314, opened 2025-08-29
Add rtracklayer to environment

sjspielman:
Continued debugging for the `cell-type-neuroblastoma-04` port to OpenScPCA-nf...

The new code is using `rtracklayer` to parse the GTF to get restricted genes when building the SingleR model, but in this repo we had previously used an SCE object instead of the GTF. So, this package isn't in the image. Here, I'm adding it to the dependencies file to snapshot into renv. Hopefully this builds without me needing to add a separate install line for `Rhtslib` which is now included, let's find out!


sjspielman:
The Docker build made it past the new and annoying bits successfully, so this is ready for review now.

allyhawkins (approved):
🐈

PR #1316, opened 2025-09-02
Use entrypoint to set conda environments in Docker

jashapiro:
I saw the changes in https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1313 and https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1311 go in last week, and I thought we might want to clean up those solutions a bit. 

While Adding the conda environment directory to the `PATH` does means that the correct `python` is used and that makes the libraries correct, it may not set the full conda environment quite as we might like always (it seems to work in these cases, but better to set a good pattern if we can).

The solution I am proposing here is inspired by https://pythonspeed.com/articles/activate-conda-dockerfile using something like their final solution. I create a little bash script that activates the environment, then executes whatever command was requested when launching Docker. This is then used as the `ENTRYPOINT` for all commands. 

This seems to work as expected, so that running `docker run IMAGENAME which python3` gives the python executable in the environment directory, and launching directly into python does allow `anndata` to be loaded as expected.

One other thing I noted while doing this was that the neuroblastoma Docker image was very slow to build, which seems to be because it is installing `scvi-tools` via `pip`, which means you have to compile quite a few things from scratch. @sjspielman is this because you were running into trouble with some dependencies? I was able to include `scvi-tools` in the standard conda dependency list and build the enviroments/Docker image just fine, but obviously you have more testing under your belt so I could be missing something!

### DIFF
--- a/analyses/cell-type-neuroblastoma-04/Dockerfile
+++ b/analyses/cell-type-neuroblastoma-04/Dockerfile
@@ -84,6 +84,9 @@ RUN Rscript -e "proc <- basilisk::basiliskStart(env = zellkonverter::zellkonvert
   basilisk::basiliskStop(proc); \
   basilisk.utils::cleanConda()"
 
-# Set ENTRYPOINT to bash to activate the environment for any commands
-ENTRYPOINT [ "conda", "run", "--no-capture-output", "-n", "${ENV_NAME}" ]
+# Set ENTRYPOINT to activate the environment for any commands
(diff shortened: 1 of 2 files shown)


sjspielman:
> I am happy to update the environment in this branch, but it is really up to you (I expect there might be a bunch of other little updates as well, which would require testing).

Yeah, because of this, it makes more sense for me to do it and test the changes locally.

PR #1323, opened 2025-09-04
Add scimilarity annotations to assigning consensus cell types

allyhawkins:

Closes #1031

Here I'm modifying the existing script for assigning consensus cell types to consider the `SCimilarity` annotations. 

- I added an optional argument to the script to provide the SCimilarity annotations TSV file as input. If that's provided, then they are read in and combined with the other cell type annotations before assigning consensus cell types. 
- I made this optional because as noted in #1321, we do not have SCimilarity annotations for multiplexed libraries (or at least we don't do that in the `OpenScPCA-nf` module). This is because we read in the `_processed.h5ad` file directly and those don't exist for multiplexed (or at least not yet). So for now, this is optional. 
- If annotations are provided then we use the combination of all three and take the annotation that is in the `consensus_annotation` column of the reference file. 
- If `SCimilarity` annotations are not provided then we use the combination betwen SingleR/CellAssign and assign that label as the consensus. 
- I also updated the script to now take in both the validation and consensus markers and include all genes in the output gene expression TSV file. 
- Finally, I updated the workflow script that gets run in GHA to optionally provide the SCimilarity file if it exists. One thing to note here is that we don't yet have SCimilarity results in the workflow results bucket. They are only in staging, so it will only run the script in the absence of annotations. Once we release those then we can update the GHA to download the results first. 

Yes, but first we should port over these changes to the `cell-type-consensus` module in `OpenScPCA-nf`. When we do that we might want to include both the singleR/CellAssign consensus output and the overall consensus so we can easily compare between the "old" and "new" consensus cell types. 

### DIFF
--- a/analyses/cell-type-consensus/assign-consensus-celltypes.sh
+++ b/analyses/cell-type-consensus/assign-consensus-celltypes.sh
@@ -27,6 +27,7 @@ consensus_ref_file="references/consensus-cell-type-reference.tsv"
 
 # marker gene file to use for validation 
 validation_markers_file="references/validation-markers.tsv"
+consensus_markers_file="references/consensus-markers.tsv"
 
 # scimilarity results directory 
 scimilarity_results_dir="${data_dir}/results/cell-type-scimilarity/${project_id}"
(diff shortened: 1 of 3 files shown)


jaclyn-taroni (approved):
A couple of docs things, but I don't need to see this again 👍🏻

PR #1334, opened 2025-09-09
Make sure the new modules are included in the run all workflow

allyhawkins:
I noticed when running the workflow to run all modules that we were missing a few. This adds the new modules that have CI setup to the `run_all-modules.yml` workflow.


sjspielman (approved):
oooof good catch!

PR #1336, opened 2025-09-10
Testing for infercnv module fixes

sjspielman:
This is a draft PR not intended to be reviewed, to try and reproduce errors from the `infercnv-consensus-cell-type` workflow, since I am not (yet? at all?) able to reproduce them locally or on RStudio server. (Might be a bug in the workflow file?)


sjspielman:
Well, this is confusing - the only difference between the run that passed and failed _should_ be that the failed one ran fewer samples which seems to suggest it should succeed... Will come back to this tomorrow fresh!

PR #1347, opened 2025-09-16
Update renv to use ontologyIndex instead of rols

allyhawkins:
Related to #1345

There was an error in the `cell-type-scimilarity` module in the script that assigns ontology Ids to `SCimilarity` annotations.

It looks like `rols::Ontologies()` was hitting an error trying to access the API to grab ontology Ids. They seem to have made some updates to the package on the more recent Bioconductor releases (see the note about using a different API in https://lgatto.github.io/rols/index.html). I'm not sure why it's all of a sudden not working, but in some of the code we use in `cell-type-consensus`, we use `ontologyIndex` instead of `rols` and provide the URL to grab the ontologies from, which also allows us to control the version. 

I'm going to make the change in the `cell-type-scimilarity` module to do that too, which requires a change in `renv`. This PR is part 1 and makes the `renv` change and once this is in, I'll file the PR to change the script so that it can be tested.

### DIFF
--- a/analyses/cell-type-scimilarity/renv.lock
+++ b/analyses/cell-type-scimilarity/renv.lock
@@ -32,47 +32,6 @@
     "Version": "3.19"
   },
   "Packages": {
-    "Biobase": {
-      "Package": "Biobase",
-      "Version": "2.64.0",
-      "Source": "Bioconductor",
-      "Title": "Biobase: Base functions for Bioconductor",
-      "Description": "Functions that are needed by many other packages or which replace R functions.",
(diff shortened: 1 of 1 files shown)


sjspielman (commented):
Looks like there's a decent amount of other package version changes in the diff - are these intentional as updated dependencies or did they accidentally come along for the ride (speaking from experience I am great at hitting "y" to update packages when I shouldn't 😬 ). For example, I see ggplot is now at 4.0.0 which is a brand new major release which I'm not sure we want?

PR #1377, opened 2025-12-18
Fix hello-clusters

sjspielman:
As noted in #1375, `hello-clusters` had some issues! We've addressed this in `rOpenScPCA`, so updating this repo accordingly:

- I added the module to the run all modules GHA
- I updated `renv.lock` with the most recent version (we were already a bit behind, as evidenced by the added `scuttle` too which has been in `Imports` for a minute!)

I ran the module locally with test data sans error, so hopefully we're good here too.


jashapiro:
LGTM. Technically, we should probably skip the module check for the `renv.lock` file, but I'm not going to worry about that now.

PR #1390, opened 2026-02-05
Re-render infercnv/consensus notebook

### LINKED ISSUE #1389: Re-render infercnv/consensus counting notebook
Once consensus cell types are updated, we should re-render this notebook: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/infercnv-consensus-cell-type/exploratory-notebooks/reference-query-counts.Rmd

This gives us a ballpark of how many libraries are run through inferCNV. The main reason to re-render this is to get a sense of how many more libraries might now be run through inferCNV than previously. We also report this in the ScPCA manuscript so if the percentage of libraries run through is very different we'd like to know!

sjspielman:
Closes #1389

This PR re-renders the notebook that inventories libraries we can run through inferCNV. Results are essentially the same (no text changes needed) but we did bump up to ~52% of libraries (excluding any multiplexed that would be run through).

This means the text we currently have here that we run about half of the libraries can stay as is:
https://github.com/AlexsLemonade/ScPCA-manuscript/blob/1f6f008c6a29339bc6d9f67b7070b15e623627e4/content/05.discussion.md?plain=1#L26

Here's the current notebook:
[reference-query-counts.nb.html](https://github.com/user-attachments/files/25100457/reference-query-counts.nb.html)

For context, here's the last version: https://htmlpreview.github.io/?https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/infercnv-consensus-cell-type/exploratory-notebooks/reference-query-counts.nb.html


allyhawkins (approved):
🚀

PR #1397, opened 2026-04-03
Draft PR for debugging CI failures

sjspielman:
the errors in #1393 and #1396 may be a GHA runner problem, looking into that here.


sjspielman:
Def not the runner - also, as I should have done first, I pulled the docker image here locally and checked the xfun version. It's 0.57 in the image, but 0.43 in renv. Hm!
https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/b09e24cab2c91df75a51ab98742c1792e8a0973f/analyses/doublet-detection/renv.lock#L2964-L2966

PR #1411, opened 2026-07-06
Get cell-type-wilms-tumor-06 docker building

sjspielman:
~This is a draft PR used for testing fixes.
I will update this comment and undraft if/when we have confirmed fixes.~

This PR is now the PR to address the Docker image for completing #1416.
The image for `cell-type-wilms-tumor-06` is now building, and after this goes in we can confirm if it runs in CI with the testing data.

PR #1417, opened 2026-07-07
Update simulate-sce to use bioc 3.20

### LINKED ISSUE #1415: Bump simulate-sce to 3.20
This module's docker image isn't building. A fix for this was addressed in this commit in a draft PR https://github.com/AlexsLemonade/OpenScPCA-analysis/pull/1411/commits/e8ce0031ecabde7ab2da347ce13d5ba8b93fead0, by pinning rcpparmadillo due it's current version's incompatibilities with lisi which scpcaTools brings into the mix.

But, more generally, since this is a core module we should update it to bioc 3.20. 

This issue tracks bumping the bioc, and ensuring it builds by pinning rcpparmadillo as needed.

sjspielman:
Closes #1415 

This PR does a renv revamp to bioc 3.20. I initially was going through packages updating them, but this was tedious and weird so I nuked it and just re-init'ed at 3.20.

 At my initial filing, I'm not including the rccparmadillo pin from #1411; let's see if we need it here first.

### DIFF
--- a/analyses/simulate-sce/.Rprofile
+++ b/analyses/simulate-sce/.Rprofile
@@ -1,4 +1,3 @@
-source("renv/activate.R")
 # Don't activate renv in an OpenScPCA docker image
 if(Sys.getenv('OPENSCPCA_DOCKER') != 'TRUE'){
   source('renv/activate.R')
--- a/analyses/simulate-sce/.Rprofile
+++ b/analyses/simulate-sce/.Rprofile
@@ -1,3 +1,4 @@
+source("renv/activate.R")
 # Don't activate renv in an OpenScPCA docker image
 if(Sys.getenv('OPENSCPCA_DOCKER') != 'TRUE'){
   source('renv/activate.R')
--- a/analyses/simulate-sce/Dockerfile
+++ b/analyses/simulate-sce/Dockerfile
@@ -49,7 +49,12 @@ RUN Rscript -e "install.packages(c('remotes', 'renv'))"
 WORKDIR /usr/local/renv
 COPY renv.lock renv.lock
 # install packages with renv and remove caches
-RUN Rscript -e "renv::restore()" && \
+RUN Rscript -e "install.packages('renv')"
+
+WORKDIR /usr/local/renv
+COPY renv.lock renv.lock
+# install packages with renv and remove caches
+RUN Rscript -e "options(renv.config.install.remotes = FALSE); renv::restore()" && \
   rm -rf ~/.local/share/renv && \
   rm -rf /tmp/downloaded_packages && \
   rm -rf /tmp/Rtmp*
--- a/analyses/simulate-sce/Dockerfile
+++ b/analyses/simulate-sce/Dockerfile
@@ -1,5 +1,5 @@
 # Docker image for the simulate-sce analysis module
-FROM bioconductor/r-ver:3.19
+FROM bioconductor/r-ver:3.20
 
 # Labels following the Open Containers Initiative (OCI) recommendations
 # For more information, see https://specs.opencontainers.org/image-spec/annotations/?v=v1.0.1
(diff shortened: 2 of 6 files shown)


analyses/simulate-sce/.Rprofile
```diff
@@ -1,3 +1,4 @@
+source("renv/activate.R")
```
jashapiro: This should be reverted
```suggestion
```

analyses/simulate-sce/renv/settings.json
```diff
@@ -1,15 +1,16 @@
...
   "package.dependency.fields": [
     "Imports",
     "Depends",
     "LinkingTo"
   ],
-  "ppm.enabled": true,
+  "ppm.enabled": null,
   "ppm.ignored.urls": [],
-  "r.version": "4.4.0",
+  "r.version": null,
```
jashapiro, line 12: Should probably be set to the version in the docker base image
```suggestion
  "r.version": "4.4.2",
```

analyses/simulate-sce/renv/activate.R
```diff
@@ -2,7 +2,8 @@
 local({
 
   # the requested version of renv
-  version <- "1.0.7"
+  version <- "1.1.8"
```
jashapiro: I would expect this to be the latest version, as long as we are updating (don't update manually, run `renv::upgrade()`)

jashapiro (commented):
A few minor comments below, but the big one is that I would expect an updated `Dockerfile` with at least the a change to the base image.

sjspielman:
> A few minor comments below, but the big one is that I would expect an updated Dockerfile with at least the a change to the base image.

Yes...this is coming but really what happened was I did not mean to request you for review yet!!! Sorry!! But thanks for the initial feedback :)

sjspielman:
Now this is actually ready for a look!

I want to note that I don't really love the RcppArmadillo fix here because the pinned version needs babysitting upon updates. The other solution to this would be to rip out `lisi` from scpcaTools....longer term this is a `lisi` issue in general - that package apparently isn't maintained and isn't compatible with the current RcppArmadillo version, so we may not want to be keeping it around there anyways. At this time we're not doing any integration metric calculations either, so I see a good argument for removing it from scpcaTools suggestions for now and circling back if/when we actually need to be using that kind of functionality in scpcaTools.

jashapiro (approved):
LGTM, in that it works.

But in regards to:

> The other solution to this would be to rip out lisi from scpcaTools....longer term this is a lisi issue in general - that package apparently isn't maintained and isn't compatible with the current RcppArmadillo version, so we may not want to be keeping it around there anyways. At this time we're not doing any integration metric calculations either, so I see a good argument for removing it from scpcaTools suggestions for now and circling back if/when we actually need to be using that kind of functionality in scpcaTools.

I am a little confused as I would have expected `lisi` to not be installed here at all since it is in the `Suggests` block. And indeed it is not listed in the `renv.lock`

The error before adding the dependency was about `basilisk` not being found (an odd error to be sure, since it was clearly installed), and I can see that it built properly after you added `RcppArmadillo`, but I can't see any reason why that would have solved the `basilisk` problem. 

So I am approving this, but I really do wonder if the prior error was somehow stochastic...

analyses/simulate-sce/Dockerfile
jashapiro: Okay, I did some testing because I was really annoyed by this, and I think the issue is that for _some reason_ only in the Docker build (this does not happen to me locally), `renv` is installing all of the remotes, even though they are only in suggests. The folllowing change seemed to fix it for my local build: 


```suggestion
RUN Rscript -e "install.packages('renv')"

WORKDIR /usr/local/renv
COPY renv.lock renv.lock
# install packages with renv and remove caches
RUN Rscript -e "options(renv.config.install.remotes = FALSE); renv::restore()" && \
```

I think if we add that, we can revert the `RcppArmadillo` pinning.

sjspielman: We share similar annoyances. Let's see if this goes.

sjspielman: it goes!

PR #1421, opened 2026-07-16
Remove clustering workflow for Ewing's from CI

allyhawkins:
Here I'm removing the clustering workflow from CI for the Ewing's module, since the results aren't actually used anywhere for annotation. We did the same thing for the inferCNV module a few months ago, so I just followed the same procedure of removing it from the GHA and then noting both in the readme and at the top of the workflow script that it is no longer run in CI and not maintained. This should hopefully decrease the run time for the module!


sjspielman (approved):
thanks!

Issue #1424, opened 2026-08-03
Evaluate reference approach for annotating osteosarcoma samples


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

https://github.com/AlexsLemonade/OpenScPCA-analysis/discussions/1410

### Describe the goals of the changes to the analysis module.

The OsteoCar reference that we will be using to annotate these samples is split across several files:
```
xeno_prim_human_cancer_cells
patient_prim 
patient_mets
mm_prim
mm_mets 
xeno_prim_mouse 
xeno_mets_mouse 
dogs_prim 
dogs_mets 
xeno_mets_human_cancer_cells 
```

We have all human samples, including primary, metastatic, and xenograft. A question has arisen about to use the references: We would probably like to consider xenograft and patient references separately, but how different would e.g. SingleR results for patient samples under either approach:
- Run SingleR with two references provided (primary and metastasis)
- Run SingleR with one concatenated reference 

We can run this on a few pilot samples and then compare annotations to guide choosing the reference approach. This also seems worth doing: https://bioconductor.org/books/release/SingleRBook/using-multiple-references.html#manual-label-harmonization

### What will your pull request contain?

This will have two PRs:

1. Establish an renv environment and add a script to convert references from Seurat -> SCE. We can save these as individual files, to be concatenated as part of exploration where needed. Depending on that outcome, the conversion script could be modified. Note that I only expect this script to convert to SCE, not to AnnData. The script will be modified later for #1413 once this investigation is complete.
2. An exploratory notebook comparing references and SingleR. I haven't run any of this yet, but it's also possible we get a lot of good insight from just running `SingleR::matchReferences()` as linked above, so that's the starting point for that notebook.

### Will you require additional software beyond what is already in the analysis module?

Yup, renv incoming with SingleR and such

### Will you require different computational resources beyond what the analysis module already uses?

Should be fine on laptop

sjspielman:
Thus far, we have been exploring how to use the OsteoCar references. The authors broke these down into several files based on sample conditions, including based on primary vs metastasis, and patient sample vs human cells from a mouse xenograft. 

[...]

Issue #1413, opened 2026-07-06
Add script to prepare references for use


sjspielman:
### If you are filing this issue based on a specific GitHub Discussion, please link to the relevant Discussion.

#1410

### Describe the goals of the changes to the analysis module.

The first step in annotating samples with OsteoCAR is to get OsteoCAR ready for use. We'll need to get SCE & AnnData objects of the Seurat files from the atlas we'll need.

Code will have similarities to this script we used to prepare the NBAtlas reference: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/scripts/00_convert-nbatlas.R

### What will your pull request contain?

- module analysis run script
- script to prepare the osteocar references
- associated environment updates

### Will you require additional software beyond what is already in the analysis module?

No

### Will you require different computational resources beyond what the analysis module already uses?

No

sjspielman:
For additional context when someone picks this up - A script for this has been started here to convert a _single_ reference qs2 file with a Seurat object to an RDS file with an SCE object: https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-osteosarcoma/scripts/convert-osteocar.R. 

This script currently only converts SCE, but needs to be extended to also convert to AnnData similar to https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-neuroblastoma-04/scripts/00_convert-nbatlas.R.  

Note that it is possible that the scope of this script will need to change as well, depending conclusions from #1424. For example, it may be the case that we conclude to use a harmonized/combined mets/primary reference rather than standalone. In this case, this script would likely want to take additional aggregation steps. Note that when combining references, R runs out of memory if I try to combine them _before_ aggregating each individually with SingleR. This consideration will affect how the script has to change in this circumstance: Likely you would want to aggregate first and then combine, and then convert that object to AnnData. O

PR #1426, opened 2026-08-05
Initial osteocar reference comparison

sjspielman:
Towards https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1424

The goal of this PR is to begin comparing references to each other. Originally I had one pretty large notebook, but I am breaking it up into multiple notebooks for easier review so there's more to come.
As such, this PR contains the following:

* An _initial_ script to convert references to SCE objects. This script will evolve when we implement #1413, since it will also need to convert to AnnData. But, this is good enough for this initial exploration now
  * This is now run in the bash script too 
* A new exploratory notebook which just runs the match references function on _aggregated_ (they are huge) references to get an initial sense of how similar their profiles and cell types are. I do this twice: one for patient prlmary/met references and one for xenograft primary/met. I conclude that the references indeed seem to have different profiles so we should explore them further with some very cursory cell typing in the next notebook. 
  * The other PR I am going to file here will have a notebook that cell types 5 patient samples with different properties to see how different reference strategies come out. This will of course be a very cursory exploration, but it should provide guidance. 
* I also have a renv environment building up, which does not contain more packages than are used in this PR (eg singler), since I was developing a larger notebook at the same time as this. We expect everything that is in renv to be needed anyways so this seemed fine.

Here is this PR's rendered notebook:
[02-compare-references.nb.html](https://github.com/user-attachments/files/30745627/02-compare-references.nb.html)


analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,157 @@
...
+author: "Stephanie J. Spielman"
+date: "`r Sys.Date()`"
+output:
+  html_notebook:
+    toc: true
+    toc_depth: 3
+    code_folding: hide
+---
+
+There are several different references in `OsteoCar`, and while we can generally map them to our sample propertoes, we have some sample which don't neatly line up.
```
allyhawkins:
```suggestion
There are several different references in `OsteoCar`, and while we can generally map them to our sample properties, we have some samples which don't neatly line up.
```

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,157 @@
...
+
+We'll compare `Ann_Level2` between primary and metastatic references by first aggregating them and then running `SingleR::matchReferences()` and plotting the resulting heatmap.
+This will tell us where cell types overlap/differ between references.
+
+Note that we read in references when we use them and remove them promptly after aggregation, to save memory.
+
+### Primary references
+
+```{r}
+patient_prim <- readr::read_rds(file.path(ref_dir, "patient_prim_sce.rds"))
```
allyhawkins, line 95: Before you dive into creating the trained references, it would be helpful to just see a summary of what cell types are present in each of the references. I want to know which cell types are present in both and which cell types are unique based on label alone. I also want to know which cell types are tumor cells vs not tumor cell types. It looks like we might have some different classes of tumor cells here?

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,157 @@
...
+In the heatmap below, the primary reference is columns and the metastasis reference is rows.
+
+```{r}
+matched <- matchReferences(
+  patient_mets_aggr$ref, 
+  patient_prim_aggr$ref, 
+  patient_mets_aggr$labels$full, 
+  patient_prim_aggr$labels$full 
+)
+pheatmap::pheatmap(
```
allyhawkins, line 204: Can we subset this to only show shared labels in the first place? I want to know if they have the same label in both references, are they similar. If they are different labels then they are probably distinct so we they just add noise to the heatmap and make it harder to interpret.

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,157 @@
...
+  matched, 
+  col=viridis::magma(500), 
+  cluster_rows = FALSE, 
+  cluster_cols = FALSE
+)
+```
+
+There's a 1:1 match of proliferative here only, but the noise around other cell types suggests that there are distinct expression profiles between shared cell types across these references.
```
allyhawkins, line 308: Sure there's only a 1:1 match for proliferative, but I see some fairly strong signals for 4 out of the 6 tumor classes. It would be helpful to see a table of how many of each cell type is present in the original dataset alongside this just for added context. As in, how much of the population is assigned to this "Interactive" population vs the proliferative population?

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,157 @@
...
+What is not clear is how much of that is signal vs noise. 
+The `Interactive` profile also appears to be completely distinct between these.
+
+## Conclusions
+
+Given the differences between primary/metastasis references for both patients and xenografts, it seems worth doing some initial exploration of annotation performance with `SingleR` next.
```
allyhawkins, line 314: I think I would still run these together as a single reference with distinct labels. So you would attach `primary-` or `met-` to each of the labels prior to running. Then you can see which labels are assigned and if you get predominantly primary or met or an even mix of both within a single sample. You could then compare that to running as a single reference with harmonized labels where you don't distinguish between primary and met when building the reference model.

analyses/cell-type-osteosarcoma/scripts/convert-osteocar.R
```diff
@@ -0,0 +1,54 @@
...
+
+# load the bigger libraries after passing checks
+suppressPackageStartupMessages({
+  library(Seurat)
+  library(SingleCellExperiment)
+})
+
+
+# export reformatted object if requested
+if (!is.null(opts$output_sce_file)) {
```
allyhawkins: This seems strange to have a script solely for converting objects to SCE and then only do it optionally? I might make the output file required and remove the nesting you have here.

allyhawkins: I see that you want to expand this script to convert to anndata here too, but I think I would still remove this as being optional and do both conversions automatically.

analyses/cell-type-osteosarcoma/run-analysis.sh
```diff
@@ -54,3 +55,16 @@ download_file $figshare_mets_url $mets_qs2
...
+mets_sce="${ref_dir}/patient_mets_sce.rds"
+prim_sce="${ref_dir}/patient_prim_sce.rds"
+xeno_mets_sce="${ref_dir}/xeno_mets_sce.rds"
+xeno_prim_sce="${ref_dir}/xeno_prim_sce.rds"
+
+
+Rscript ${script_dir}/convert-osteocar.R --input_ref_file $mets_qs2 --output_sce_file $mets_sce
+Rscript ${script_dir}/convert-osteocar.R --input_ref_file $prim_qs2 --output_sce_file $prim_sce
+Rscript ${script_dir}/convert-osteocar.R --input_ref_file $xeno_mets_qs2 --output_sce_file $xeno_mets_sce
+Rscript ${script_dir}/convert-osteocar.R --input_ref_file $xeno_prim_qs2 --output_sce_file $xeno_prim_sce
```
allyhawkins: This is fine, but there's gotta be a way in bash to turn this into a loop that's not terribly ugly and minimizes the duplication. Maybe try this? 
```suggestion
input_files=("$mets_qs2" "$prim_qs2" "$xeno_mets_qs2" "$xeno_prim_qs2")
output_files=(
  "${ref_dir}/patient_mets_sce.rds"
  "${ref_dir}/patient_prim_sce.rds"
  "${ref_dir}/xeno_mets_sce.rds"
  "${ref_dir}/xeno_prim_sce.rds"
)

for i in "${!input_files[@]}"; do
  Rscript "${script_dir}/convert-osteocar.R" \
    --input_ref_file "${input_files[$i]}" \
    --output_sce_file "${output_files[$i]}"
done
```

allyhawkins (commented):
I think this is a good start, but I had a few suggestions to add to the notebook just to help add more context and make the one large heatmap a little easier to interpret.

sjspielman:
Changes include:

- More heatmaps, showing both all cell types and the shared cell types (which of course are sometimes just shared in name only and not quite expression profiles...)
- Barplots for distributions of cell types, as well as some more explicit printing of where they match up and don't between references
- Script updates: use arrays (differently than suggested but still tighter than it was) in `run-analysis.sh` and remove the `if` from the conversion script

[02-compare-references.nb.html](https://github.com/user-attachments/files/30765614/02-compare-references.nb.html)

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
allyhawkins: Just bumping this spelling comment.

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,307 @@
...
+date: "`r Sys.Date()`"
+output:
+  html_notebook:
+    toc: true
+    toc_depth: 3
+    code_folding: hide
+---
+
+There are several different references in `OsteoCar`, and while we can generally map them to our sample propertoes, we have some sample which don't neatly line up.
+For example, `OsteoCar` has patient primary and metastatic (lung only) references, but we have a sample from metastatic tissue in bone which differs from the metastatic reference.
```
allyhawkins: A question for the future, is it really just one sample or is it multiple samples?

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,307 @@
...
+
+The metastasis reference uniquely has these cell types, across several categories.
+
+```{r}
+setdiff(
+  unique(mets_coldata$Ann_Level2), unique(prim_coldata$Ann_Level2)
+)
+```
+
+The majority of cell types, including all classes of tumor cells, appear to be shared, but this does not guarantee they have similar expression.
```
allyhawkins, line 155: It's been a while since I've read this paper, but can you remind me exactly which of these cell types they are calling "tumor" cell types vs. normal cell types? I'm actually surprised that the osteoblasts and chondrocytes population aren't overlapping with tumor, so I want to double check thats the case? Or is tumor vs normal represented by a different column in the data?

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,307 @@
...
+shared
+```
+
+
+We'll visualize this as well with barplots, faceted by cell type category, of cell type counts and proportions for these references for complementary views on their distributions.
+
+```{r}
+#| fig.height: 8
+#| fig.width: 10
+make_barplot(prim_coldata, mets_coldata) |>
```
allyhawkins, line 170: Ah I see now which categories are which. Thank you for adding this since it's helpful to see them grouped by broad cell type.

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,307 @@
...
+
+There are several different references in `OsteoCar`, and while we can generally map them to our sample propertoes, we have some sample which don't neatly line up.
+For example, `OsteoCar` has patient primary and metastatic (lung only) references, but we have a sample from metastatic tissue in bone which differs from the metastatic reference. 
+
+The goal of this notebook is to lightly compare some of the references using `SingleR`:
+* patient primary vs metastatic (lung only) references
+* xenograft primary vs metastatic references
+
+We compare the references to one another to see how similar the cell type distributions and expression signatures are.
+As these references are quite large, they have to be aggregated before use, but we'll also explore their raw data too.
```
allyhawkins, line 21: Can you add a note here about what's in `Ann_Level1` vs `Ann_Level2`? That might help with some of the confusion I was experiencing (but have since resolved).

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
allyhawkins: I actually don't think you need this heatmap at all. It's fine to keep, but if the labels don't match up then the value is going to be 0 anyways so it doesn't totally make sense to me to show both?

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,307 @@
...
+  col=viridis::magma(500), 
+  cluster_rows = FALSE, 
+  cluster_cols = FALSE
+)
+```
+
+Now, only the cell types shared between the references:
+
+```{r}
+shared_matched <- matched[shared, shared]
```
allyhawkins, line 215: This is much more helpful because you can see the cell types that don't have any match that should have a match, which look to be mostly tumor cells. This tells me there probably is a difference in the marker gene expression for tumor cell subpopulations between primary and mets that we should be sure to consider.

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,307 @@
...
+    geom_col(position = position_dodge()) +
+    scale_fill_brewer(palette = "Set1") +
+    ggtitle("Proportions") +
+    theme_bw() + 
+    theme(axis.text.x = element_text(angle = 30, hjust = 1))
+  
+  if (facet) {
+    p_counts <- p_counts + facet_wrap(vars(Ann_Level1), scales = "free", nrow=3)
+    p_props <- p_props + facet_wrap(vars(Ann_Level1), scales = "free", nrow=3)
+  }
```
allyhawkins:
```suggestion
  p_counts <- p_counts + facet_wrap(vars(Ann_Level1), scales = "free", nrow=3)
  p_props <- p_props + facet_wrap(vars(Ann_Level1), scales = "free", nrow=3)
 
```

It looks like you don't actually use the facet option, so you can probably remove it.

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
```diff
@@ -0,0 +1,307 @@
...
+
+module_dir <- file.path(repository_base, "analyses", "cell-type-osteosarcoma")
+ref_dir <- file.path(module_dir, "references")
+```
+
+
+## Functions
+
+```{r}
+make_barplot <- function(prim_coldata, mets_coldata, facet = TRUE) {
```
allyhawkins:
```suggestion
make_barplot <- function(prim_coldata, mets_coldata) {
```

allyhawkins (approved):
This looks good to me. My comments are all fairly minor so I don't need to see this again. I think the biggest thing is just to make sure you add in an explanation at the top about what `Ann_Level1` and `Ann_Level2` are and then you probably don't need two copies of the same heatmap and can just show the shared cells.

sjspielman:
> I actually don't think you need this heatmap at all. It's fine to keep, but if the labels don't match up then the value is going to be 0 anyways so it doesn't totally make sense to me to show both?

I like keeping it because this function does specific things to help you identify cases when cell types are effectively equivalent between references _but the labels don't match_ - that's really its main purpose, but here we're using to compare references more directly. So, we actually do learn something from this heatmap - that a cell type labeled X in one reference apparently does not really correspond to cell type wit different label Y in the other. I will add this additional textual context to explain why to keep ot.

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
sjspielman: Disappointed you don't like propertoes, but fine. 🩴 🍹

analyses/cell-type-osteosarcoma/exploratory-notebooks/02-compare-references.Rmd
sjspielman: Definitely a typo, it's multiple......

```
> sample_df |> dplyr::filter(primary_or_metastasis=="Metastasis") |> dplyr::count(tissue_location)
# A tibble: 10 × 2
   tissue_location           n
   <chr>                 <int>
 1 Back                      2
 2 Femur                     2
 3 Kidney                    1
 4 Left Lung                 1
 5 Left upper lobe lung      1
 6 Lung                     29
 7 Paraspinal                1
 8 Right Lung                1
 9 Right upper lobe lung     2
10 Thigh                     1
```

PR #1427, opened 2026-08-07
Compare patient references

sjspielman:

Towards #1424 

(edit: reminder that I won't be able to get back to this until at least week of 8/17, so review timeline can match!)

This PR adds a notebook where I ran SingleR over a set of samples using several reference strategies, focusing on patient samples and references.

Samples include as control a lung met sample, a bone primary sample, and then the 5 additional patient met samples which do not appear in lung. 
References include metastasis, primary, both provided to SingleR in the same call, or a merged reference where I concatenate them (both ensure labels are identifiable eg T cells from both aren't collapsed) and send that single reference into SinglerR.

To evaluate results, I have barplots of the cell types they infer as well as heatmaps comparing different reference results. For all plots, I focus on just the top 10 cell types returned per reference so it doesn't get too out of control. The heatmaps also include the number of inferred cell types in the labels to provide additional context.

We would hope for a few things here:
* tight correspondence between the two combined reference approaches - we sometimes see this, but sometimes there are pretty salient differences
* tight correspondence for controls: 
  * for a primary bone, I would hope that results from the primary reference are similar to a combined reference, and for a lung metastasis I would hope that results from the mets reference are similar to a combined reference. Admittedly I have only 2 control samples here so it's not a full exploration, but we only see this really for the primary, not the metastasis whose combined refs are coming back with a blend of cell types from the references

Although there is a decent amount of noise in many of these results, we do see some clear cell types consistently emerging across reference strategies. The noise makes it hard to nail down a precise strategy, so I tend to lean towards just using a combined reference, and as I sort of start to get at in the notebook I think this might be the concatenated one (but I remain back and forth here). 

There are other analyses I did not include here that would be good to explore, such as looking at delta next or the SingleR result heatmaps themselves. I didn't want to include them just yet since there's already quite a bit here that I thought would be good to get more eyes on before expanding further in case you see patterns that I am missing, as my eyes now need a break from looking at heatmaps!

The renv environment has been updated as well. Here's the notebook:
[03-compare-patient-references.nb.html](https://github.com/user-attachments/files/30835008/03-compare-patient-references.nb.html)

Note that it's probably true some text in the notebook needs to be fleshed out further, but again I wanted this to go through a round of review before polishing. The conclusions also need to be nailed down and I think we'll benefit from some discussion before doing that.


analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references.Rmd
```diff
@@ -0,0 +1,468 @@
...
+
+* the primary reference
+* the metastasis reference
+* both references provided to `SingleR`
+* a concatenated version of these references with cell types between them kept distinct
+
+This is obviously not a comprehensive look at our osteosarcoma samples, but it may still help guide decisions.
+Note that this notebook does not consider any xenograft references or xenograft samples.
+
+We particularly explore samples with a mismatch from the reference: We have severalnon-lung metastasis samples, but the `OsteoCar` metastasis reference was built from only lung tissue, so it's not immediately clear which reference is the best match.
```
allyhawkins, line 22: I don't know if this is where I would start the exploration. I think I first want to know how do these references perform in samples that are actually part of OsteoCAR vs other samples with similar characteristics that are not part of OsteoCAR. We don't really have a good ground truth for these samples that have a mismatch, so I think we first need a baseline of how each reference performs on samples where we do have some sort of expectation or "ground truth" to compare to.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references.Rmd
```diff
@@ -0,0 +1,468 @@
...
+  readr::read_tsv(file.path(data_dir, "SCPCP000018", "single_cell_metadata.tsv")) |>
+    # rename to match
+    dplyr::rename(primary_or_metastasis = disease_timing), 
+  readr::read_tsv(file.path(data_dir, "SCPCP000023", "single_cell_metadata.tsv"))
+) |>
+  dplyr::select(all_of(cols_to_keep))
+```
+
+Let's grab the weird ones.
+We'd also like to test a couple libraries as controls here, as a way to get a sense of merged vs. both.
```
allyhawkins, line 125: This sentence is confusing to me and I had to go back to the PR comment to understand what you meant by controls. I would just add more context about what you mean by control here.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references.Rmd
```diff
@@ -0,0 +1,468 @@
...
+* "merged": for this, I merge the aggregated references and provided this single reference to `SingleR`.
+Cell type labels are not merged but were renamed so they reflect the originating reference.
+
+For each dataset, I make three plots:
+
+1. A barplot showing cell types inferred with each reference
+2. A Jaccard index heatmap comparing the primary to metastasis reference.
+This will be helpful for understanding how similarities or differences between labels in these references play out.
+3. A Jaccard index heatmap comparing the "merged" and "both" references. 
+This will be helpful for understanding which approach may be more robust for our needs
```
allyhawkins, line 196: I think the key comparison is missing here, which is comparing the tissue-appropriate reference (eg.., primary for primary samples and metastasis for metastasis samples) to the merged. And then the same comparison to the "both" reference. I want to know if you use the combined reference compared to the single reference, are you gaining any additional cell types that might not be present in one of the references? And if you are, do those cell types seem like valid matches? This is where having the delta next information would be helpful, but only for those "new" cell types. 

Additionally, do you see that cells that get assigned to celltype X in the single reference, do those cells get split across "prim_celltypeX" and "mets_celltypeX" or are they consistently assigned to the appropriate reference type.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references.Rmd
```diff
@@ -0,0 +1,468 @@
...
+### Control lung mets sample
+
+* This did not return a straightforward answer that suggests to use the mets reference - it seems that many primary reference cell types are coming back from each of the "merged" and "both" strategies and these results do not fully agree with the results from mets alone.
+* Importantly, `Fibroblasts` is _only_ a cell type in the mets reference, not the prim, but we only see this dominating in the results when mets is used. 
+When either of the both/merged is used, cell types from the prim reference seem to become much more prominent, in particular CAFs and Endothelial cells, and fibroblasts tends to lose out to those options.
+
+For this one, it's helpful to also look at the mets vs. both heatmap:
+
+```{r}
+make_jaccard_df(result_list$SCPCL000856_control_Lung$mets, result_list$SCPCL000856_control_Lung$both) |>
```
allyhawkins, line 402: To me this is the more informative of the plots.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references.Rmd
```diff
@@ -0,0 +1,468 @@
...
+```
+
+
+## Conclusions
+
+### Control lung mets sample
+
+* This did not return a straightforward answer that suggests to use the mets reference - it seems that many primary reference cell types are coming back from each of the "merged" and "both" strategies and these results do not fully agree with the results from mets alone.
+* Importantly, `Fibroblasts` is _only_ a cell type in the mets reference, not the prim, but we only see this dominating in the results when mets is used. 
+When either of the both/merged is used, cell types from the prim reference seem to become much more prominent, in particular CAFs and Endothelial cells, and fibroblasts tends to lose out to those options.
```
allyhawkins, line 397: This tells me we probably need to use a combined reference, but I also think this is where delta next could _maybe_ be helpful. In particular, for the cells that match to a primary cell type in the reference instead of a mets cell type, how confident are we in that prediction and what's the second best label? 

For example, if the cell type is "mets_Fibroblasts" when using mets only but those cells get split between primary fibroblasts and mets fibroblasts then could we instead harmonize the labels?

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references.Rmd
```diff
@@ -0,0 +1,468 @@
...
+
+```{r}
+#| fig.height: 12
+#| fig.width: 8
+result_list |>
+  purrr::imap(plot_results)
+```
+
+
+## Conclusions
```
allyhawkins, line 391: Just a small organizational note that this would be nicer to have below each section of plots. Its hard to scroll up and down for the conclusions and the plots.

allyhawkins (commented):
Thanks for getting this comparison started and there is definitely a lot to think about. I have a lot of ideas for things that we could be explored, but I also want to make sure we keep this PR well-scoped. So I think we can use a lot of what you have here, but I think we need to take a step back before we get into the "tricky" samples. 

To me, the main question that we are trying to answer is: is it appropriate to use both primary and metastatic references together or should we treat them separately based on disease timing? I think to answer the question about which reference is "appropriate", we need some sort of ground truth or as close to a ground truth as we can get. Because of that, I would make some adjustments to the experimental design. 

I would pick two "control" samples, one sample that's part of both ScPCA and OsteoCar for primary and one for metastasis. Then for each sample you run primary or mets only and compare to a joined reference. For each sample/reference combination, you can compare the labels to the original labels and quantify how many cells get the same label. I want to see how much this number changes when you use primary-only vs the combined versions. 

Along that same line, I would add a harmonized version of the reference where you don't include "prim" or "met" in the label so that CAFs from both get the same label. Then compare those results to the both and merged versions, where you do have a label. Do you see that cells are evenly split between the primary and mets version of the same label or do they favor one over the other? 

I think once we narrow in on an approach that would work well for the majority of samples then we can revisit the "tricky" samples.

sjspielman:
@allyhawkins sending an update - I've been working on the this and the overhaul is going to be pretty extensive, even though some of the code is shared. It's going to make a lot more sense for a fresh PR, so I'm going to go ahead and close this. Also note, I agree that delta.next is important to bring in here, so I have decided to drop the approach where we literally provide both references to SingleR, because this does not return a delta.next due to the more involved procedure.

I'll also point out, because of SingleR's scores fine-tuning step, my understanding is that it is not definitively possible to determine the actual "second best label." For example, there are many times that the second best label _based ont he returned scores alone_ is in fact that is reported as the actual assigned label. But, I think the SingleR heatmaps get us close to what we want to actually be answering here so that is a strategy I am using.

PR #1428, opened 2026-08-20
Compare reference use on patient samples from OsteoCAR

sjspielman:

Part of #1424
Emerged out of the closed PR #1427 

The goal of this PR is to do some reference exploration specifically for the patient references at the `Ann_Level2` annotation.

- I chose two samples with good barcode overlap between ScPCA and OsteoCAR, one primary and one metastasis (but I do not show my work here! There is another exploratory notebook I can send in as a separate PR to see precisely how I did this, please let me know if this is something you'd want to review. I didn't think it was critical to record so I didn't want to waste reviewer time on it, is why it's not here.).
- I run with three references: the one matched to its type (prim/mets), a harmonized version (same labels collapsed into one), and a concatenated version (labels retain their original prim/mets identity). I do not use a strategy where I provide both to SingleR since this nixes delta.next scores and these are pretty informative imo.
- I then compare cell types in a variety of ways:
  -  Barplots of the cell types identified per reference, along with the delta.next distributions for those cell types.
  - SingleR heatmaps which give us a sense of what other cell types are contending for the label
  - Jaccard heatmaps comparing the labels to ground truth annotations
  - Note that at one point I considered some actual accuracy measures e.g. with `caret`, but since the matrix comparison isn't square this is a bit tricky to interpret since some of the assumptions of confusion matrix metrics are violated. The plots seemed to communicate well enough what was going on without putting a number on it I think; let me know if you disagree. 
- The analysis is the same between the primary and the metastasis sample, but with conclusions interspersed throughout the text.

There are also quite a few functions used here, so I split them out to be called as a child report since it's just way too overwhelming otherwise, and the code is long enough as is!

One of the conclusions I draw here is we might want to do this same exploration with `Ann_Level1` next, since the "divvying up" of those cell type seems to create challenging noise. But, the harmonized reference seems like a reasonable bet otherwise.

[03-compare-patient-references-osteocar.nb.html](https://github.com/user-attachments/files/31274559/03-compare-patient-references-osteocar.nb.html)

Laptop, renv

The notebook!

We'll discuss how many next steps I should take, if any, after this PR. For example, I could pursue analogous notebooks for the xenograft references or for the patient reference but at `Ann_Level1`.


analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+
+In this notebook, we'll do some cursory runs of `SingleR` on a couple osteosarcoma samples of varying metadata and compare results when using:
+
+* the reference that matches its type
+* a concatenated version of these references with cell types between them kept distinct ("concatenated")
+* a concatenated version of these references with cell types between them _merged_ where they share a label ("harmonized")
+
+Note that we don't explore here the approach of providing both references to `SingleR`, because the way that approach works necessarily does not return a `delta.next` value which we would really like here for diagnostic purposes.
+
+We'll specifically explore a two samples which appear in ScPCA _and_ OsteoCar, since this represents the closest we can get to "ground truth," - one primary and one metastasis.
```
allyhawkins:
```suggestion
We'll specifically explore two samples which appear in ScPCA _and_ OsteoCar, since this represents the closest we can get to "ground truth," - one primary and one metastasis.
```

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+
+
+Source functions which are separated into their own file for both legibility and re-use:
+
+```{r child='exploratory-utils.Rmd'}
+```
+
+
+We'll define the samples we're using here.
+These were chosen based on manual inspection of how many cells are present and how many barcodes overlap with `OsteoCAR` to be able to get some decent signal here.
```
allyhawkins, line 117: Just curious what was the percentage of overlap for each sample? That might be helpful to know and record here.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+
+* `prim_aggr`: Patient primary
+* `mets_aggr`: Patient metastasis
+* `harmonized_aggr`: Merged references, where cell type labels are retained. 
+In other words, if both references have T cells, we end up with only a single T cell label in `harmonized_aggr`
+* `concatenated_aggr`: Merged references, where cell type labels are prefixed with `mets_` and `prim_`.
+In other words, if both references have T cells, we end up with two labels in `concatenated_aggr`, e.g. `mets_tcell` and `prim_tcell`
+
+Note that the `harmonized_aggr` and `concatenated_aggr` are created from `prim_aggr` and `mets_aggr`, because of memory.
+In other words, the raw references could not be combined on their own.
```
allyhawkins, line 148: What do you mean that the raw references couldn't be combined on their own?

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+  rOpenScPCA::sce_to_symbols(reference = "sce")
+
+sce
+```
+
+
+Next, we'll run `SingleR`:
+
+```{r}
+results <- run_singler(
```
allyhawkins, line 288: I might be missing something, but it looks like this just produces one result from `SingleR`? I think you might be missing code that actually runs this on each reference and then creates a list of results to use as input to making the plots below.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+#| fig.height: 6
+purrr::imap(results, plot_results) |>
+  patchwork::wrap_plots(nrow = 3) + patchwork::plot_annotation(title = "Primary patient sample") 
+```
+
+
+Across all references, we see various tumor cell types coming back as most common.
+When a primary reference is used or explicitly included as a standalone cell type, we see that proliferative tends to dominate, but the harmonized reference has basal progenitors dominating.
+These are the second highest category for primary and concatenated. 
+Labels for the concatenated reference tend to be primary, not metastasis.
```
allyhawkins, line 317: I think it would be helpful to add a plot here that specifically focuses on one cell type and compares the delta next scores for those cell types across methods. For example, if you look at cells that are always labeled as proliferative cells, regardless of the method, how does the delta next score change? You can maybe get an idea of that here but with different x axes limits it's hard to make a direct comparison. 

What I would envision is a sina or ridge plot with the reference type on one axis and then the other axis is delta. next. Only cells that are consistently labeled the same across all three types should be included to see how the "confidence" changes with each reference type.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+
+```{r}
+SingleR::plotScoreHeatmap(results$`Harmonized reference`, main = "Harmonized reference")
+```
+
+```{r}
+SingleR::plotScoreHeatmap(results$`Concatenated reference`, main = "Concatenated reference")
+```
+
+The same general pattern is present all around: there are several cell types that have very high scores for most cells, making it challenging to definitively call cell types with confidence.
```
allyhawkins, line 360: I think this is an important finding and helpful to show. One could argue that SingleR isn't doing a very good job of being able to differentiate between tumor cell types regardless of the reference type.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+        res_name
+      )
+    }
+  ) |>
+  patchwork::wrap_plots(ncol = 1)
+```
+
+There is certainly some good agreement, but we again the the same pattern play out there isn't always precise agreement between which type of tumor cell is bring called.
+This story is similar across all references used.
+This suggests we might actually be well served by using `Ann_Level1`, even though it is less fine-grained.
```
allyhawkins, line 389: I tend to agree with this. For the primary reference only we see that we are almost doubling the amount of cells that are getting labeled as proliferative. This tells me we may not have enough information in the SingleR model to properly differentiate the other tumor cell types with this level of annotation.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+#| fig.width: 6
+#| fig.height: 6
+purrr::imap(results, plot_results) |>
+  patchwork::wrap_plots(nrow = 3) + patchwork::plot_annotation(title = "Metastasis patient sample") 
+```
+
+For all references, TAMs always come back as the most prevalent cell type, and it's nice to see that the median `delta.next` for these calls is decently above 0.
+When the primary reference is included whether either the harmonized or concatenated strategy, we see that basal progenitors get picked up as a major cell type, and we end up with somewhat fewer proliferative cells.
+
+For `delta.next` in general, the immune and epithelial cell types (and endothelial when detected) are coming back with higher medians compared to all other cell tyopes which are quite close to 0, suggesting challenges in confidently resolving these.
```
allyhawkins:
```suggestion
For `delta.next` in general, the immune and epithelial cell types (and endothelial when detected) are coming back with higher medians compared to all other cell types which are quite close to 0, suggesting challenges in confidently resolving the individual tumor cell types. 
```
I think this is what you mean here? The higher values for non-tumor cell types tells me that those have distinct profiles used for assigning cell types while the various tumor cell types are harder to differentiate.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+SingleR::plotScoreHeatmap(results$`Metastasis patient reference`, main = "Metastasis patient reference")
+```
+
+```{r}
+SingleR::plotScoreHeatmap(results$`Harmonized reference`, main = "Harmonized reference")
+```
+
+```{r}
+#| fig.height: 6
+SingleR::plotScoreHeatmap(results$`Concatenated reference`, main = "Concatenated reference")
```
allyhawkins, line 493: One thing I note about this plot, which I think is really helpful to us deciding between harmonized vs concatenated is that there's a group of immune cells that appear to be "mets_T_cells" also have high scores for "prim_T_cells". You see a similar phenomenon with the proliferative tumor cell types. This is the information I'm looking for to help determine if specifying if the reference is from primary or mets is important or not, and I'm not convinced that it is.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+        compare_df$pruned.labels, 
+        "",
+        res_name
+      )
+    }
+  ) |>
+  patchwork::wrap_plots(ncol = 1)
+```
+
+On the whole, this is a much cleaner result compared to the primary sample we explored in the previous section, but we also have many more cells being compared here!
```
allyhawkins: I think you mean we have more cell types being compared here. I agree that this is cleaner, but this is probably because the mets sample has both normal and tumor cells and we don't see that for the primary samples. I wonder how representative this primary sample is to the rest of the primary samples? Are most of them just tumor cells or do we see evidence for non-tumor cells. I think if we have samples with a "ground truth" and both normal and tumor cells, it would be interesting to see how they compare to this primary sample. You don't need to do that here, but it would be helpful to know.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,398 @@
...
+The metastasis (matched) and harmonized references seem to do well, and from the concatenated reference we can see that when a "primary" label is coming back, it generally matches the cell type we expect based on ground truth.
+For example, see the ground truth for monocytes - this is largely split between mets and primary monocytes.
+
+## Conclusions
+
+There appears to be a decent match between ground truth and `SingleR` labels for many of the references explored here.
+We see (in particular for primary patient sample) that there tends to be ambiguity among the tumor cell types, as the `SingleR` heatmaps emphasize.
+One limitation here is that the primary sample doesn't have a ton of cell type diversity besides tumor cells.
+
+Based on this exploration, the harmonized reference (where we combine aggregated references and _merge the labels_) seems like a decent bet, but it would also be worthwhile to explore using the coarser-grained `Ann_Level1` to see what we would gain vs lose compared to `Ann_Level2`.
```
allyhawkins, line 534: Before comparing what we gain or lose, we should just see if we can better recapitulate the ground truth with the less granular cell types.

allyhawkins (commented):
This looks like a good start and I think will definitely be helpful in determining the reference type we use. I had one suggestion for an additional plot that I think should be added to better compare delta next values across reference types. Other than that I just left some general food for thought questions and comments.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
sjspielman: > Just curious what was the percentage of overlap for each sample

To confirm, do you mean just these or all of them? If all samples, I'd put that into a different notebook.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
sjspielman: If I try to `cbind` SCE objects before aggregating them, R always runs out of memory no matter how much I gave it. I was only able to combine references after I aggregated each individual reference. This means that the pseudobulking is done for each reference separately - the same cell type label will be pseudobulked just for metastasis cells and separately just for primary cells, and those pseudobulked outputs end up getting combined with `cbind`.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
sjspielman: Yes, this is also my conclusion

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
allyhawkins: I think having the information for just these samples is sufficient.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
allyhawkins: Hmmm it might be worth creating a combined object and then downsampling each cell type to have X number of cells. That might ensure we are properly creating the combined references but then we are dealing with the effect of downsampling. I think if we decide to use a harmonized reference we should test both ways of building the reference to see if there are any differences.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
sjspielman: This function in defined in the `exploratory-utils.Rmd` file: https://github.com/sjspielman/OpenScPCA-analysis/blob/5aa8dfa6a9fe9f9842be5ef7486cb98428cf009d/analyses/cell-type-osteosarcoma/exploratory-notebooks/exploratory-utils.Rmd#L13

It runs on three different references. I can add more comments to be clear where functions come from.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
sjspielman: >  I think if we have samples with a "ground truth" and both normal and tumor cells, it would be interesting to see how they compare to this primary sample. You don't need to do that here, but it would be helpful to know.

This suggests me it would be helpful for me to send in a separate PR with a notebook looking over all ScPCA samples that are also in OsteoCAR, to overall look at the overlap of barcodes and the cell type distributions? Not to run SingleR on, but to get a sense of the landscape there.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
sjspielman: > Hmmm it might be worth creating a combined object and then downsampling each cell type to have X number of cells.

I can do this, but again it would be reversed - need to downsample cells _in order to_ combine them in the first place.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
allyhawkins: Can you actually move this function in particular to be in this notebook? I think it's easier to follow for exploratory notebooks to have some of the key implementations, in this case how `SingleR` was actually run in the notebook itself.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
allyhawkins: Yes that could be informative to get some basic information on where we are starting with this dataset.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
allyhawkins: > I can do this, but again it would be reversed - need to downsample cells in order to combine them in the first place.

I think you should be able to `cbind` them, downsample, and then create the aggregated reference.

sjspielman:
I want to note that we discussed offline taking the following next steps here:

1. Complete this notebook with the current references as implemented. Since prim/mets references as-is (in their original SCE form) cannot be readily combined locally due to memory limitations, we'd want to downsample these references first before concatenating the raw versions, for testing different approaches to make a combined reference. Exploring the strategy for combining references is something to pick up later. 

2. Add a PR with a second notebook doing a similar exploration but using the `Ann_Level1` reference instead. This may also involve using `Ann_Level2` for non-tumor cell types but collapsing all tumor cell types into a single `Tumor` category, as `Ann_Level1` reflects.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
sjspielman: >  I wonder how representative this primary sample is to the rest of the primary samples? Are most of them just tumor cells or do we see evidence for non-tumor cells.

Let me just note: In exploring other samples one could use here, what I found is that samples with more diversity in cell types tend to have many fewer overlapping barcodes with the ScPCA version of the library, which makes the ground truth comparison harder to do simply because of dataset size.

sjspielman:
Based on the scope we have discussed for this PR, this should be ready for re-review. 

- I brought the `run_singler()` function into the main notebook, but left the rest in the `utils` Rmd. 
- At the beginning of each sample analysis section, I include the fraction of ScPCA cells that are present in OsteoCar as well as barplots of the "ground truth labels," faceted by `ann_level1`.
- I added one function to be able to make the plot you requested, which shows delta.next scores across references for matching labels. I ended up making this as ridge plots faceted by cell type so we can have a close comparison. There is some minor signal here, but there is no definitive trend; distributions are generally the same.

Let me know any other changes or recommended additions! After this, I'll proceed to do a fairly similar notebook but with the less fine-grained tumor annotations.

[03-compare-patient-references-osteocar.nb.html](https://github.com/user-attachments/files/31522326/03-compare-patient-references-osteocar.nb.html)

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,540 @@
...
+
+```{r}
+nrow(prim_ground_truth) / ncol(sce)
+```
+
+
+What are the cell types as recorded in OsteoCAR?
+
+```{r}
+ggplot(prim_ground_truth) +
```
allyhawkins, line 274: I think this plot is fine, but you could also just print out a table of cell type counts. I think that would make it easier to see the imbalance between tumor and normal cells.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
```diff
@@ -0,0 +1,540 @@
...
+The plot considers only the top 10 cell types (if there are that many in common), and only plots cell types with at least 3 cells per reference.
+This will show us if any of the references is more confidently assigning these cell types.
+
+
+```{r}
+#| fig.width: 4
+#| fig.height: 5
+#| message: false
+
+plot_delta_next_per_reference(results)
```
allyhawkins, line 333: This plot is helpful thank you! I personally don't see any clear trend in one reference always being better than the others, which is good information to know. That tells me we don't get any benefit from using the primary only vs a combined reference.

allyhawkins (approved):
LGTM. I just have one small comment about turning the cell type plot into a table but either one is fine.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
sjspielman: I ended up making it `facet_grid()` which I thinks get the point across better! This is easier for me to read than a table which spans multiple pages, so this way we see it all at once.

Screenshot 2026-08-28 at 1 34 03 PM

analyses/cell-type-osteosarcoma/exploratory-notebooks/03-compare-patient-references-osteocar.Rmd
sjspielman: Agreed.

PR #1429, opened 2026-08-28
Compare patient references with Ann_Level1 for tumor/mesenchymal

sjspielman:

Ever towards #1424

This PR adds a notebook that is _very_ similar to the one we completed in #1428, except I use Ann_Level1..ish! In the last notebook, it appeared that SingleR had similar difficult with cell types in the "Tumor" and "Mesenchymal" groups. For this round, I therefore decided to group all into a single category _each_ (all tumor together, all mesenchymal together); this is the Ann_Level1 usage. For the other cell types (endo, epi, immune), I still use their Ann_Level2 finer annotations. 

When constructing the combined reference versions here, I performed aggregation on Ann_Level2 and _then_ collapsed the tumor & mesenchymal categories, rather than aggregating on an object with all those labels combined in the first place. This allows each of the sub cell types to still contribute more even weight towards the final aggregation, although this of course was a choice which can/should be empirically tested, as I indicated at the bottom of the notebook.

With this approach, we definitely get more confidence in calling tumor cells, but mesenchymal delta.next are still pretty low overall. Based on this notebook, I think this is a safer annotation approach and a harmonized reference (where we merge labels) would work well. I imagine that next steps would likely be comparing different ways to aggregate. 

Notebook:
[03b-compare-patient-references-combined-ann-levels.nb.html](https://github.com/user-attachments/files/31566020/03b-compare-patient-references-combined-ann-levels.nb.html)

I also found and fixed a few typos in the last notebook, which I renamed to 03a.


analyses/cell-type-osteosarcoma/exploratory-notebooks/03b-compare-patient-references-combined-ann-levels.Rmd
```diff
@@ -0,0 +1,566 @@
...
+patient_mets <- readr::read_rds(file.path(ref_dir, "patient_mets_sce.rds"))
+counts(patient_prim) <- NULL
+counts(patient_mets) <- NULL
+```
+
+Before we aggregate, we'll grab the ground truth cell types we're interested in for comparison from the raw reference files.
+In this case, we're interested in:
+
+- All tumor and mesenchymal from level 1 combined into a single label each
+- Other labels take from level 1
```
allyhawkins: I think this bullet needs to be updated?

analyses/cell-type-osteosarcoma/exploratory-notebooks/03b-compare-patient-references-combined-ann-levels.Rmd
```diff
@@ -0,0 +1,566 @@
...
+```{r}
+#| fig.width: 6
+#| fig.height: 6
+purrr::imap(results, plot_results) |>
+  patchwork::wrap_plots(nrow = 3) + patchwork::plot_annotation(title = "Primary patient sample") 
+```
+
+
+We see that the Tumor `delta.next` scores are all much higher than what we previously observed in the `03a` notebook, suggesting that combining this into a single category is giving us higher confidence to label malignant cells.
+But, it is notable that the Mesenchymal category still has low associated `delta.next` values.
```
allyhawkins, line 363: This makes sense to me given how similar tumor is to mesenchymal here. But I do think this is a good validation of why combining the tumor labels makes sense for now.

analyses/cell-type-osteosarcoma/exploratory-notebooks/03b-compare-patient-references-combined-ann-levels.Rmd
```diff
@@ -0,0 +1,566 @@
...
+  rOpenScPCA::sce_to_symbols(reference = "sce")
+
+sce
+```
+
+Here are the ground truth cell types:
+
+```{r}
+#| fig.width: 6
+#| fig.height: 4
```
allyhawkins: Can you increase the height here so we can read all the labels?

analyses/cell-type-osteosarcoma/exploratory-notebooks/03b-compare-patient-references-combined-ann-levels.Rmd
```diff
@@ -0,0 +1,566 @@
...
+
+```{r}
+#| fig.width: 4
+#| fig.height: 7.5
+#| message: false
+
+plot_delta_next_per_reference(results)
+```
+
+There's not much of a strong trend here; median values look about the same for each set of `delta.next` distributions, but it's worth noting that the mesenchymal `delta.next` appears larger for the matched metastasis reference vs the combined.
```
allyhawkins, line 506: I agree with this and I don't think it's strong enough evidence to choose the matched references over the harmonized ones. That is to say, I like the choice of using the harmonized reference for all samples.

allyhawkins (approved):
I think this all looks good and answers the questions we want to address here. I like the approach you used to combine tumor and mesenchymal and then use the normal cell labels from level 2. And I think at least with these samples, using the harmonized reference makes sense. I had a few minor comments but nothing that I need to see again.

PR #1433, opened 2026-09-14
Add script to create the NB exploratory plots

### LINKED ISSUE #1431: Create preliminary plots evaluating metaprograms from Neuroblastoma pilot

_No response_

We generated metaprograms on all Neuroblastoma samples in SCPCP000004 with `ews-nf`, which follows the same pipeline being implemented in `metafactory`. We want to create some initial figures that we can use to summarize these metaprograms. 

A script that reads in the metaprograms and outputs figures that might be helpful to show for this pilot group when presenting the `metafactory` workflow. The exact figures I want to make are TBD, but might include a heatmap of the top gene sets across each metaprogram, the correlation heatmap, and/or a UMAP of all neuroendocrine cells where cells are colored by metaprogram scores. 

_No response_

_No response_

_No response_

allyhawkins:

Here I'm adding a script that was used to generate the exploratory plots that we are displaying for the Neuroblastoma results in the pediatric AACR poster and presentation. 

Closes #1431 

We have two plots that we want to show - a heatmap of the top gene sets for each metaprogram and a faceted UMAP showing the metaprogram scores across all samples for 4 of the identified metaprograms. 

For the faceted UMAP, I read in the merged object and integrate using Harmony. I chose to consider both library ID and participant ID since many of these cases are multiple samples from the same patient. I did both with and without patient and it didn't look that different, so I think for these purposes this integration approach is reasonable. The metaprograms were chosen because they are ones that were present in NBAtlas and had obvious variation across the dataset. 

For the heatmap, I read in the ORA results that are stored in the metaprogram object and create the heatmap using the same styling as the heatmap for the SKC results. 

I stored the results from running `ews-nf` through in `s3://researcher-211125375652-us-east-2/metafactory-analysis/results` and copied them to the `results` folder in this module so that I could run this script. 

The output here is a set of plots which I saved as individual files in the `plots` directory. 

### Author checklists

_Check all those that apply._


analyses/metafactory-analysis/scripts/neuroblastoma-exploratory-plots.R
```diff
@@ -0,0 +1,181 @@
...
+      geom_point(size = 0.1) +
+      theme_classic() +
+      scale_color_viridis_c(option = "turbo") +
+      labs(
+        x = "UMAP1",
+        y = "UMAP2",
+        title = mp_name
+      ) +
+      theme(text = element_text(size = 12), 
+            aspect.ratio = 1)
```
jashapiro: Do we want to remove the UMAP labels?
```suggestion
      theme(
        text = element_text(size = 12), 
        axis.text = element_blank(),
        axis.ticks = element_blank(),
        aspect.ratio = 1
      )
```

analyses/metafactory-analysis/scripts/neuroblastoma-exploratory-plots.R
```diff
@@ -0,0 +1,181 @@
...
+# list of metaprograms and associated names
+metaprograms_to_plot <- c(
+  "Translation" = "MP02", 
+  "MYC" = "MP04", 
+  "Differentiated" = "MP05", 
+  "Cycling" = "MP06"
+)
+
+# make a faceted plot for the specified metaprograms
+metaprograms_to_plot |>
```
jashapiro: For explicitness, save to a variable and then use that in the `ggsave()`?
```suggestion
metaprogram_plot <- metaprograms_to_plot |>
```

analyses/metafactory-analysis/scripts/neuroblastoma-exploratory-plots.R
```diff
@@ -0,0 +1,181 @@
...
+ora_ht <- ComplexHeatmap::Heatmap(
+  heatmap_mtx,
+  name = "-log10(pvalue)\n",
+  row_names_side = "left",
+  row_dend_side = "right",
+  cluster_rows = TRUE,
+  cluster_columns = FALSE,
+  col = circlize::colorRamp2(c(0, 20), colors = c("gray95", "darkslateblue")),
+  rect_gp = grid::gpar(col = "white", lwd = 2), # add some white lines around each box
+  row_names_gp = grid::gpar(fontsize = 12),
```
jashapiro: There is something just a bit odd with this heatmap where the spacing on the multi-line labels is a bit bigger than the spacing between labels, so the line groupings look wrong (if that makes sense). I think the simplest solution is to just make this font a bit smaller. 
```suggestion
  row_names_gp = grid::gpar(fontsize = 10),
```

But you could also try adjusting the line height directly, which I think is something like:

```suggestion
  row_names_gp = grid::gpar(fontsize = 12, lineheight = 1),
```
(default is 1.2, according to google)

analyses/metafactory-analysis/scripts/neuroblastoma-exploratory-plots.R
```diff
@@ -0,0 +1,181 @@
...
+
+# make a faceted plot for the specified metaprograms
+metaprograms_to_plot |>
+  purrr::imap(\(mp, mp_name){
+
+    ggplot(coldata_df,
+           aes(x = UMAP_harmony.1, y = UMAP_harmony.2, color = .data[[mp]])) +
+      geom_point(size = 0.1) +
+      theme_classic() +
+      scale_color_viridis_c(option = "turbo") +
```
jashapiro: I wonder if we want to use different colors for each? I'm not too particular here, but I don't _love_ `turbo`.

jashapiro (commented):
Overall this looks good, but I had a couple little thoughts about formatting.

analyses/metafactory-analysis/scripts/neuroblastoma-exploratory-plots.R
allyhawkins: > I wonder if we want to use different colors for each? I'm not too particular here, but I don't _love_ `turbo`.

Do you mean different colors for each facet? Or just different colors in general? I feel like different colors for each facet might be confusing.

jashapiro: I did mean different colors for each facet, but I don't really think it is necessary. I was thinking something where each was a different color gradient like the colorbrewer hues.

allyhawkins: Okay I played around with this. Do you like any of these better?

Image

Image

jashapiro: I like the first one, as long as you don't think it is confusing...

allyhawkins: > I like the first one, as long as you don't think it is confusing...

It looks better than I thought it would so I updated this to use that palette instead.

analyses/metafactory-analysis/scripts/neuroblastoma-exploratory-plots.R
```diff
@@ -0,0 +1,193 @@
...
+ora_ht <- ComplexHeatmap::Heatmap(
+  heatmap_mtx,
+  name = "-log10(pvalue)\n",
+  row_names_side = "left",
+  row_dend_side = "right",
+  cluster_rows = TRUE,
+  cluster_columns = FALSE,
+  col = circlize::colorRamp2(c(0, 20), colors = c("gray95", "darkslateblue")),
+  rect_gp = grid::gpar(col = "white", lwd = 2), # add some white lines around each box
+  row_names_gp = grid::gpar(fontsize = 12, lineheight = 0.8),
```
jashapiro: This still looks a bit rough to me, but I think it is  probably just a function of the long names. I might try shrinking the font again, but it depends on readability.

jashapiro (approved):
LGTM, even if the long labels are still trouble...

analyses/metafactory-analysis/scripts/neuroblastoma-exploratory-plots.R
allyhawkins: Yea I tried a couple different ways and none of them look great but I did shrink the font size and it looks slightly better, so I'll stick with that change.

